PR #6, opened 2022-04-04
Enable SonarCloud and testrun as part of the pipeline

haakemon:
## Description
- Extend `build` workflow with
	- run eslint
	- run unit tests
	- run SonarCloud analysis + code coverage report
- Simplify `tsconfig.json`, where the main config is on root, and each project has minor overrides where needed. 
- Add some eslint-ignore statements to get build to go through -not sure how we should deal with these issues, but it would need some investigation. Created #8 to look into this.


lorang92 (approved):
Nice, looks great 🥳

PR #9, opened 2022-04-08
Extend build action to push new app-frontend version to CDN on main branch

haakemon:
When running the build action on main branch, it will now deploy a new version of app-frontend to CDN, similar to the old pipeline we had in AzureDevOps.

### DIFF
--- a/src/altinn-app-frontend/package.json
+++ b/src/altinn-app-frontend/package.json
@@ -1,6 +1,6 @@
 {
   "name": "altinn-app-frontend",
-  "version": "3.33.4",
+  "version": "3.33.5",
   "description": "",
   "main": "index.js",
   "scripts": {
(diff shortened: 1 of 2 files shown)


src/altinn-app-frontend/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "altinn-app-frontend",
-  "version": "3.33.4",
+  "version": "3.33.5",
```
lorang92, line 3:
```suggestion
  "version": "3.33.4",
```

No need for a new version here, right?

lorang92 (approved):
Looks good to me! Do you have any other thoughts @tjololo ?

tjololo (approved):
:godmode:

src/altinn-app-frontend/package.json
haakemon: Not really, but I didnt want to overwrite whats currently on CDN, so I figured a new patch version would be better.

I guess this can be removed though, and we can merge some branches with actual new features before this, and we will anyway get a new version out without overwriting the stuff thats already there.

Issue #8, opened 2022-04-05
Fix accessibility issues reported by eslint


haakemon:
We had to add some eslint-disable statements as part of #6, to get the build to go through.

We should investigate what the issues are, and fix them properly. See commits in #6 for details.

PR #48, opened 2022-04-20
Wcag/8 fix accessibiliy issues reported by eslint

mjulstein:
## Description
Focusing on the accessibility issues reported by eslint (a11y )
**Note:** These changes are not tested with other means than the unittests,
TODO:
 - I need to find or create a case where the components are in use in order to test if I broke anything.
 - I should create unit-tests once the components are verified.

## Fixes
- #8 

## Verification
- [x] All tests run green


src/shared/src/components/molecules/AltinnModal.tsx
```diff
@@ -98,66 +92,53 @@ const styles = createStyles({
...
+      <div className={classes.header}>
+        {props.hideCloseIcon && props.hideCloseIcon === true ? null :
+          <IconButton
+            className={classNames(
+              classes.iconBtn,
+              {[classes.closeButtonOutsideModal]: props.closeButtonOutsideModal === true},
+            )}
+            onClick={props.onClose}
+          >
+            <i className={classNames('ai ai-exit-test', classes.iconStyling)} tabIndex={-1}/>
```
haakemon, line 110: ~~This doesnt seem entirely correct to me - this will remove the ability to navigate to this by tabbing. I would think moving the tabIndex property to the parent `IconButton` would be better?~~ 

Or maybe its not needed at all, since I guess `IconButton` uses a `button` under the hood anyway, and so it will by default be possible to navigate to it.

So `tabindex` could probably removed entirely in this case.

WDYT?

src/shared/src/components/molecules/AltinnModal.tsx
```diff
@@ -98,66 +92,53 @@ const styles = createStyles({
...
+  if (!props.printView) {
+    const {isOpen, hideBackdrop, allowCloseOnBackdropClick, onClose} = props;
+    return (
+      <Modal
+        open={isOpen}
+        className={classes.scroll}
+        hideBackdrop={hideBackdrop}
+        onBackdropClick={allowCloseOnBackdropClick === false ? null : onClose}
+      >
+        <StyledModal {...props}/>
```
haakemon, line 135: I would suggest here as well to destruct immediately:

```suggestion
const AltinnModal = ({isOpen, hideBackdrop, allowCloseOnBackdropClick, onClose, printView, ...rest}: React.PropsWithChildren<IAltinnModalComponentProvidedProps>) => {
  const classes = useStyles();
  if (!printView) {
    return (
      <Modal
        open={isOpen}
        className={classes.scroll}
        hideBackdrop={hideBackdrop}
        onBackdropClick={allowCloseOnBackdropClick === false ? null : onClose}
      >
        <StyledModal onClose={onClose} {...rest}/>
```

src/shared/src/components/AltinnInput.tsx
```diff
@@ -92,8 +92,7 @@ function AltinnInput(props: IAltinnInputProps) {
         className={validationError ? classes.altinnInputValidationError : classes.altinnInput}
       >
       {iconString ?
-        // eslint-disable-next-line jsx-a11y/click-events-have-key-events,jsx-a11y/no-static-element-interactions
-        <i data-testid="altinninput-iconString" className={`${classes.altinnInputIcon} ${iconString}`} onClick={focusInput}/> :
```
haakemon, line 96: Do we not need the click handler here to focus the input?

src/altinn-app-frontend/src/components/summary/SummaryComponent.tsx
```diff
@@ -150,48 +148,46 @@ export function SummaryComponent(props: ISummaryComponent) {
...
           <SummaryGroupComponent
-            onChangeClick={onChangeClick}
-            changeText={changeText}
+            {...change}
             {...props}
           />
         );
       }
-      case 'FileUpload': {
+      case 'fileupload': {
```
haakemon, line 164: Not sure we can change the casings here - this is case sensitive, isnt it?

src/shared/src/components/molecules/AltinnModal.tsx
mjulstein: I think you are right. My first intent was to remove ( see first commit ), but because I don't currently know how this affects the component in practice I wanted a second opinion. From what I could get by asking @lorang92 yesterday, the other case where `tabindex={0}` was used it was intentionally because of the need to use `focus()` on an non interactive element.
In this case I'm not sure, but left it if that was the case here as well.
WDYT?

src/shared/src/components/AltinnInput.tsx
mjulstein: My understanding was that it was redundant as it is also on the surrounding grid on line 74?

src/altinn-app-frontend/src/components/base/InputComponent.tsx
```diff
@@ -64,9 +64,10 @@ function NumberFormatCustom(props: IFormattedNumberInputProps) {
 }
 
 export function BasicInputComponent(props: IBasicInputProps) {
+  const { inputRef, ...rest } = props;
   return (
     <>
-      <input data-testid={props.id} {...props} />
+      <input data-testid={props.id} ref={ inputRef } {...rest} />
```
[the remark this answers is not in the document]
haakemon: Ah, you are right. Sorry for the confusion :) 

The ID is still part of the `props` as you mention, so it is not lost.

src/shared/src/components/AltinnInput.tsx
haakemon: Ah, that part was not visible by default in the PR changes tab, so didnt notice that. Could very well be that this is covered by the grid parent.

src/altinn-app-frontend/src/components/summary/SummaryComponent.tsx
mjulstein: On line 155 i set the formComponent.type.toLowerCase().
I also removed case `Group` as it seems that having a case for every individual casing possible seems a bit messy to me.

src/altinn-app-frontend/src/components/base/InputComponent.tsx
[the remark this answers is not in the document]
mjulstein: I realized where you were coming from, and had to reference the id in order to have the destructor in the args.
Could probably rename the rest to something more fitting to the intent ie. htmlProps, but I think these problems are more important to cover when porting over to the design system.

src/shared/src/components/molecules/AltinnModal.tsx
mjulstein: In this case I'm not sure if I want to move so much at this point.
I will then have write the unit-tests first, so I know that all args are properly included in both branches.

src/altinn-app-frontend/src/components/summary/SummaryComponent.tsx
haakemon: Aha, I missed that part... But is it a good idea to change this? Then it will be out of sync with casing from formlayout.

An even better approach would be to use an enum for these instead I think. Then we would also get an error if something is misspelled.

src/shared/src/components/molecules/AltinnModal.tsx
haakemon: We also know what props `StyledModal` needs. But I agree its not really clear - this is one of the downsides of just spreading props like this :) 

Adding more tests to this is a good idea though, and that would make refactoring safer.

src/altinn-app-frontend/src/components/base/InputComponent.tsx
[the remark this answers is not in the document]
haakemon: I dont agree with that argument :) Why not change it now, and clean up the code immediately? Whats the reason to wait until this is ported to the design system? Then we have to remember to fix it then instead.

src/altinn-app-frontend/src/components/summary/InputSummaryBoilerplate.tsx
```diff
@@ -0,0 +1,21 @@
+import * as React from 'react';
+
+interface BoilerplateProps extends React.PropsWithChildren<any> {
+  formData: any;
+  setDisplayData: (v: any) => void;
+}
+
+export default function InputSummaryBoilerplate({ setDisplayData, formData, children }: BoilerplateProps) {
```
haakemon, line 8: This approach works fine, but I wanted to mention custom hooks as an alternative 🙂
(I wrote the hook directly here, so its not tested and may typos or other issues - so if you change to this, be sure to check that it works as expected 😅)

```jsx
const useDisplayData = ({formData}) => {
  const [displayData, setDisplayData] = React.useState('');

  React.useEffect(() => {
    if (formData && typeof formData === 'object') {
      let displayString = '';
      Object.keys(formData).forEach((key, index) => {
        displayString += `${index > 0 ? ' ' : ''}${formData[key]}`;
      });
      setDisplayData(displayString);
    } else {
      setDisplayData(formData);
    }
  }, [formData, setDisplayData]);

  return {displayData}
}
```

And usage would be
```jsx

function GroupInputSummary({ formData, label }: ISingleInputSummary) {
  const classes = useStyles();

  const {displayData} = useDisplayData({formData});

  return (
      <Typography variant='body1'>
        <span className={classes.label}>
          {label} {': '}
        </span>
        <span>{displayData}</span>
      </Typography>
    )
}
```

Custom hooks is a nice way to reuse logic across components that doesnt have rendered output. We avoid the wrapping components and having to pass down extra props or state handlers etc..

src/altinn-app-frontend/src/components/summary/SummaryWrapper.test.tsx
```diff
@@ -0,0 +1,38 @@
...
+}
+  test('should render the wrapper with div inside', () => {
+    render(<SummaryWrapper {...defaultProps}>
+      <div data-testid={'innermost'}>
+        <p>Something</p>
+      </div>
+    </SummaryWrapper>)
+    expect(screen.queryByRole('button')).not.toBeNull();
+    expect(screen.getByRole('heading')).toBeInTheDocument();
+    expect(screen.getByRole('button')).toBeInTheDocument();
```
haakemon, line 20: Isnt
```
expect(screen.queryByRole('button')).not.toBeNull();
```
and 
```
expect(screen.getByRole('button')).toBeInTheDocument();
```
the same thing?

src/altinn-app-frontend/src/components/summary/InputSummaryBoilerplate.tsx
mjulstein: Nice

src/altinn-app-frontend/src/components/summary/SummaryWrapper.test.tsx
mjulstein: Indeed they are checking for the same, I was thinking more in the lines of the other tests that use the queryByRole to verify the opposite. I guess line 20 would fail if 18 wasn't implicitly passed so I can take it away.

src/altinn-app-frontend/src/components/hooks/useDisplayData.ts
```diff
@@ -0,0 +1,17 @@
+import { useEffect, useState } from 'react';
+
+export const useDisplayData = ({ formData }: any) => {
```
haakemon, line 3: We should try to not introduce new `any` types - I think we have typings for this already (`IFormData`), can you fix? 🙂

haakemon (approved):
This task grew quite quickly from the small fix it was supposed to be 😅

But really great to get this cleaned up! Great work 🥇

haakemon:
Dont forget to bump version in package.json before merge

- [ ] bump package.json

PR #65, opened 2022-04-27
Bug/recipt superflous title submitted

olemartinorg:
## Description
Hiding titleSubmitted when no pdfs are passed, as per referenced issue

## Fixes
- Altinn/altinn-studio/issues/8296

## Verification
- [X] Relevant automated test added (if you find this hard, leave it and we'll help out)
- [X] All tests run green

## Documentation

### DIFF
--- a/src/shared/src/components/organisms/AltinnReceipt.test.tsx
+++ b/src/shared/src/components/organisms/AltinnReceipt.test.tsx
@@ -30,14 +30,30 @@ const attachment2 = {
 };
 
 describe('AltinnReceipt', () => {
-  it('should show titleSubmitted when set', () => {
+  it('should not show titleSubmitted when there are no pdfs', () => {
     render();
+
+    expect(
+      screen.queryByRole('heading', {
+        name: /titlesubmitted/i,
+      }),
+    ).not.toBeInTheDocument();
+
+    // No list of attachments expected
+    expect(screen.queryByRole('list')).not.toBeInTheDocument();
+  });
+
+  it('should show titleSubmitted when there are pdfs', () => {
+    render({ pdf: [{}] });
 
     expect(
       screen.getByRole('heading', {
         name: /titlesubmitted/i,
       }),
     ).toBeInTheDocument();
+
+    // A list of attachments expected
+    expect(screen.getByRole('list')).toBeInTheDocument();
   });
 
   it('should not show titleSubmitted when not set', () => {
--- a/src/shared/src/components/organisms/AltinnReceipt.test.tsx
+++ b/src/shared/src/components/organisms/AltinnReceipt.test.tsx
@@ -39,8 +39,7 @@ describe('AltinnReceipt', () => {
       }),
     ).not.toBeInTheDocument();
 
-    // No list of attachments expected
-    expect(screen.queryByRole('list')).not.toBeInTheDocument();
+    expect(screen.queryByTestId('attachment-list')).not.toBeInTheDocument();
   });
 
   it('should show titleSubmitted when there are pdfs', () => {
@@ -52,8 +51,7 @@ describe('AltinnReceipt', () => {
       }),
     ).toBeInTheDocument();
 
-    // A list of attachments expected
-    expect(screen.getByRole('list')).toBeInTheDocument();
+    expect(screen.getByTestId('attachment-list')).toBeInTheDocument();
   });
 
   it('should not show titleSubmitted when not set', () => {
--- a/src/shared/src/components/organisms/AltinnReceipt.tsx
+++ b/src/shared/src/components/organisms/AltinnReceipt.tsx
@@ -141,7 +141,7 @@ export function ReceiptComponent(props: IReceiptComponentProps) {
         >
           {props.body}
         </Typography>
-        {props.pdf && props.pdf.length && (
+        {props.pdf && props.pdf.length > 0 && (
           <React.Fragment>
               {props.titleSubmitted && (
               <Typography
--- a/src/shared/src/components/organisms/AltinnReceipt.tsx
+++ b/src/shared/src/components/organisms/AltinnReceipt.tsx
@@ -141,8 +141,8 @@ export function ReceiptComponent(props: IReceiptComponentProps) {
         >
           {props.body}
         </Typography>
-        {props.pdf && props.pdf.length
-          ? <React.Fragment>
+        {props.pdf && props.pdf.length && (
+          <React.Fragment>
               {props.titleSubmitted && (
               <Typography
                 variant='h3'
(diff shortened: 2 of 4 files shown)


src/shared/src/components/organisms/AltinnReceipt.test.tsx
```diff
@@ -30,14 +30,30 @@ const attachment2 = {
...
+  it('should not show titleSubmitted when there are no pdfs', () => {
     render();
 
+    expect(
+      screen.queryByRole('heading', {
+        name: /titlesubmitted/i,
+      }),
+    ).not.toBeInTheDocument();
+
+    // No list of attachments expected
```
haakemon, line 42: nitpick; this comment is maybe a bit redundant? 🙂

Same on line 55.

They say the same thing as the expect assertion, just with different words

src/shared/src/components/organisms/AltinnReceipt.tsx
```diff
@@ -141,22 +141,27 @@ export function ReceiptComponent(props: IReceiptComponentProps) {
...
+                {props.titleSubmitted}
+              </Typography>
+            )}
+            <AltinnAttachmentComponent
+              attachments={props.pdf}
+              id='attachment-list-pdf'
+            />
+          </React.Fragment>
+          : null
+        }
```
haakemon, line 164: nitpick; could keep using the `&&` here, to avoid the redundant `null` render. 

Not hugely important right now, but these inline conditionals tend to get quite complex over time and sometimes difficult to read. Could extract the conditional statement for easier readability, and/or change to optional chaining. 

```jsx

{props.pdf?.length && (
//...


// or before return
const hasAttachments= props.pdf?.length;

{hasAttachments && (
//...

```

```suggestion
        {props.pdf && props.pdf.length && (
          <React.Fragment>
              {props.titleSubmitted && (
              <Typography
                variant='h3'
                style={{
                  paddingTop: '4.1rem',
                  paddingBottom: '0.5rem',
                  fontWeight: 600,
                }}
              >
                {props.titleSubmitted}
              </Typography>
            )}
            <AltinnAttachmentComponent
              attachments={props.pdf}
              id='attachment-list-pdf'
            />
          </React.Fragment>
          )
        }
```

haakemon (commented):
Nice stuff 🙂 Left a couple of minor comments

src/shared/src/components/organisms/AltinnReceipt.test.tsx
olemartinorg: I disagree somewhat, but I also expected this comment.. ;-) I think I would find it non-obvious this is meant to look for the attachment list, as there could be any amount of lists in the receipt component (there is already a table, which is in itself a kind of listing). I looked at the query documentation in order to construct a more specific query, but looking for the specific dom-id was apparently advised against. I'd prefer one comment too many rather than having to look through history for the original intention. I might represent a minority on this though.

src/shared/src/components/organisms/AltinnReceipt.tsx
olemartinorg: Nice! :1st_place_medal: Not used to doing it that way, but I'll make a note of this!

ivarne: Double nitpick:
`{array.length && renderContent()}` and `{array.length ? renderContent() : null}` is slightly different in that the first will render `0` when the array is empty, but the second will render nothing. You need to write `{array.length>0 && renderContent()}` in order for the first part of the `&&` expression to be false.

src/shared/src/components/organisms/AltinnReceipt.test.tsx
haakemon: I would also find it non-obvious that this comment is talking about a specific list (even though "attachments" is mentioned). A list is a list🙂 Its not easy to understand that underlying info just by reading this (which is what happened to me, and so I wrote the initial comment😅 ). So I would argue that it is still redundant - the comment and the expect has the same meaning for someone who doesnt know the context.

There is a "safeguard" here though in the `queryByRole` function. If there are more than 1 returned items, it will throw an error. This will not cover the case where the `list` is replaced with another incorrect list though (f.ex which contains something other than PDF's).

We could alternatively change this to `queryByTestId` f.ex, and attach a more specific `data-testid` attribute to the element we are looking for. Then we can still change the implementation without updating the test (but would have to remember to keep the data attribute)

src/shared/src/components/organisms/AltinnReceipt.tsx
olemartinorg: Fantastic catch, @ivarne! Fixed now.

ivarne: I used to write `&&`, but have since been convinced to do `{cond?render:null}` instead, probably because of this issue.

mjulstein (approved):
Nice

src/shared/src/components/organisms/AltinnReceipt.tsx
```diff
@@ -141,22 +141,27 @@ export function ReceiptComponent(props: IReceiptComponentProps) {
...
-          >
-            {props.titleSubmitted}
-          </Typography>
-        )}
-        <AltinnAttachmentComponent
-          attachments={props.pdf}
-          id='attachment-list-pdf'
-        />
+        {props.pdf && props.pdf.length > 0 && (
+          <React.Fragment>
```
lorang92, line 145:
```suggestion
          <>
```

src/shared/src/components/organisms/AltinnReceipt.tsx
```diff
@@ -141,22 +141,27 @@ export function ReceiptComponent(props: IReceiptComponentProps) {
...
+                }}
+              >
+                {props.titleSubmitted}
+              </Typography>
+            )}
+            <AltinnAttachmentComponent
+              attachments={props.pdf}
+              id='attachment-list-pdf'
+            />
+          </React.Fragment>
```
lorang92, line 162:
```suggestion
          </>
```

lorang92 (approved):
Good stuff 💯 

Minor nitpicking I just find more pleasing for the eyes. Merge only if you agree.

src/shared/src/components/organisms/AltinnReceipt.tsx
olemartinorg: Neat trick! Didn't know about this, but it certainly looks nicer. :+1:

lorang92 (commented):
Oh! And another annoying thing. These `AltinnReceipt` and `AltinnAttachment`-components are currently duplicated in our [altinn-stdio](https://github.com/Altinn/altinn-studio/tree/master/src/Altinn.Apps/AppFrontend/react/shared/src) repo - which are used by the receipt platform component.

I think we need to duplicate this PR to also get this fix out for our receipt component...

src/shared/src/components/organisms/AltinnReceipt.tsx
haakemon: Well, thats a bit surprising to me, but now I know. Thanks @ivarne 😁

Issue #55, opened 2022-01-13
Support for dynamic rule configuration in repeating groups that are nested


acn-dgopa:
## What needs to be solved?
It should be possible to configure dynamic rules for components inside nested repeating groups.

## How do you want it solved?
-  Configure the repeating group in the rule configuration with the nested group id and the input value configuration similar to existing dynamic
      "7831f53f-7ede-4b27-941e-2643a92c1713": {
        "selectedFunction": "showWhenPersonIsSelected",
        "inputParams": {
          "value": "A{0}.B{0}.X"
        },
        "selectedAction": "Show",
        "selectedFields": {
          "56edae00-478d-482a-8443-ec285c6a55f4": "B-dropdownY{0}"
        },
        "repeatingGroup": {
          "groupId": "A",
           "nestedGroupId": "B"
        }
      },

lorang92:
@acn-dgopa this feature has now been implemented. Would you mind taking a look and verifying the need? 
See related docs: https://docs.altinn.studio/nb/app/development/logic/dynamic/#n%C3%B8stede-repeterende-grupper

acn-dgopa:
@RogerKjaernsrod  @ingunnjv FYI

RogerKjaernsrod:
Testet ok for mt-3 (Økokrim)

lorang92:
Closing issue as done.

PR #70, opened 2022-05-02
feat: dynamics in nested groups

lorang92:
## Description
Now supports dynamics in nested groups.

## Fixes
- #55 

## Verification
- [x] All tests run green

## Documentation

### DIFF
--- a/src/altinn-app-frontend/src/features/form/dynamics/types/index.ts
+++ b/src/altinn-app-frontend/src/features/form/dynamics/types/index.ts
@@ -12,6 +12,7 @@ export interface IConditionalRenderingRule {
 
 export interface IConditionalRenderingRepeatingGroup {
   groupId: string;
+  childGroupId?: string;
 }
 
 export interface IParameters {
--- a/src/altinn-app-frontend/package.json
+++ b/src/altinn-app-frontend/package.json
@@ -1,6 +1,6 @@
 {
   "name": "altinn-app-frontend",
-  "version": "3.35.2",
+  "version": "3.36.0",
   "description": "",
   "main": "index.js",
   "scripts": {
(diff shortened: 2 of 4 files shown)


src/altinn-app-frontend/src/features/form/dynamics/types/index.ts
```diff
@@ -12,6 +12,7 @@ export interface IConditionalRenderingRule {
 
 export interface IConditionalRenderingRepeatingGroup {
   groupId: string;
+  childGroupId?: string;
```
nkylstad, line 15: Does the app developer have to specify this in the configuration?

nkylstad (approved):
Looks great! 🍰

src/altinn-app-frontend/src/features/form/dynamics/types/index.ts
lorang92: Yeah, I just extended the existing setup.

But it is a point that we could potentially just fetch the group id's by manipulating the `inputParams` and looking for the group with the matching `group`-databinding in the layout files. 

Given the example input-from the test:

```
        inputParams: {
          number: 'parentGroup{0}.childGroup{1}.mockField',
        },
```

We could do some manipulation here and extract `parentGroup` and `parentGroup.childGroup` strings and find the matching group-component from the layout.

This would maybe be smoother for the app-developer in one sense. But they still would have to define these index indicators manually - so I'm not really sure how much value it would add. The group-id's are probably the easiest thing about this configuration. Could also be made smoother if we have a better GUI in studio to make this configuration a little nicer.

Any thoughts? Worth looking into?

nkylstad: I guess my worry is that the config for dynamics is already quite complicated. But, this is in part because the GUI in studio isn't great, and a lot has to be done manually.
I think that improving the GUI in studio is something that we want to do anyway. Auto-populating the group/child group fields is something we might do to make app developer experience simpler, so I think we can leave this as is for now.

PR #80, opened 2022-05-06
Bump version

mjulstein:
Bump version for issue #8

## Description
Forgot to update the version number before merging #8 

## Fixes
- #8

### DIFF
--- a/src/altinn-app-frontend/package.json
+++ b/src/altinn-app-frontend/package.json
@@ -1,6 +1,6 @@
 {
   "name": "altinn-app-frontend",
-  "version": "3.36.0",
+  "version": "3.36.1",
   "description": "",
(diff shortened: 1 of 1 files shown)


lorang92 (commented):
I guess we can bump now, but the previous version has already been overwritten with new changes. I think this bump actually accomplishes little. Maybe we should just update the release log for `3.36.0` so that it includes both changes, and drop bumping after the fact? 

Apps fetch the latest minor versions by default, so it's no crisis. Manual processes are error-prone :)  *

WDYT?

olemartinorg:
3 seconds after discussing this on slack, I managed to push the 'squash and merge' button where I also bumped to 3.36.1.. So I guess this PR is redundant now?

PR #83, opened 2022-05-06
Application language support

ZuperNova:
Have a look at the main PR: https://github.com/Altinn/altinn-studio/pull/8483


src/altinn-app-frontend/src/components/base/DropdownComponent.tsx
```diff
@@ -89,26 +69,23 @@ function DropdownComponent({
       {fetchingOptions ? (
         <AltinnSpinner />
       ) : (
-        <select
+        <AltinnDropdown
```
mjulstein, line 72: `AltinnDropdown` has been deleted from the main branch as it was not in use.
What value did switching from a native `select` to the `AltinnDropdown` give?
Is it essential for this change to have the expected effect?

lasseklovstad: The AltinnDropdown component is rewritten in this PR, its not the same as you deleted in main branch. Since the AltinDropdown wasn't used, we overwrote the old implemntation with the one that was in DropDownComponent.tsx and reused it for our language selector. We were not sure where to place this new reusable select component, so we just placed it where the old one was.

mjulstein: I see. Maybe we should give this new component a new name that is more descriptive of that it is trying to solve?
I think we are starting to move away form the `Altinn` prefix as it's not really helpful.

src/shared/src/components/AltinnDropdown.tsx
```diff
@@ -0,0 +1,73 @@
...
+  options,
+  value,
+  disabled = false,
+  error = false,
+  fullWidth = false,
+  className,
+}: IAltinnDropdownProps) => {
+  const classes = useStyles();
+  return (
+    <React.Fragment>
```
mjulstein, line 44: The `React.Fragment` element is redundant here, as there as the `select` element is a single element already.

src/altinn-app-frontend/src/components/base/DropdownComponent.tsx
lasseklovstad: So maybe just call it Select or Dropdown? Should we move to another folder or keep it were it is for now?

mjulstein: So sorry for keeping you waiting.
Select seems fine :). The folder is fine, as we are going to move over to the design-system eventually.

src/altinn-app-frontend/__mocks__/initialStateMock.ts
```diff
@@ -78,19 +80,14 @@ export function getInitialStateMock(customStates?: Partial<IRuntimeState>): IRun
           logo: '',
           orgnr: '',
           homepage: '',
-          environments: [
-            'tt02',
-            'production',
-          ],
+          environments: ['tt02', 'production'],
```
mjulstein, line 83: Is this file changed or just re-formatted?

src/altinn-app-frontend/src/components/base/DropdownComponent.tsx
```diff
@@ -1,15 +1,9 @@
...
-
-import { AltinnAppTheme } from 'altinn-shared/theme';
 import { useAppSelector, useHasChangedIgnoreUndefined } from 'src/common/hooks';
 import { IComponentProps } from '..';
-
-import '../../styles/shared.css';
 import type { IMapping, IOptionSource } from 'src/types';
 import { getOptionLookupKey } from 'src/utils/options';
-import { AltinnSpinner } from 'altinn-shared/components';
+import { AltinnSpinner, Select } from 'altinn-shared/components';
```
mjulstein, line 6: ❤️

src/altinn-app-frontend/src/components/presentation/NavBar.test.tsx
```diff
@@ -1,26 +1,70 @@
-import { mount } from 'enzyme';
 import React from 'react';
-import NavBar from './NavBar';
+import NavBar, { INavBarProps } from './NavBar';
+import { render, screen } from '@testing-library/react';
```
mjulstein, line 3: Nice update here :) 👍

src/altinn-app-frontend/src/components/presentation/NavBar.test.tsx
```diff
@@ -1,26 +1,70 @@
...
-    expect(wrapper.find('button.a-modal-close')).toHaveLength(0);
+  it('should render back button', async () => {
+    const { mockBack } = renderNavBar({ showBackArrow: true });
+    const backButton = screen.getByRole('button', { name: /Tilbake/i });
+    await userEvent.click(backButton);
+    expect(mockBack).toHaveBeenCalled();
+  });
+  it('should not render back button', async () => {
+    renderNavBar();
+    expect(screen.queryByRole('button', { name: /Tilbake/i })).toBeNull();
```
mjulstein, line 56: Could this query be less explicit?
It doesn't really protect against if there for instance is a button without  ´/Tilbake/i´ in the name prop.
If I were to change anything with the button this would still pass.

src/altinn-app-frontend/src/components/presentation/NavBar.test.tsx
```diff
@@ -1,26 +1,70 @@
...
+    const { mockAppLanguageChange } = renderNavBar({
+      showLanguageSelector: true,
+    });
+    const dropdown = screen.getByRole('combobox', { name: /Språk/i });
+    await userEvent.selectOptions(dropdown, 'English');
+    expect(mockAppLanguageChange).toHaveBeenCalledWith('en');
+  });
+  it('should not render app language combobox', async () => {
+    renderNavBar();
+    expect(screen.queryByRole('combobox', { name: /Språk/i })).toBeNull();
```
mjulstein, line 68: This test is also a bit too explicit.
We might also go away from having this text localised and only use the English ('Language') version.
If I do not understand the language on the page in the first place, and in need to change the language, I would probably not understand the word for language in that language either.

src/altinn-app-frontend/src/components/presentation/NavBar.tsx
```diff
@@ -1,49 +1,107 @@
...
+            type='button'
+            className='a-modal-back a-js-tabable-popover'
+            aria-label={getLanguageFromKey('general.back', props.language)}
+            onClick={props.handleBack}
+          >
+            <span className='ai-stack'>
+              <i className='ai-stack-1x ai ai-back' aria-hidden='true' />
+            </span>
+            <span className='hidden-button-text'>
+              {getLanguageFromKey('general.back', props.language)}
```
mjulstein, line 66: I did a little digging and noticed that the `hidden-button-text` class above only has a `display: none` rule, making it completely irrelevant, as it's never rendered.
We are moving away form using global stylesheets and using component based styles, but for now you can switch alle the cases with `hidden-button-text` to `sr-only` as it is also defined in the old design-system and has a more screen-reader-friendly definition.

src/altinn-app-frontend/__mocks__/initialStateMock.ts
lasseklovstad: It is changed. On line 130 we added the appLanguagestate to the mock

src/altinn-app-frontend/src/components/presentation/NavBar.test.tsx
lasseklovstad: Would it be better if I define a variable `const backButtonNamRegex = /Tilbake/i` at the top of the test file and then use this variable in all the Role queries?

Or if you have any other suggestions, I would be glad to hear :D

src/altinn-app-frontend/src/shared/resources/appLanguage/appLanguageReducer.ts
```diff
@@ -0,0 +1,69 @@
...
+  IFetchAppLanguagesFulfilled,
+  IFetchAppLanguagesRejected,
+  IUpdateAppLanguages,
+} from './fetch/fetchAppLanguagesActions';
+import * as ActionTypes from './fetch/fetchAppLanguagesActionTypes';
+
+import { IAltinnWindow } from 'src/types';
+import { IAppLanguage } from 'altinn-shared/types';
+const altinnWindow = window as Window as IAltinnWindow;
+const { app } = altinnWindow;
```
mjulstein, line 13: Would it be possible to get the language key part from the URI instead? `app` could possibly not be defined in a unit-test for instance.

src/altinn-app-frontend/src/shared/resources/appLanguage/fetch/fetchAppLanguagesSagas.ts
```diff
@@ -0,0 +1,24 @@
...
+    const resource = yield call(get, appLanguagesUrl())
+    yield call(AppLanguagesActions.fetchAppLanguagesFulfilled, resource);
+  } catch (error) {
+    yield call(AppLanguagesActions.fetchAppLanguagesRejected, error);
+    yield put(appTaskQueueError({ error }));
+  }
+}
+
+export function* watchFetchAppLanguagesSaga(): SagaIterator {
+  yield all([
```
mjulstein, line 20: The `take` alone should be enough?

src/altinn-app-frontend/src/components/presentation/NavBar.test.tsx
mjulstein: What I mean is that you are checking for the button to not exist. Therefore the button-text should probably not be part of the query.
Maybe even counting the amount of buttons, would be a more useful check here?
If the base language changed the tests would still pass, because there never would be a button that matched the query.
The test in it self is not really protecting against any changes.

src/shared/src/components/Select.tsx
```diff
@@ -0,0 +1,71 @@
...
+            'validation-error': error,
+            'disabled !important': disabled,
+          },
+          className,
+        )}
+        style={{ width: fullWidth ? '100%' : 'unset' }}
+        onChange={onChange}
+        onBlur={onBlur}
+      >
+        <option style={optionStyle} />
```
mjulstein, line 61: why is there a hidden first option here?

src/shared/src/components/Select.tsx
```diff
@@ -0,0 +1,71 @@
...
+        {options.map((option, index) => (
+          <option key={index} value={option.value}>
```
mjulstein, line 63: Using index as key isn't recommended.
You could use the `option.value` or `option.label` instead.
It would make sense as a select with two equal values would be quite confusing.
WDYT?
If you keep the firs option, on line 61, you could give it a key as well. Maybe something describing intent?

src/shared/src/components/Select.tsx
```diff
@@ -0,0 +1,71 @@
...
+}: ISelectProps) => {
+  const classes = useStyles();
+  return (
+      <select
+        id={id}
+        value={value}
+        disabled={disabled}
+        className={cn(
+          classes.select,
+          'custom-select a-custom-select',
```
mjulstein, line 50: These classes seem to be only defined in the old design-system.
Maybe you could remove them and just add `width: '100%'` in the select class directly?

src/shared/src/components/Select.tsx
```diff
@@ -0,0 +1,71 @@
...
+          {
+            'validation-error': error,
+            'disabled !important': disabled,
```
mjulstein, line 53: does this work?

src/shared/src/components/Select.tsx
lasseklovstad: I just copied this implementation from the `DropDownComponent.tsx`. If the value for the dropdown is empty this hidden option will be displayed. If the hidden option is removed the select will display the first option in the list. 

If we make the option visible, we are able to select an empty value after we have selected something.

This is an issue btw: https://github.com/Altinn/app-frontend-react/issues/17

src/shared/src/components/Select.tsx
lasseklovstad: jupp, option.value is better

src/shared/src/components/Select.tsx
lasseklovstad: This is how the select was implemented in DropDownComponent.tsx before i refactored the code into this Select component. If I remove these classes the dropdown looses its altinn look. 
image
to this: 
image

src/shared/src/components/Select.tsx
lasseklovstad: This is also something I copyd the and does actually work. The Select looks like this: 
image

mjulstein (commented):
Code looks good to me. @lorang92 @RonnyB71  are there some things regarding versioning or other repos that need to be approved?
@lasseklov is all this opt in, does it require any configuration to keep current behavior in other apps?

lasseklovstad:
> Code looks good to me. @lorang92 @RonnyB71 are there some things regarding versioning or other repos that need to be approved? @lasseklov is all this opt in, does it require any configuration to keep current behavior in other apps?

If you try to run this frontend with an altinn app that does not have the new nugget packages it will fail. This is because of the new endpoint does not exist: 
image

Should we change the behavior to only call this endpoint if showLanguageSelector is set to true and is present? So it won't fail :)

src/altinn-app-frontend/src/components/presentation/NavBar.test.tsx
```diff
@@ -1,26 +1,69 @@
...
-    const wrapper = mount(
-      <NavBar language={{}} handleClose={null} handleBack={null} />,
-    );
-    expect(wrapper.find('div.a-modal-navbar')).toHaveLength(1);
-    expect(wrapper.find('button.a-modal-close')).toHaveLength(1);
+  it('should render close button', async () => {
+    const { mockClose } = renderNavBar();
+    const closeButton = screen.getByRole('button', { name: /Lukk Skjema/i });
+    // Should not render back button
+    expect(screen.queryAllByRole('button')).toHaveLength(1);
```
mjulstein, line 41: This check isn't cohesive as the context is rendering the close button.
Good thing that you have a comment on top, but I would extract it into a separate test.
I would consider using the `data-testid='uniqueId'` attribute on the component.
That way you actually can check for it with a query.
Not that big of a deal though.

lasseklovstad: I changed to test id for the back button and updated the app language saga to fetch only if showLanguageSelector exists and is set to true in your layoutsettingsfile. The frontend will work as normal (not displaying an error) if you havent configured app language yet.

mjulstein:
> 

Please do :)
Anything that can prevent others from requiring updates are very welcome.

src/altinn-app-frontend/src/shared/resources/appLanguage/fetch/fetchAppLanguagesSagas.ts
```diff
@@ -0,0 +1,22 @@
...
+import { appTaskQueueError } from '../../queue/queueSlice';
+import { FETCH_APP_LANGUAGES } from './fetchAppLanguagesActionTypes';
+
+function* fetchAppLanguages(): SagaIterator {
+  try {
+    const resource = yield call(get, appLanguagesUrl());
+    yield call(AppLanguagesActions.fetchAppLanguagesFulfilled, resource);
+  } catch (error) {
+    yield call(AppLanguagesActions.fetchAppLanguagesRejected, error);
+    yield put(appTaskQueueError({ error }));
```
mjulstein, line 21: Maybe instead of crashing the site (as I expect happens by putting this error) you can stop the combobox from being instantiated?
That way language support can be toggled based on actual availability.

src/altinn-app-frontend/src/components/presentation/NavBar.test.tsx
```diff
@@ -1,26 +1,74 @@
...
-  it('should render back button by default', () => {
-    const wrapper = mount(
-      <NavBar language={{}} handleClose={null} handleBack={null} />,
-    );
-    expect(wrapper.find('div.a-modal-navbar')).toHaveLength(1);
-    expect(wrapper.find('button.a-modal-close')).toHaveLength(1);
+  it('should render close button', async () => {
+    const { mockClose } = renderNavBar();
+    const closeButton = screen.getByRole('button', { name: /Lukk Skjema/i });
+    // Should not render back button
```
haakemon, line 40: I'm probably misunderstanding, or missing some context, but I dont understand this comment. It mentions a "back button", but the test title is mentioning a "close button". Are they the same thing? Theres nothing in the test that verifies that a "back button" is not rendered?

lasseklovstad: Well, the comment points out that since line 39 checks that the close button is rendered and line 41 checks that there is in total 1 button, this means that the back button is not rendered (or that the close button is the only "button" rendered). Maybe not relevent to point out.

src/altinn-app-frontend/src/shared/resources/appLanguage/fetch/fetchAppLanguagesSagas.ts
```diff
@@ -0,0 +1,25 @@
...
+import { call, select, takeLatest } from 'redux-saga/effects';
+import { get } from 'src/utils/networking';
+import { appLanguagesUrl } from 'src/utils/appUrlHelper';
+import AppLanguagesActions from '../appLanguagesActions';
+import { IRuntimeState } from 'src/types';
+import { FormLayoutActions as Actions } from 'src/features/form/layout/formLayoutSlice';
+
+function* fetchAppLanguages(): SagaIterator {
+  try {
+    const showLanguageSelector = yield select(
```
mjulstein, line 11: Could this done where you forked the watcher instead (see other comment)?

src/altinn-app-frontend/src/sagas/index.ts
```diff
@@ -32,6 +33,7 @@ function* root(): SagaIterator {
   yield fork(FormDataModelSagas);
   yield fork(LanguageSagas);
   yield fork(TextResourcesSagas);
+  yield fork(AppLanguagesSagas)
```
mjulstein, line 36: Is it possible to check the config for the language selector here instead?
That way the app won't need to have a watcher sitting around doing nothing.
(I'm not 100% sure what's available in a given time during runtime.)
Also if `AppLanguageSagas` are very different form `LanguageSagas`. on line 34, maybe they should be named more explicitly?
Maybe `LanguageSelectorSagas`? And could it maybe be moved into a branch inside of `LanguageSagas`?

The config check could also alternatively be done inside the watcher, to keep this file free from tests.
Is this

src/altinn-app-frontend/src/shared/resources/appLanguage/fetch/fetchAppLanguagesSagas.ts
```diff
@@ -0,0 +1,25 @@
...
+      yield call(AppLanguagesActions.fetchAppLanguagesFulfilled, resource);
+    }
+  } catch (error) {
+    yield call(AppLanguagesActions.fetchAppLanguagesRejected, error);
+  }
+}
+
+export function* watchFetchAppLanguagesSaga(): SagaIterator {
+  yield takeLatest(Actions.fetchLayoutSettingsFulfilled, fetchAppLanguages);
+}
```
mjulstein, line 25: Optional location to check the config.

src/altinn-app-frontend/src/sagas/index.ts
lasseklovstad: Im not sure if this is possible. Since I want it to run it every time the Layout file changes. Becuase the Stateless app has different config file from the Statefull app, we need to listen to the change in config.

I am not that used to redux saga, so if you have suggestions for what could work I am open to it :P

ZuperNova:
@mjulstein Yesterday I did some testing and now I'm facing an issue that I don't seem to figure out. Maybe you know how to fix this or maybe you could give me some pointers on how to?

So the issue is related to this translation implementation. When changing the language, the text resource is fetched for that chosen language. But if the text resource contain any variables the interface don't seem to pick this up and re-render the variable in the text. So, the text itself is translated (fetched & rendered), but the variables do not.

Please have a look at the images below. Here you can see that the application is started in norwegian bokmål, but when I change the language to english the variable is no longer properly rendered.

image
image

ZuperNova:
> > @mjulstein Yesterday I did some testing and now I'm facing an issue that I don't seem to figure out. Maybe you know how to fix this or maybe you could give me some pointers on how to?
> > So the issue is related to this translation implementation. When changing the language, the text resource is fetched for that chosen language. But if the text resource contain any variables the interface don't seem to pick this up and re-render the variable in the text. So, the text itself is translated (fetched & rendered), but the variables do not.
> > Please have a look at the images below. Here you can see that the application is started in norwegian bokmål, but when I change the language to english the variable is no longer properly rendered.
> > image image
> 
> @ZuperNova this seem to be related to how we run `replaceTextResourcesSaga`, which is currently only replacing variables on the first language fetch. But we should also update on subsequent fetching of language.
> 
> I guess this would be so simple as to spawn the saga on new fetches of text resources, simply by adding ` yield takeLatest(FETCH_TEXT_RESOURCES_FULFILLED, replaceTextResourcesSaga);` in a new line here: https://github.com/Altinn/app-frontend-react/blob/main/src/altinn-app-frontend/src/shared/resources/textResources/replace/replaceTextResourcesSagas.ts#L62
> 
> Would you mind having look? :)

Thanks @lorang92. I did actually try this, but it is still not working. What I find interesting about this is that the redux-devtools show that the text resource does contain the correct variable, but the actual web page does not. And another strange thing I noticed is when I minimize the browser and pull it up again, the variable is rendered correctly.

image
image

lorang92:
@ZuperNova weird indeed. I did a little digging. It appears to me that we have a weakness in our code that might be the underlying problem you are running in to. It appears that the `replaceTextResourceParams` util method that is called in the saga actually mutates the object and not creating a new one. Which is bad.. This should be fixed (I guess this can be handled as a seperate issue than what you are working on).

I have a hacky workaround you could try until we get this fixed. Could you try creating a new object that is passed to the redux store when text params are replaced?

So, add:
`yield call(TextResourceActions.replaceTextResourcesFulfilled, textResources.language, JSON.parse(JSON.stringify(updatedTextsResources)));` to the line https://github.com/Altinn/app-frontend-react/blob/main/src/altinn-app-frontend/src/shared/resources/textResources/replace/replaceTextResourcesSagas.ts#L45

Could you check if this actually solves the problem?

ZuperNova:
> @ZuperNova weird indeed. I did a little digging. It appears to me that we have a weakness in our code that might be the underlying problem you are running in to. It appears that the `replaceTextResourceParams` util method that is called in the saga actually mutates the object and not creating a new one. Which is bad.. This should be fixed (I guess this can be handled as a seperate issue than what you are working on).
> 
> I have a hacky workaround you could try until we get this fixed. Could you try creating a new object that is passed to the redux store when text params are replaced?
> 
> So, add: `yield call(TextResourceActions.replaceTextResourcesFulfilled, textResources.language, JSON.parse(JSON.stringify(updatedTextsResources)));` to the line https://github.com/Altinn/app-frontend-react/blob/main/src/altinn-app-frontend/src/shared/resources/textResources/replace/replaceTextResourcesSagas.ts#L45
> 
> Could you check if this actually solves the problem? Needs to be combined with the earlier mentioned change.

@lorang92 It works like a charm! 👍 Do you want me to include the workaround in this PR?

ZuperNova:
> Nice 🥳 I think that is fine for now. Ill create an issue to fix the underlying issue, I'm a little unsure how much work this is. This is something that can be solved as a seperate thing 👍

Great! I have now pushed the workaround 📦 🚀

lorang92:
@ZuperNova I have created a PR that should resolve this issue altogether: https://github.com/Altinn/app-frontend-react/pull/186. 

Hope to get this merged pretty fast, then you should not experience this issue anymore and the workaround can be removed.

lorang92:
@ZuperNova the issue discussed earlier should now be fixed. Would you mind merging with master and verifying that removing the workaround now works as expected? :)

ZuperNova:
> @ZuperNova the issue discussed earlier should now be fixed. Would you mind merging with master and verifying that removing the workaround now works as expected? :)

Nice! 😄 I tested it and it works as expected. Reverted the workaround

lorang92 (changes requested):
First of all, thank you so much for your contribution.

I have left some comments that require some refactoring to a more modern "stack". This will simplify the code and is more in sync with the way we want to move forward, and avoid introducing technical debt. 

I'm really sorry for coming with these comments so late, this should have been caught a little earlier... This would also be something we could have discussed if we had a little chat before we started development. Lessons learned I guess 🙃 .

Please let me know if anything is unclear, we are more than willing to give pointers and help out if needed. Also always open for a little chat on slack if you need quick feedback!

src/altinn-app-frontend/src/shared/resources/appLanguage/appLanguageReducer.ts
```diff
@@ -0,0 +1,69 @@
+import update from 'immutability-helper';
```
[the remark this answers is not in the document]
lorang92: Yeah, please do :) In the same manner as the InstancesApi 👍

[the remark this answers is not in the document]
lasseklovstad: Okei thx!

src/altinn-app-frontend/src/components/presentation/NavBar.test.tsx
haakemon: I see. Could I suggest to instead add an extra assertion for this instead of the comment?

Something like this:

```suggestion
    expect(screen.queryByTestId('altinn-back-button')).not.toBeInTheDocument();
```

This way its explicitly part of the test, and not a comment that someone may or may not read, and have to interpret (and misunderstand like I did) 🙂

lasseklovstad:
> First of all, thank you so much for your contribution.
> 
> I have left some comments that require some refactoring to a more modern "stack". This will simplify the code and is more in sync with the way we want to move forward, and avoid introducing technical debt.
> 
> I'm really sorry for coming with these comments so late, this should have been caught a little earlier... This would also be something we could have discussed if we had a little chat before we started development. Lessons learned I guess 🙃 .
> 
> Please let me know if anything is unclear, we are more than willing to give pointers and help out if needed. Also always open for a little chat on slack if you need quick feedbac

> First of all, thank you so much for your contribution.
> 
> I have left some comments that require some refactoring to a more modern "stack". This will simplify the code and is more in sync with the way we want to move forward, and avoid introducing technical debt.
> 
> I'm really sorry for coming with these comments so late, this should have been caught a little earlier... This would also be something we could have discussed if we had a little chat before we started development. Lessons learned I guess 🙃 .
> 
> Please let me know if anything is unclear, we are more than willing to give pointers and help out if needed. Also always open for a little chat on slack if you need quick feedback!

@lorang92 It should be implemented with redux toolkit query as you asked and refactered all the tests to work. I decided to extend the language saga and not create a new appLanguageSaga, for the selectedAppLanguageState. 

I appreciate it if you could take another look 😄

src/altinn-app-frontend/src/services/AppLanguageApi.tsx
```diff
@@ -0,0 +1,15 @@
+import { appApi } from 'src/services/AppApi';
+import { IAppLanguage } from 'altinn-shared/types';
+
+export const appLanguageApi = appApi.injectEndpoints({
```
lorang92, line 4: Rename to a little more generic `LanguageApi`

src/altinn-app-frontend/src/services/AppLanguageApi.tsx
```diff
@@ -0,0 +1,15 @@
...
+  endpoints: (builder) => ({
+    getAppLanguage: builder.mutation<IAppLanguage[], void>({
```
lorang92, line 6: This does not have to be a mutation as we simply fetch data and never really mutates anything. Can simply be a query 🥳  See docs: https://redux-toolkit.js.org/rtk-query/usage/queries

lorang92 (commented):
@lasseklovstad starting to look really good. Hope you agree that the refactoring made it a little cleaner 😎 . Left a couple of comments 👍

lorang92:
@lasseklov @ZuperNova sorry for the late reply, we have been running a three day workshop the last couple of days so this has been put on hold for a little while.

The code looks good to me now! Would you mind merging with main so that we can run the cypress tests? (had a misconfigured workflow).

When that passes will start some internal testing and hopefully get this merged pretty soon 🥳 Great job!

RonnyB71:
@lasseklov @ZuperNova I picked up the testing again after the latest changes in this branch - are there any changes with regards to how this is configured/set up ref. our previous discussion here: https://github.com/Altinn/altinn-studio/pull/8483. The reason why I asked is that I cheked out this branch and now the language selector is gone and no requests are done against the backend for application languages.

ZuperNova:
> @lasseklov @ZuperNova I picked up the testing again after the latest changes in this branch - are there any changes with regards to how this is configured/set up ref. our previous discussion here: [Altinn/altinn-studio#8483](https://github.com/Altinn/altinn-studio/pull/8483). The reason why I asked is that I cheked out this branch and now the language selector is gone and no requests are done against the backend for application languages.
> 
> My test application is still the same with regards to the setup.

@RonnyB71 I just tested it myself and it works as expected. There has not been any changes to the configuration/setup since last time you tested. Maybe a silly suggestion, but did you run `yarn --immutable`?

RonnyB71:
> @RonnyB71 I just tested it myself and it works as expected. There has not been any changes to the configuration/setup since last time you tested. ~Maybe a silly suggestion, but did you run `yarn --immutable`?~ It is just the frontend that has been updated.
> 
No frontend questions are silly when you talk to a backend developer:) But yes, I did run `yarn --immutable`

> Is it possible to get a link to your test app, so that I could have a look?
It's running in our dev environment and locallly on my machine, but I'll give it another go and check the setup. If I'm still struggling maybe I can give you a ping on Slack and we can take a look together?

RonnyB71:
Figured it out. When I checked out your changes yesterday I ran into some git issues and I did a reset, apparently not all changes where checked out after the reset, so I ended up in some unknown state. New round and things worked out fine:)

ZuperNova:
> Figured it out. When I checked out your changes yesterday I ran into some git issues and I did a reset, apparently not all changes where checked out after the reset, so I ended up in some unknown state. New round and things worked out fine:)

Great! Please let me know if there is anything that you find during your internal testing

lorang92 (approved):
Internal testing OK 🥳 Merging and creating a release. Should be published in a couple of minutes.

Great work guys, looks fantastic.

PR #84, opened 2022-05-09
Fixing flaky options test

olemartinorg:
## Description
The `options` cypress test was flaky. The root cause of this was assumed to be the way `preselectedOptionIndex` is implemented, as the first dropdown component (`source`) is first rendered once before it figures out it should be pre-selecting an option. The component will then set its data and render again shortly after. When this field value is then a dependency for a second dropdown with a list of options (`reference`), the list of options would first get fetched with `source=undefined` (during the first render) and then right after it would fetch options with `source=something`. Only then can the second dropdown (queried in this test) contain the expected option.

For cypress, this meant that it would find the element it was looking for (and verify it was visible), but then try to select an option before the XHR had finished fetching relevant options. Even if we fixed the minor issue where there is a first round of rendering (and a useless XHR to fetch options) before the correct `preselectedOptionIndex` is chosen, the test would still be flaky as it would have had to wait for the XHR to finish before the expected option was available.

The solution here is [in the cypress docs](https://docs.cypress.io/guides/core-concepts/retry-ability#Commands-vs-assertions); we have to make sure cypress waits until the expected option is found.

## Fixes
- Altinn/altinn-studio/issues/8255

## Verification
- [X] Relevant automated test added (if you find this hard, leave it and we'll help out)
- [X] All tests run green

## Documentation
- ~~[ ] Changelog is updated with a separate linked PR~~
  - ~~[ ] [altinn-app-frontend](https://docs.altinn.studio/community/changelog/app-frontend/)- (if applicable)~~
  - ~~[ ] [nuget-packages](https://docs.altinn.studio/community/changelog/app-nuget/) - (if applicable)~~

### DIFF
--- a/test/cypress/e2e/integration/app-frontend/options.js
+++ b/test/cypress/e2e/integration/app-frontend/options.js
@@ -6,17 +6,30 @@ import AppFrontend from '../../pageobjects/app-frontend';
 const appFrontend = new AppFrontend();
 
 describe('Options', () => {
-  // disabled until https://github.com/Altinn/altinn-studio/issues/8255 is solved
-  it.skip('is possible to retrieve options dynamically', () => {
+  it('is possible to retrieve options dynamically', () => {
     cy.navigateToChangeName();
-    // Case: options are dynamicly refetched based on what the user selects as source
+    // Case: options are dynamically refetched based on what the user selects as source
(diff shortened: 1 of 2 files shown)


lorang92 (approved):
So simple 💹 

Nice 👏

PR #87, opened 2022-05-11
Second fix for flaky options test

olemartinorg:
## Description

As mentioned in the issue, the root cause of this (still sometimes) failing cypress test is slow requests not returning before the user (or more likely, `preselectedOptionIndex`) performs an action that is meant to affect the choice of options shown. It was easy to reproduce by artificially slowing down certain options responses, but not others.

With this change, the metadata is set into the state store early, so that `checkIfOptionsShouldRefetchSaga` could do its job even if the target options had not been fetched for the first time yet.

## Fixes
- Altinn/altinn-studio/issues/8255

## Verification
- [X] All tests run green (finally!? :crossed_fingers: )


src/altinn-app-frontend/src/shared/resources/options/fetch/fetchOptionsSagas.ts
```diff
@@ -1,10 +1,10 @@
 import { SagaIterator } from 'redux-saga';
 import { fork, call, select, takeLatest, takeEvery } from 'redux-saga/effects';
-import { IRuntimeState, IOption, IFetchSpecificOptionSaga, IOptionData, IOptions } from 'src/types';
+import { IRuntimeState, IOption, IFetchSpecificOptionSaga, IOptions, IOptionsMetaData } from 'src/types';
 import { ILayouts, ISelectionComponentProps } from 'src/features/form/layout';
```
haakemon, line 4:
```suggestion
import type { IRuntimeState, IOption, IFetchSpecificOptionSaga, IOptions, IOptionsMetaData } from 'src/types';
import type { ILayouts, ISelectionComponentProps } from 'src/features/form/layout';
```

haakemon (approved):
Good investigative efforts! 🥇

Issue #28, opened 2021-08-20
Moderate wcag issues in altinn appfrontend


jeevananthank:
## Describe the bug
Altinn app frontend has moderate WCAG issues and the automated tests are failing.

## To Reproduce
1. Install the axe devtools chrome [plugin](https://chrome.google.com/webstore/detail/axe-devtools-web-accessib/lhdoppojpmngadmnindnejefpokejbdd)
2. Start an app instance and run the axe scan from the dev tools
3. Moderate issues are reported by the scan.

## Expected behavior
No issues should be reported.

## Screenshots
![image](https://user-images.githubusercontent.com/49391725/130199520-257f21e0-e59b-4d87-9521-0e720c8dfc1c.png)

## Additional info
env: at22
app: ttd/apps-test

Febakke:
This should be fixed as a part of #10 
Closing this issue

PR #88, opened 2022-05-12
feature/10-wcag-skip-blocks

lorang92:
## Description

- Added landmarks for our solution
  - https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA11.html
  - https://www.w3.org/WAI/tutorials/page-structure/regions/
- Enables jumping to main content from top of page
  - https://www.uutilsynet.no/wcag-standarden/241-hoppe-over-blokker-niva/103 

Follows same design pattern as altinn.no:
AD8D7D37-A680-42D9-9CFA-86B2D54E7D0E

Also fixes the two moderate WCAG issues reported in #28 
96E1A85A-73D3-4ADE-8D2A-D8A571C9CBBB

## Fixes
- #10

## Verification
- [x] All tests run green

## Documentation


src/altinn-app-frontend/src/features/form/components/MessageBanner.tsx
```diff
@@ -7,14 +7,14 @@ import classNames from 'classnames';
 
 const useStyles = makeStyles({
```
mjulstein, line 8: Maybe the Grid should be a header container?

src/altinn-app-frontend/src/features/instantiate/containers/InstantiationContainer.tsx
```diff
@@ -41,7 +41,9 @@ function InstantiateContainer(props: IInstantiateContainerProps) {
         profile={profile}
         type={props.type}
       />
-      {children}
+      <main id='main-content'>
```
mjulstein, line 44: Maybe add `role="main"` to the main-container for extended assistive support.

lorang92: No need:
`The [<main>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/main) element has a role of main. Developers should use semantic HTML -- in this case <main> -- over using ARIA.`

src/altinn-app-frontend/src/shared/components/__snapshots__/altinnAppHeader.test.tsx.snap
```diff
@@ -4,6 +4,17 @@ exports[`features/altinnAppHeader.tsx should match snapshot 1`] = `
 <div
   className="AltinnAppHeader-appBarWrapper-1"
 >
+  <nav
```
mjulstein, line 7: Why not a list of landmark-links?
Wouldn't a link to `href="#main"` make more more sense here?

lorang92: Does not work as we use hash-router in our solution as of now..

See workaround in `MainContentNav.tsx`.

src/shared/src/components/MainContentNav.tsx
```diff
@@ -0,0 +1,65 @@
...
+  const handleOnClick = () => {
+    // workaround because we still use a hash-router (sigh...)
+    // can be replaced by the more elegant solution <a href="#main-content></a> once this is no longer the case.
+    const main = document.getElementById('main-content');
+    main.tabIndex = -1;
+    main.focus();
+  }
+
+  return (
+    <nav className={cn(classes.jumpNav)}>
```
mjulstein, line 55: How will this `<nav>` scale to allow more items?
If this is the "Jump to landmark link list component", maybe a different name?
MainContentNav sounds like something that is displayed over the main content to me.
Maybe LandmarkShortcuts

lorang92: Great suggestion, renaming and refactoring to make scaling more straight forward.

src/shared/src/components/MainContentNav.tsx
```diff
@@ -0,0 +1,65 @@
...
+});
+
+export function MainContentNav({
+  navigateText,
+}: IMainContentNavProps) {
+  const classes = useStyles();
+
+  const handleOnClick = () => {
+    // workaround because we still use a hash-router (sigh...)
+    // can be replaced by the more elegant solution <a href="#main-content></a> once this is no longer the case.
```
mjulstein, line 48: AH, sorry I did not recall this, so disregard the comment about using links instead of buttons.
Maybe we can still use links, but change a query variable, so something like
https://.......#application-name?landmark=main-content
That way it might be possible to send a link to the landmark inside an app...
It's not really the way to go long term, as queries are usually before the hash and usually causes a page refresh when they change, but on the right hand they don't. Just taught maybe it could help us keep the buttons out.

src/shared/src/components/organisms/AltinnAppHeader.tsx
```diff
@@ -53,16 +53,22 @@ const styles = createStyles({
...
+  userParty,
+  language
+}: IAltinnAppHeaderProps) {
+
   return (
     <AppBar
       position='relative'
       classes={{ root: classes.altinnAppHeader }}
-      style={{ backgroundColor: headerColor, color: logoColor }}
+      style={{ backgroundColor: headerBackgroundColor, color: logoColor }}
```
mjulstein, line 69: Are we not moving away form inline styles?

src/shared/src/components/MainContentNav.tsx
lorang92: I would prefer to keep this as simple as possible for now. At some point we have to move away from hash-router, then this will all be solved by simple anchors to the relevant element id.

src/altinn-app-frontend/src/features/form/containers/Form.test.tsx
```diff
@@ -1,106 +1,99 @@
...
-  getFormLayoutStateMock,
-} from '../../../../__mocks__/mocks';
-import type { ILayoutState } from '../layout/formLayoutSlice';
-import type { ILayoutGroup } from '../layout';
-
 import { Form } from './Form';
+import { PreloadedState } from '@reduxjs/toolkit';
+import { RootState } from 'src/store';
+import { renderWithProviders } from 'src/../testUtils';
+import { ILayout, ILayoutComponent } from '../layout';
```
haakemon, line 6:
```suggestion
import type { ILayout, ILayoutComponent } from '../layout';
```

src/shared/src/components/MainContentNav.tsx
```diff
@@ -0,0 +1,65 @@
...
+    },
+  }
+});
+
+export function MainContentNav({
+  navigateText,
+}: IMainContentNavProps) {
+  const classes = useStyles();
+
+  const handleOnClick = () => {
```
haakemon, line 46: Redundant `on`

```suggestion
  const handleClick = () => {
```

src/shared/src/components/MainContentNav.tsx
```diff
@@ -0,0 +1,65 @@
...
+    const main = document.getElementById('main-content');
+    main.tabIndex = -1;
+    main.focus();
+  }
+
+  return (
+    <nav className={cn(classes.jumpNav)}>
+      <button
+        role='link'
+        onClick={handleOnClick}
```
haakemon, line 58: Redundant `on`

```suggestion
        onClick={handleClick}
```

src/altinn-app-frontend/src/features/form/components/MessageBanner.tsx
lorang92: Please elaborate on that?

test/cypress/e2e/integration/app-frontend/mobile.js
```diff
@@ -35,8 +35,8 @@ describe('Mobile', () => {
     cy.contains(mui.button, texts.next).click();
     cy.get(appFrontend.group.sendersName).should('be.visible').type('automation');
     cy.get(appFrontend.navMenu).should('have.attr', 'hidden');
-    cy.get('nav > button').should('be.visible').and('have.attr', 'aria-expanded', 'false').click();
-    cy.get('nav > button').should('have.attr', 'aria-expanded', 'true');
+    cy.get('#form-content-nav2 > div > nav > button').should('be.visible').and('have.attr', 'aria-expanded', 'false').click();
```
mjulstein, line 38: Could the selector possibly be a bit more helpful in terms of explaining what we are selecting. Maybe we should make it a habit to wrap the selector strings in variables, to make these tests more readable?

src/shared/src/components/LandmarkShortcuts.tsx
```diff
@@ -0,0 +1,65 @@
...
+const useStyles = makeStyles({
+  button: {
+    textDecoration: 'underline',
+    textAlign: 'left',
+    width: 'fit-content',
+  },
+  jumpNav: {
+    backgroundColor: "#fff",
+  },
+  'sr-only': {
```
mjulstein, line 18: maybe move the content of this class into sr-only-focusable

src/shared/src/components/LandmarkShortcuts.tsx
```diff
@@ -0,0 +1,65 @@
...
+  const handleClick = () => {
+    // workaround because we still use a hash-router (sigh...)
+    // can be replaced by the more elegant solution <a href="#main-content></a> once this is no longer the case.
+    const main = document.getElementById('main-content');
+    main.tabIndex = -1;
+    main.focus();
+  }
+
+  return (
+    <nav className={cn(classes.jumpNav)}>
```
mjulstein, line 55: does this need styling, can't it just be the button?

src/shared/src/components/LandmarkShortcuts.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+export interface ILandmarkShortcutsProps {
+  navToMainText: string;
+}
+
+const useStyles = makeStyles({
+  button: {
+    textDecoration: 'underline',
+    textAlign: 'left',
+    width: 'fit-content',
```
mjulstein, line 13: maybe some padding would be nice?
also not sure if 'fit-content' does much here.

src/shared/src/components/LandmarkShortcuts.tsx
```diff
@@ -0,0 +1,65 @@
...
+    height: '1px',
+    padding: 0,
+    margin: '-1px',
+    overflow: 'hidden',
+    clip: 'rect(0,0,0,0)',
+    border: 0,
+  },
+  'sr-only-focusable': {
+    '&:focus': {
+      position: 'relative',
```
mjulstein, line 30: Should the entire page jump, or is the intent to only have the button appear as a layover.
Maybe position: fixed and top: 0px, left: 0px etc. could work?

src/shared/src/components/LandmarkShortcuts.tsx
mjulstein, line 18: Given this likely not being the only place where sr-only needs to be used, maybe make it an export in a `srOnly.ts` file and use spreading where it is applicable?
```js
import {srOnly} from ´altinn-shared/styles´;
...
'sr-only': {
{...srOnly}
...
}
```

src/shared/src/components/LandmarkShortcuts.test.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+  it('should set active element to main when nav is clicked', async () => {
+    renderMainContentNav();
+    await userEvent.click(screen.getByRole('link'));
+    expect(document.activeElement).toBe(document.getElementById('main-content'));
+  });
+
+  function renderMainContentNav(props: Partial<ILandmarkShortcutsProps> = {}) {
+    const defaultProps: ILandmarkShortcutsProps = {
+      navToMainText: 'Hopp til hovedinnhold'
```
mjulstein, line 21: I think this landmark api could be easier to config:
this could be an array with `[ {id, shortcutText} ]`.
```js
[{ id: 'main-content', shortcutText: 'Hopp til hovedinnhold'}]
```
Then it would be easier to extend to other landmarks later.
Some changes would have to be made in LandmarkShortcuts to iterate through these.

test/cypress/e2e/integration/app-frontend/mobile.js
lorang92: Good point, extracted to variable.

src/shared/src/components/LandmarkShortcuts.tsx
```diff
@@ -0,0 +1,65 @@
...
+    main.tabIndex = -1;
+    main.focus();
+  }
+
+  return (
+    <nav className={cn(classes.jumpNav)}>
+      <button
+        role='link'
+        onClick={handleClick}
+        className={cn(classes.button, classes["sr-only"], classes["sr-only-focusable"])}
```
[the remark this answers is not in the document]
lorang92: Good point, combined for better readability. Also removed some non-needed stuff (was a pure copy of the existing class in the old altinn-design-system).

src/shared/src/components/LandmarkShortcuts.tsx
lorang92: Done ✔️

src/shared/src/components/LandmarkShortcuts.tsx
lorang92: In order to follow the same design pattern as on altinn.no, we need this.

src/shared/src/components/LandmarkShortcuts.tsx
lorang92: Agreed, now has same padding as altinn.no design.

src/shared/src/components/LandmarkShortcuts.tsx
lorang92: Entire page, as we want consistency across solutions.

src/shared/src/components/LandmarkShortcuts.test.tsx
lorang92: Done ✔️

src/altinn-app-frontend/src/components/presentation/Header.tsx
```diff
@@ -19,7 +19,7 @@ const Header = ({
   appOwner
 }: IHeaderProps) => {
   return (
-    <div
+    <header
```
mjulstein, line 22: Nice :)

src/altinn-app-frontend/src/features/form/components/MessageBanner.tsx
mjulstein: I think I put this comment on the wrong location.
Sorry.

src/altinn-app-frontend/src/features/form/containers/Form.test.tsx
```diff
@@ -1,106 +1,99 @@
...
+      disabled: false,
+    },
+    {
+      id: 'field3',
+      type: 'Input',
+      dataModelBindings: {
+        simple: 'Group.prop3',
+      },
+      textResourceBindings: {
+        title: 'Third title',
```
mjulstein, line 46: This is great :) Makes for better test AND test results.

src/altinn-app-frontend/src/features/form/components/MessageBanner.tsx
mjulstein: I think I meant that the component in this file could be a header instead of a div.
Only based it on the name of the file containing banner, which is often in a header.

src/altinn-app-frontend/src/features/form/containers/Form.tsx
```diff
@@ -98,34 +98,41 @@ export function Form() {
...
-          error={requiredFieldsMissing}
-          messageKey={'form_filler.required_description'}
-        />
+      {navComponent &&
+        <Grid
+          container={true}
+          spacing={3}
+          alignItems='flex-start'
+        >
+          {renderLayoutComponent(navComponent, layout)}
```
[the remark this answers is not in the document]
lorang92: It is. `SummaryGroupComponent.tsx`

src/altinn-app-frontend/src/features/form/containers/Form.tsx
```diff
@@ -98,34 +98,41 @@ export function Form() {
...
 
   return (
     <div>
-      {hasRequiredFields(layout) &&
-        <MessageBanner
-          language={language}
-          error={requiredFieldsMissing}
-          messageKey={'form_filler.required_description'}
-        />
+      {navComponent &&
```
mjulstein, line 109: Maybe this can be a landmark as well?
You can use the same condition to consider adding it to the landmarkShortcuts-array.

lorang92: Don't think it is needed. The point of having a "skip to main content" shortcut is to jump over repeating stuff like this.

src/altinn-app-frontend/src/features/instantiate/containers/InstantiationContainer.tsx
mjulstein: Indeed! Maybe this is added automatically by the compiler if we support lesser browser versions.

src/shared/src/components/LandmarkShortcuts.tsx
```diff
@@ -0,0 +1,65 @@
...
+    padding: '5px',
+    paddingLeft: '10px',
+  },
+  jumpNav: {
+    backgroundColor: "#fff",
+  },
+  srOnly: {
+    position: 'absolute',
+    clip: 'rect(0,0,0,0)',
+    border: 0,
```
mjulstein, line 27: Is removing all the other attributes "sr-only enough" for all supported browsers?

src/shared/src/components/LandmarkShortcuts.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+export function LandmarkShortcuts({
+  shortcuts,
+}: ILandmarkShortcutsProps) {
+  const classes = useStyles();
+
+  const handleClick = (id: string) => {
+    // workaround because we still use a hash-router (sigh...)
+    // can be replaced by the more elegant solution <a href="#main-content></a> once this is no longer the case.
+    const main = document.getElementById(id);
```
mjulstein, line 44: maybe `main` can be `target` after the last change... no biggy.

mjulstein (approved):
Some non-important tweaks, but mostly good stuff :)

src/shared/src/components/LandmarkShortcuts.tsx
lorang92: Tested OK on safari + chrome. Do you have any reason to believe this is not gonna work?

Suggest we test this. Dragged to test in the board.

src/shared/src/components/LandmarkShortcuts.tsx
```diff
@@ -0,0 +1,65 @@
...
+    main.tabIndex = -1;
```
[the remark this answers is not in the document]
lorang92: Good catch, did not think of this. Fixed.

src/shared/src/components/LandmarkShortcuts.tsx
lorang92: Agreed, fixed.

src/altinn-app-frontend/src/features/form/containers/Form.tsx
mjulstein: I believe you are right. It's probably more common on webpages with more boilerplate and advertisement before the page nav etc.

nkylstad:
@lorang92 Tested ok in Chrome, Edge and Firefox.
2 code smells that can be fixed before merge 😇

PR #89, opened 2022-05-12
chore: dont create a new version for every merge to master

haakemon:
## Description
This will stop creating a new release to CDN on every merge to master. This is problematic because we need to manually remember to update package.json to avoid overwriting whats already published.

Creating a new release will now require that we go to Actions in Github, and run the `release` pipeline. Here you will need to input `major`, `minor` or `patch` to decide which type of version to bump.

This approach is similar to what we do in https://github.com/Altinn/altinn-design-system and https://github.com/Altinn/figma-design-tokens repos, so you can have a look at the action in one of those repos to see how it will work.

We can always improve it further later, but I think its important that we get away from the current error-prone situation 😁

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,90 @@
+name: Create release
+on:
+  workflow_dispatch:
+    inputs:
+      release_type:
+        description: "The type of release (one of 'patch', 'minor', or 'major')"
+        required: true
+        default: patch
+        type: string
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    name: Build
+    steps:
+      - name: checkout
+        uses: actions/checkout@v3
+        with:
+          path: app-frontend
+          fetch-depth: 0 # Shallow clones should be disabled for a better relevancy of analysis
+
+      - name: install node
+        uses: actions/setup-node@v2
+        with:
+          node-version: '16'
+
+      - name: install dependencies
+        working-directory: app-frontend/src
+        run: yarn --immutable
+
+      - name: Prepare release version
+        working-directory: app-frontend/src/altinn-app-frontend
+        # This step runs if the trigger is run via github action
+        if: ${{ ! github.event.after }}
+        run: |
+          git config --global user.name 'github-actions[bot]'
+          git config --global user.email 'github-actions[bot]@users.noreply.github.com'
+          yarn release ${{ github.event.inputs.release_type }}
+          echo "release_version=`git describe`" >> $GITHUB_ENV
+          git push --follow-tags
+
+      - name: run build
+        working-directory: app-frontend/src/altinn-app-frontend
+        run: yarn build
+
+      - name: Checkout Altinn-CDN repository
+        if: github.ref == 'refs/heads/main'
+        uses: actions/checkout@v3
+        with:
+          repository: 'Altinn/altinn-cdn'
+          token: ${{secrets.ALTINN_CDN_TOKEN}}
+          path: cdn
+
+      - name: Publish to CDN
+        if: github.ref == 'refs/heads/main'
+        working-directory: app-frontend
+        run: |
+          echo Clone, copy, commit and push to Altinn-CDN
+          APP_VERSION=$(jq '.version' ./src/altinn-app-frontend/package.json)
+          APP_MAJOR=$(echo $APP_VERSION | tr --delete '"' | cut -d'.' -f1)
+          APP_MINOR=$(echo $APP_VERSION | tr --delete '"' | cut -d'.' -f2)
+          APP_PATCH=$(echo $APP_VERSION | tr --delete '"' | cut -d'.' -f3)
+          echo altinn-app-frontend version: $APP_MAJOR.$APP_MINOR.$APP_PATCH
+          AUTHOR_FULL=$(git log -1 | grep Author)
+          AUTHOR_NAME=$(git log -1 | grep Author | cut -d' ' -f2)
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,90 @@
+name: Create release
```
haakemon, line 1: @tjololo or anyone else who knows workflows better than me, is there any easy way to restrict this to only be able to run this from main branch?

tjololo: I dont't think you can restrict it from running. The workflow as it is will skip the checkout cdn and publish to cdn steps as far as I can see.

.github/workflows/release.yml
```diff
@@ -0,0 +1,90 @@
...
+        run: yarn --immutable
+
+      - name: Prepare release version
+        working-directory: app-frontend/src/altinn-app-frontend
+        # This step runs if the trigger is run via github action
+        if: ${{ ! github.event.after }}
+        run: |
+          git config --global user.name 'github-actions[bot]'
+          git config --global user.email 'github-actions[bot]@users.noreply.github.com'
+          yarn release ${{ github.event.inputs.release_type }}
```
olemartinorg, line 38: I tried to find any documentation for this step, but failed.. Is it github workflow magic, or does it depend on a release script we don't currently have in package.json?

olemartinorg:
Loving this! Next step, as previously mentioned on slack, I think would be to automate pushing release notes/change log to the docs repo as well. I could try adding that functionality if nobody else badly wants to, but I might need some hand holding with github workflows.

After this change, I guess it would either be up to the person running the release action to create a docs PR detailing all the latest changes, or we could do a collective draft PR where everyone adds their changes to the change log and the last person specifies the version number and date the changes were released.

olemartinorg (approved):

.github/workflows/release.yml
haakemon: Yeah, this is wrong - it should be `npm release`, not `yarn release`. Ill fix!

- [ ] fix incorrect command

.github/workflows/release.yml
haakemon: Made some changes to this so none of the steps will run unless its on master. That should be good enough I think

mjulstein (approved):
🥇

haakemon:
@olemartinorg Yep, next step would be to tackle changelogs. Not sure how we should deal with it though, we may need a discussion around it. Conventional commits is one option, but requires some effort in how we work, and will also require some extra effort from external contributors I think.

Should we have a short discussion around this next week?

haakemon:
I just thought about a couple potential challenges with this approach... 

* The release pipeline will update package.json and create a new commit, and push this commit to the repo. Pushing this to origin will not be possible if someone has merged a PR during this period, because the commit history will be diverged. 
	* This is a bit edge-case I guess, but it means we still need to pay a bit extra attention when creating releases.


* Since the release pipeline pushes a commit back to master, it needs to have rights to do this. I am not sure if pushing commits directly to master is allowed in this repo. @altinnadmin Can you check the rights for this?

Any other thoughts on these challenges? @olemartinorg @mjulstein @nkylstad @lorang92

tjololo:
Not sure if it is possible, or desirable, but one option could be to not have the version defined in package.json and set this at build/release-time. We can then make the pipeline trigger on release/tag-created?

haakemon:
@tjololo Yeah that is a good point. Since this project is not published as a npm package itself, having the version in package.json really isnt nescessary. Perhaps this is a a better approach then. Any other thoughts/objections to that approach (essentially remove version in package.json) from the others?

altinnadmin:
>  I am not sure if pushing commits directly to master is allowed in this repo. @altinnadmin Can you check the rights for this?

@haakemon Our default setting (also for this repo) is to require a PR with review. Is this still relevant given the suggestion from @tjololo ?

haakemon:
@altinnadmin No, its not relevant anymore. We will remove version nr from package.json and rely on git tags for version instead. This way we dont need to push commits directly to master.

haakemon:
@tjololo Can you have another look at the updated workflow?

It will work mostly the same as previously, except it will create a git tag with the version instead of a commit with version update in package.json.

tjololo:
> @tjololo Can you have another look at the updated workflow?
> 
> It will work mostly the same as previously, except it will create a git tag with the version instead of a commit with version update in package.json.



> @tjololo Can you have another look at the updated workflow?
> 
> It will work mostly the same as previously, except it will create a git tag with the version instead of a commit with version update in package.json.

Can't see anything wrong with the workflow. Sorry to come with another suggestion so late: We could also just use Github releases to set the version and have the workflow that publishes to altinncdn trigger on release published. Creating a release creates tags in the repository. The release can also automatically create release notes based on merged PRs

tjololo:
> @tjololo Can you have another look at the updated workflow?
> 
> It will work mostly the same as previously, except it will create a git tag with the version instead of a commit with version update in package.json.



> @tjololo Can you have another look at the updated workflow?
> 
> It will work mostly the same as previously, except it will create a git tag with the version instead of a commit with version update in package.json.

Can't see anything wrong with the workflow. Sorry to come with another suggestion so late: We could also just use Github releases to set the version and have the workflow that publishes to altinncdn trigger on release published. Creating a release creates tags in the repository. The release can also automatically create release notes based on merged PRs

haakemon:
@tjololo  Great suggestion, I will update to use github releases

tjololo (approved):
:godmode:

PR #119, opened 2022-05-20
chore: remove invalid condition in release workflow

haakemon:
Related to https://github.com/Altinn/app-frontend-react/pull/89 , which had an incorrect condition to only run on refs/heads/main . This will not work, since the ref will be a tag instead.


tjololo (approved):
:godmode:

PR #179, opened 2022-05-27
Support pre-releases in workflow and generate a list of versions as a .json on CDN

olemartinorg:
## Description
Supporting pull request for an upcoming PR in studio, where a list of released versions would be helpful when building a selector for testing previous versions.

`jq` [should be available in ubuntu-latest](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-Readme.md#tools)

## ~~Fixes~~ Supports
- Altinn/altinn-studio/pull/8562

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -62,7 +62,9 @@ jobs:
           mkdir -p ./cdn/toolkits/altinn-app-frontend/$APP_MAJOR.$APP_MINOR.$APP_PATCH
           cp -fr ./app-frontend/src/altinn-app-frontend/dist/altinn-app-frontend.js ./cdn/toolkits/altinn-app-frontend/$APP_MAJOR.$APP_MINOR.$APP_PATCH/altinn-app-frontend.js
           cp -fr ./app-frontend/src/altinn-app-frontend/dist/altinn-app-frontend.css ./cdn/toolkits/altinn-app-frontend/$APP_MAJOR.$APP_MINOR.$APP_PATCH/altinn-app-frontend.css
-          cd cdn
+          cd cdn/toolkits/altinn-app-frontend
+          ls -1 | grep --perl-regexp '^[\d\.]+$' | sort --version-sort | jq --raw-input --slurp 'split("\n") | map(select(. != ""))' > index.json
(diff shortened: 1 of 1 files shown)


ivarne:
Seems like a reasonable thing to do.

What does the output look like? 
image

olemartinorg:
It's just a list of strings, sorted with latest versions at the bottom. A snip:

```shell
$ ls -1 | grep --perl-regexp '^[\d\.]+$' | sort --version-sort | jq --raw-input --slurp 'split("\n") | map(select(. != ""))' | tail
  "3.36.3",
  "3.36.4",
  "3.37",
  "3.37.0",
  "3.37.1",
  "3.37.2",
  "3.37.4",
  "3.38",
  "3.38.0"
]
```

olemartinorg:
Attention reviewers: I'm piling on additional functionality. I'm going on vacation tomorrow, and while I'm not quite done with my current tasks, they should be ready to test-drive by application developers while I'm gone. Along with the changes related to this PR, it would be very easy for an application developer to test out a pre-release version (maybe even released from a feature branch instead of main) without rolling that release out to everyone using the major tag.

PS: I don't know how to test github workflows, so squint a bit more than usual when reviewing.

tjololo:
The code looks good, but I have a question regarding the pre-release logic.
The current setup is to release a x.y.z version of the files but don't upgrade the x and x.y folders. This is fine for the cdn repo logic, but a prerelease should follow the semver2 convetion x.y.z-alpha.
Will the current setup assume all tags are on the x.y.z format and then the tag x.y.z would be consumed to create the pre-release

tjololo (changes requested):
We need to address the version number logic when it comes to pre-releases.

tjololo:
After reading the script again I see that my first comment is false. Prereleas part of the string will be part of the PATCH part atleast if it does not contain `.` 
After discussing the PR with @olemartinorg we have reached the conclusion that the logic is safer with the prerelease logic and I will create an Issue for handling `.` in prereleases

tjololo (approved):
:godmode:

PR #182, opened 2022-05-30
chore: remove need to override config in package json for local testrun

### LINKED ISSUE #180: Some cypress test commands are not working with Yarn
## Description
Scripts like `before:appfrontend` which is suggested to use in the readme to get started are not working with yarn. We should update this to be yarn compatible, so we dont have to force developers to use both npm and yarn.

Additionally there are some config values that needs to be changed (paths to test repos, and localtest directory) and these are currently set in package.json. This is a bit unfortunate, since it is easy to commit these changes by mistake. We should change how these config values are read, so they are not part of the git history.

## Tasks
- [ ] Make all scripts listed in package.json and readme yarn compatible
- [ ] Ensure config values are not part of git history
- [ ]

haakemon:
## Description
- PDF and PNG testfiles are now checked in so they wont have to be created before testruns, these files are never modified and take only 1kb each. Having to generate them before testruns seems unnescessary.
- Removed config object in package.json and rely on `.env` file instead
- Simplify code around environment loading, and show error message if invalid environment is specified
- use explicit [`node:` prefix ](https://2ality.com/2021/12/node-protocol-imports.html#benefits-of-node%3A-imports) for built-in node module imports, to avoid using external modules when we meant to use node modules (f.ex `path`)
- Fixed an issue in `formComponentUtils.ts` which caused tests to fail
- Removed `start:frontend-no-deps` script to reduce amount of steps/scripts in the test project. The extra checks that are done without this (just running `start:frontend` instead) only takes a few seconds. I can add this script back though if it seems useful for others.

## Fixes
- fixes #180 

## Verification
- [ ] ~Relevant automated test added (if you find this hard, leave it and we'll help out)~
- [x] All tests run green

## Documentation
- [ ] ~User documentation is updated with a separate linked PR in [altinn-studio-docs.](https://github.com/Altinn/altinn-studio-docs) (if applicable)~


olemartinorg:
Nice! :1st\_place\_medal: After a bit of debugging, I got this to run on Linux. It's still not without friction, as you'd have to manually start the two other apps. Maybe we should take a look at starting the correct apps on the fly before running each test suite, when running locally - or starting them all when running LocalTest. I noticed the setup right now assumes you have set up LocalTest to work with multiple apps at once, which I haven't.. :wink:

haakemon:
> It's still not without friction, as you'd have to manually start the two other apps. Maybe we should take a look at starting the correct apps on the fly before running each test suite, when running locally - or starting them all when running LocalTest. I noticed the setup right now assumes you have set up LocalTest to work with multiple apps at once, which I haven't.. 😉

Good point, I hadnt really thought about this 😬 Ill look into this a bit more 🙂

lorang92 (approved):

Issue #159, opened 2022-01-21
Cleanup of store in app-frontend 


lorang92:
## Description

Currently we have disabled the default `serializableCheck` and `immutableCheck` redux middlewares as we have some operations that currently mutates the redux state + non-serializable elements. This could lead to unexpected behaviour and should be fixed.

Seems to be related to our handling of variables in text. Need to investigate further.

```
getDefaultMiddleware({
      // TODO: enable once we have cleaned up our store
      serializableCheck: false,
      immutableCheck: false,
    })
```

## To Reproduce
1. Enable default redux middleware
2. See errors

## Expected behavior
No mutations and state should be serializable.

## Tasks

- [x] Fix immutability issues https://github.com/Altinn/app-frontend-react/pull/186
- [x] Fix serializable issues

olemartinorg:
Should be fixed in #372! :partying_face: The TODO has been removed, and both serializable and immutable checks are on now.

PR #186, opened 2022-06-01
bug: avoid mutating redux state

lorang92:
## Description
- avoid mutating text resources in our redux store (just a pure copy existing logic but create new object instead of mutation)
  - cleanup of tests
- avoid mutating file upload state, removed duplicated local + global reducer as this caused issues
  - added tests for touched parts of code 
- enabled immutability check for our redux store to avoid such errors again

## Fixes
- Partially solves #159, where we get rid of logic that mutates text resource array which caused problems for #https://github.com/Altinn/app-frontend-react/pull/83, and might have caused other bugs elsewhere.

## Verification
- [x] All tests run green

## Documentation
- [x] User documentation is updated with a separate linked PR in [altinn-studio-docs.](https://github.com/Altinn/altinn-studio-docs) (if applicable)

### DIFF
--- a/src/altinn-app-frontend/src/store/index.ts
+++ b/src/altinn-app-frontend/src/store/index.ts
@@ -14,7 +14,7 @@ export const setupStore = (preloadedState?: PreloadedState<RootState>) => {
     middleware: (getDefaultMiddleware) => getDefaultMiddleware({
       // TODO: enable once we have cleaned up our store
       serializableCheck: false,
-      immutableCheck: false,
+      immutableCheck: true,
     }).concat(middlewares),
     preloadedState
   });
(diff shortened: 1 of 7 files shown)


src/altinn-app-frontend/src/store/index.ts
```diff
@@ -14,7 +14,7 @@ export const setupStore = (preloadedState?: PreloadedState<RootState>) => {
     middleware: (getDefaultMiddleware) => getDefaultMiddleware({
       // TODO: enable once we have cleaned up our store
       serializableCheck: false,
```
haakemon, line 16: Any idea whats missing to enable this as well (and do we have a task to fix it)?

haakemon (approved):
Great job getting this fixed! 🥇

src/altinn-app-frontend/src/store/index.ts
lorang92: Not exactly sure. Will have to investigate further. I have added it as a sub-task on https://github.com/Altinn/app-frontend-react/issues/159.

I suggest we try to get this done in our next sprint.

PR #212, opened 2022-06-03
Make labelGrid work for radio buttons.

ivarne:
Some schemas with many questions, benefit from having the answers on the right side and label on the left. I implemented this in https://github.com/Altinn/altinn-studio/pull/7027, ([docs](https://docs.altinn.studio/app/development/ux/styling/#innergrid-og-labelgrid)), but just found out that it did not work for `RadioButtons` or other components designated as [`noLabelComponents`](https://github.com/Altinn/app-frontend-react/blob/28244732ba9c978019be3e9e6153d639548e705c/src/altinn-app-frontend/src/components/GenericComponent.tsx#L292-L305).

image

### Notes for review
* The radio button is wrapped in a `<fieldset>` tag, that makes it somewhat harder to work with the grid system.
* I had to use a `class="innerGrid"` (not jss) to add correct padding to the content element.
* One might consider dropping the padding, and rather align the checkboxes to the right of the screen.
* I changed the html for all `noLabelComponents` to have 1 less nested <div, for no particular reason, other than that it looked correct.

## Verification
- [ ] All tests run green

### DIFF
--- a/src/altinn-app-frontend/src/components/GenericComponent.tsx
+++ b/src/altinn-app-frontend/src/components/GenericComponent.tsx
@@ -56,40 +56,40 @@ const useStyles = makeStyles((theme) => ({
   },
   xs: {
     'border-bottom': '1px dashed #949494',
-    '& > div:nth-child(2)':{
-      paddingLeft: theme.spacing(3/2) // Half the spacing of <Grid in <Form
-    }
+    '& .innerGrid': {
+      paddingLeft: theme.spacing(3 / 2), // Half the spacing of <Grid in <Form
+    },
   },
   sm: {
     [theme.breakpoints.up('sm')]: {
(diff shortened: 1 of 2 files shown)


ivarne:
I see that this branch has gone stale and need conflict resolution. Is this something you would consider reviewing and merging?

I see that the build and run unit tests failed, but it looks like it's because I'm not in a team that have access to the secret that is set in the `SONAR_TOKEN` environment variable.

@olemartinorg

ivarne:
I'm very sorry. The merge conflict was because the Likert component added a little bit more complexity to `GenericComponent` and rewrote `RadioButtonsContainerComponent`. I'm giving up on making the changes required for this functionality without just adding more branching complexity. I'll open an issue and close this.

PR #223, opened 2022-06-16
remove create:testfiles from external cypress run

lorang92:


lorang92:
Handled in https://github.com/Altinn/app-frontend-react/pull/222

Issue #130, opened 2022-03-17
Validation: Present better error messages for required fields 


mrosvik:
## Description
_As a user I want a clear description of the requirements that are needed to complete the form._

We need to explain in the text what field the error is about. Example: "You need to fill out your email address" instead of "This field is required".

## Design
https://www.figma.com/file/Pvti3aRcOwH0k5Z7DrLznk/Arbeidsomr%C3%A5de%3A-Altinn-Studio-APPS?node-id=2545%3A50628

<img width="1055" alt="image" src="https://user-images.githubusercontent.com/1464915/158777458-c9788335-508b-41db-9f46-639f4a22a194.png">

## Possible solutions
- Use a convention where texts with the same id as the corresponding data model element is used instead of the generic error texts.
- Get the field-name from input label + generate requirements from json-schema?
UPDATE:
Displaying WHICH field should be completely doable, in that we can access information about the component and fetch the label text and add this in to the validation message. 
Displaying requirements for the field is not trivial to do consistently and well. Suggest that any such requirements that are important for the end user should be displayed elsewhere, f.ex. in a help text or description to the field and not in the error message.

We also need to consider how we handle required field validation, as we have some custom behaviours that are not present for other kinds of validation (showing a banner with information about required fields, which turns red when there are missing required fields). Currently we loop through the available validation messages for the current view and pick up any that are for missing required fields. This might be difficult to do once we implement this issue, so we need to decide on how we structure information about missing required fields. Some options:
- Continue as before, but make sure that the checks for if any required fields are missing in the view also handle more complex/detailed validation messages.
- Add some property to the fields validations, like `missing`, to be sidelined with `errors` and `warnings` that we can use as a flag, instead of looping through validation errors.
- Add a separate structure under `validations` in store to contain information about missing required fields for each view. Something similar to the `invalidData`-structure that we already have.

## Development tasks
- [x] Consider possible solutions above. As a minimum we should tell the user which field is required. For example "Du må fylle ut fødselsnummer. ". But could we also include the requirements? For example: "Du må fylle ut fødselsnummer (11 siffer uten mellomrom)". 
- [ ] Decide how to structure info about missing required fields.
- [ ] Implement solution - telling user which field is required in the error message by using field label text.

## Guidelines for app developers
For custom validation messages, we should remind the app developer that they need to include keywords relating to the label in the validation message. "Dette feltet er påkrevd" Is not allowed.
[...]

PR #230, opened 2022-06-16
Better error messages for required fields

nkylstad:
## Description
Implemented new error messages for required fields where field name (title) is used in the message. 
Added option for setting a `shortName` property on component dataModelBinding to handle cases where component title is too long to be useful in the error message.

## Related Issue(s)
- #130 

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/altinn-app-frontend/src/utils/databindings.test.ts
+++ b/src/altinn-app-frontend/src/utils/databindings.test.ts
@@ -4,6 +4,7 @@ import type { IMapping } from 'src/types';
 
 import {
   flattenObject,
+  getFormDataFromFieldKey,
   getKeyWithoutIndex,
   mapFormData,
   removeGroupData,
@@ -270,5 +271,32 @@ describe('utils/databindings.ts', () => {
         expect(mapFormData(formData, mapping)).toEqual(formData);
       },
     );
+  });
+
+  describe.only('getFormDataFromFieldKey', () => {
+    const formData = {
+      'field1': 'value1',
+      'group[0].field': 'someValue',
+      'group[1].field': 'another value',
+    }
+    it('should return correct form data for a field not in a group', () => {
+      const result = getFormDataFromFieldKey(
+        'simpleBinding',
+        { simpleBinding: 'field1' },
+        formData,
+      );
+      expect(result).toEqual('value1');
+    });
+
+    it('should return correct form data for a field in a group', () => {
+      const result = getFormDataFromFieldKey(
+        'simpleBinding',
+        { simpleBinding: 'group.field' },
+        formData,
+        'group',
+        1,
+      );
+      expect(result).toEqual('another value');
+    });
   });
 });
--- a/src/altinn-app-frontend/src/utils/formComponentUtils.test.ts
+++ b/src/altinn-app-frontend/src/utils/formComponentUtils.test.ts
@@ -20,6 +20,7 @@ import {
   isComponentValid,
   componentValidationsHandledByGenericComponent,
   componentHasValidationMessages,
+  getFieldName,
 } from './formComponentUtils';
 
 describe('formComponentUtils', () => {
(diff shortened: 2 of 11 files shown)


src/altinn-app-frontend/src/utils/formComponentUtils.test.ts
```diff
@@ -478,4 +479,59 @@ describe('formComponentUtils', () => {
         expect(result).toEqual(true);
       });
   });
+
+  describe.only('getFieldName', () => {
```
lorang92, line 483:
```suggestion
  describe('getFieldName', () => {
```

src/altinn-app-frontend/src/utils/formComponentUtils.ts
```diff
@@ -412,3 +413,25 @@ export const atleastOneTagExists = (attachments: IAttachment[]): boolean => {
...
+  if (fieldKey)
+  {
+    return getTextFromAppOrDefault(`form_filler.${fieldKey}`, textResources, language, null, true);
+  }
+
+  if (textResourceBindings.shortName) {
+    return getTextResourceByKey(textResourceBindings.shortName, textResources);
+  }
+
+  if (textResourceBindings.title) {
```
lorang92, line 432: Just a note here. This will be a "breaking" change in the sense that all fields with long title names will suddenly have validation messages that contain the whole title, right?

I'm a little unsure about how real this case is, hopefully, all applications in prod have chosen fairly short and concise titles. Is it worth some time just to investigate if this really is a problem?

Alternative would be to introduce a new major version but would be sweet not to force this on app developers.

Thoughts?

lorang92 (changes requested):
Neat 🥳  Surprised by how little code change this actually required.

src/altinn-app-frontend/src/utils/formComponentUtils.ts
nkylstad: Hmm fair point. I think our options are:
1. use only new shortName prop from reedy resources, making the functionality opt-in
  - fallback will be the same, just remove the title condition from this function
3.  Leave as-is
  A. bump to new major version, making the functionality opt-in but with no changes to form layout needed
  B. Bump minor version and push the change to everyone

We don’t have that many apps in production, but we do have a lot of apps underway. I am leaning towards 2A, so we don’t push this out immediately for all apps, and app developers can look over their apps to see if they need to introduce the shortName field anywhere the title field might be too long.
My understanding is that having this kind of information in validation messages is something of a requirement going forward, so making it completely opt-in by option 1 makes it too easy to just skip 🤷‍♀️

src/altinn-app-frontend/src/utils/databindings.test.ts
```diff
@@ -271,4 +272,31 @@ describe('utils/databindings.ts', () => {
       },
     );
   });
+
+  describe.only('getFormDataFromFieldKey', () => {
```
lasseklovstad, line 276: Remove `.only` or jest won't run all the tests in the file.

Magnusrm (commented):
Tested the feature to work with "shortName" and with "title". Seems to be working well! 🥇 
I have a commen about the validation text itself that should be looked at.

lorang92 (approved):

test/cypress/e2e/fixtures/texts.json
```diff
@@ -15,6 +15,7 @@
   "missingRights": "Du mangler rettigheter for å se denne tjenesten",
   "next": "Neste",
   "requiredField": "Feltet er påkrevd",
```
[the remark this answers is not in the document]
olemartinorg: Litt sent å svare - men jeg kom over denne kommentaren nå.. :wink: Fornavn-feltet i frontend-test har overraskende nok serverside-validering som sender denne meldingen, så det feltet henger fortsatt igjen med den gamle feilmeldingen. Noen tester sjekker også mot denne, så teksten må beholdes.

Issue #128, opened 2021-07-13
Deleted item values from nested group are added back on adding a new item


jeevananthank:
## Describe the bug
When deleting an item from the nested group, and adding a new item in the group, the fields in the nested group is filled with the values from the deleted item

## To Reproduce
1. Start an app instance with nested group configured.
2. Add an item to the main group and a new item to the nested group and fill the value.
3. Delete the nested group item and add a new item to the nested group.
4. the fields in the group is prefilled with the values from the deleted item.

## Expected behavior
The fields on the new item in the nested group should not be filled with value from the deleted item.

## Additional info
env: localtest
app: ttd/frontend-test

olemartinorg:
FYI testers: I found this bug while implementing attachments in repeating groups (#246) and fixed it in that PR. There is also a cypress test where this (erroneous) behavior was tested, and I changed that test when implementing #246.

You can run this to see the change:
`git diff f481331..1ae84e4 -- test/cypress/e2e/integration/app-frontend/group.js`

nkylstad:
@olemartinorg  Tested ok in frontend-test app, new item is not prefilled with values from deleted item.

However I see that it's not possible to completely delete the item if there is only one item in the nested group. The content of the item is deleted, but the item itself is still in the list (although empty). If I then click "add new item", a second item is added.

![image.png](https://images.zenhubusercontent.com/5dc93d9470573a0001558cc6/222fd2ba-ad02-4b09-819f-76a14a64a38d)

![image.png](https://images.zenhubusercontent.com/5dc93d9470573a0001558cc6/64c6fc39-eae4-490a-a63f-35c545d9f28a)

I would expect to be able to completely delete the item, even if it is the only item in the nested group. 
The behaviour works as expected for the parent group. I'm not sure if this was an existing issue for nested groupsor if it has been introduced with #246. I don't see an existing issue for it at least.

olemartinorg:
This is partly caused by the main group having `openByDefault` off, but the subgroup having it on. In addition to that, I actually broke that `openByDefault` functionality in [this commit](https://github.com/Altinn/app-frontend-react/pull/246/commits/e2e4c4bd195bb6da491ce72c38c170a7eb42b85c) where I removed the events dispatched to set the `editIndex`. The old approach is no longer possible - as deleting a group row can now be rejected/cancelled (if deleting attachments inside that group fails). I have a fix somewhat ready, and will write cypress tests for it.

nkylstad:
[...]

PR #246, opened 2022-06-22
Support for attachments in repeating groups

olemartinorg:
## Description
This feature adds support for uploading attachments to repeating groups. This requires a new 'list' dataModelBinding when using FileUpload/FileUploaderWithTag components allowing for more than 1 attachment. The dataModelBinding is used to add the attachment UUID to the form data, so that the target system can know which attachment belongs to which row of the repeating group.

As for code review, @nkylstad and @lorang92 have already been somewhat introduced to the feature. This change contains a lot more code than I anticipated at first, so I'm happy to assist and explain as needed - I'm also keen to experiment with a more pair-programming-like approach to code review if anyone wants.

## Related Issue(s)
- altinn/altinn-studio/issues/7659
- #128 (rediscovered and fixed while developing this feature)

## Functionality not yet implemented:
- Altinn/altinn-studio/issues/8664

## Verification
- [x] All tests run green

## Documentation
- [x] User documentation is updated with a separate linked PR in [altinn-studio-docs.](https://github.com/Altinn/altinn-studio-docs) (if applicable)
  - Altinn/altinn-studio-docs/pull/550


src/altinn-app-frontend/src/store/index.ts
```diff
@@ -25,6 +25,11 @@ export const setupStore = (preloadedState?: PreloadedState<RootState>) => {
 
 export const store = setupStore();
 
+// Expose store when running in Cypress
```
haakemon, line 28: Why do we need to do this?

olemartinorg: Have a look here:  
[https://github.com/Altinn/app-frontend-react/blob/5430ad42b31b0edd327dd992b22237e6164d862d/test/cypress/e2e/support/app-frontend.js#L141-L149](https://github.com/Altinn/app-frontend-react/blob/5430ad42b31b0edd327dd992b22237e6164d862d/test/cypress/e2e/support/app-frontend.js#L141-L149)  
  
And for examples of usage: [https://github.com/Altinn/app-frontend-react/blob/5430ad42b31b0edd327dd992b22237e6164d862d/test/cypress/e2e/integration/app-frontend/attachments-in-group.js#L412](https://github.com/Altinn/app-frontend-react/blob/5430ad42b31b0edd327dd992b22237e6164d862d/test/cypress/e2e/integration/app-frontend/attachments-in-group.js#L412)  
  
Basically, testing the expected redux state directly lets me write a more robust test, as it doesn't test the effect of known/expected bugs, but tests the redux state where errors will cause the bugs. When reloading the page and reading data from the instance, `mapAttachments()` runs in order to find all attachments in `formData` to rebuild the state. By reading the state, the e2e test can assert that the state is the same as it was before a reload (thus testing that `mapAttachments()` work).  I could also have done this by exhaustively testing all reasonable thinkable effects of a broken state, but since they are (in principle) unlimited, testing the state directly is far more robust.

src/altinn-app-frontend/src/features/form/layout/index.ts
```diff
@@ -57,10 +57,34 @@ export type ComponentTypes =
...
+ */
+export interface IDataModelBindingsList {
+  list: string;
+}
+
+export interface IDataModelBindingsForAddress {
+  address: string;
+  zipCode: string;
+  postPlace: string;
 }
```
lorang92, line 80:
```suggestion
export interface IDataModelBindingsForAddress {
  address: string;
  zipCode: string;
  postPlace: string;
  careOf: string;
  houseNumber: string;
}
```

src/altinn-app-frontend/src/features/form/layout/index.ts
```diff
@@ -57,10 +57,34 @@ export type ComponentTypes =
...
 
+export type IDataModelBindings =
```
lorang92, line 82: We are missing some types here if we want to go this route. For instance the `likert` binding for our new likert component.

src/altinn-app-frontend/src/store/index.ts
haakemon: Okay, thanks for the explanation. Could you try to sum up the "why" in the comment (not just the "what"), that would be valuable for later I think 🙂

src/altinn-app-frontend/src/selectors/getErrors.ts
```diff
@@ -1,13 +1,23 @@
...
+let lastLogged:Error|null = null;
+
 /**
  * Selector for determining if we have an error in one of our api calls.
  * Returns true any errors is set in relevant states, false otherwise
  * @param state the redux state
  */
 const getHasErrorsSelector = (state: IRuntimeState) => {
-  const error = (
+  const exceptIfIncludes = (maybeError:Error|null, lookFor:string):Error|null => {
```
lorang92, line 12: Just a note on the selector here in general, it appears we are not using reselect correctly. We are doing the calculation in the inputSelector and not the resultFunc, from docs:

`createSelector(...inputSelectors | [inputSelectors], resultFunc, selectorOptions?)
` (https://github.com/reduxjs/reselect#api)

See that this is not something you introduced. I'm guessing we probably have more issues like this.

test/cypress/e2e/integration/app-frontend/attachments-in-group.js
```diff
@@ -0,0 +1,462 @@
...
+          [
+            'nested-row1-sub1-1.pdf',
+            'nested-row1-sub1-2.pdf',
+            'nested-row1-sub1-3.pdf',
+          ]
+        ],
+      }
+    ];
+
+    uploadFile(appFrontend.group.rows[0].uploadSingle, 0, filenames[0].single, true);
```
haakemon, line 179: I would really prefer if we used [RORO pattern](https://www.freecodecamp.org/news/elegant-patterns-in-modern-javascript-roro-be01e7669cbd/) here (and in general when accepting multiple arguments) - makes it so much easier to read and understand what the arguments are when calling the function. What does `0` and `true` mean here f.ex? We have to go to the definition to get the answers

Using RORO would also make it possible to call the function with `isTaggedUploader` set to true, without having to specify false for `verifyTableRow`.


```suggestion
    uploadFile({
	item: appFrontend.group.rows[0].uploadSingle, 
	attachmentIndex: 0, 
	fileName: filenames[0].single,
	verifyTableRow: true
    });
```

src/altinn-app-frontend/src/utils/formComponentUtils.ts
```diff
@@ -200,12 +215,31 @@ export const getDisplayFormData = (
       });
       return label;
     }
+    if (isFileUploadComponent(component) || isFileUploadWithTagComponent(component)) {
```
lorang92, line 218: Some unit tests for this whole block would be sweet!

haakemon (approved):
LGTM 🙂 as mentioned I would prefer changing to RORO, but not a blocker

lorang92 (approved):
This is some seriously nice work. What a way to get to know the codebase 👏 !

I left some minor comments for your consideration. I don't think any of them is a real blocker, but please have a look.

src/altinn-app-frontend/src/store/index.ts
olemartinorg: Yup, I agree! For this specific case I guess I didn't want to tie it to "why" in the context of attachments in repeating groups, but I can update with a better comment (although generic).

src/altinn-app-frontend/src/features/form/layout/index.ts
olemartinorg: Last two should be optional then, maybe? I haven't seen those before.

src/altinn-app-frontend/src/selectors/getErrors.ts
haakemon: Good catch, I missed this.

I think we have several cases of incorrect usage of reselect - perhaps we should create a separate task to clean them up?

src/altinn-app-frontend/src/features/form/layout/index.ts
lorang92: Ah, yes! Good point.

test/cypress/e2e/integration/app-frontend/attachments-in-group.js
olemartinorg: Yup, I totally get it. In IntelliJ, hints on parameter names are built-in (so I rarely have the same need for this) but I rewrote to using RORO for this function now.

src/altinn-app-frontend/src/features/form/layout/index.ts
olemartinorg: Huh, that should have produced typescript errors then? I haven't had any, and searching for `.likert` doesn't yield me any code accessing that property.

lorang92: Ah, you are absolutely right. In an earlier PR for the likert component we had a `likert` binding, but I see that this was changed. Also checked docs to be sure: https://docs.altinn.studio/app/development/ux/components/likert/

Sorry for the confusion :)

src/altinn-app-frontend/src/utils/formComponentUtils.ts
olemartinorg: Done! :tada:

Issue #234, opened 2022-06-17
Make it possible for PR's from forks to run the SonarCloud scan


haakemon:
### Description

Fork PR's cannot run the scan, because forks do not have access to `secrets` (except `GITHUB_TOKEN`).

SonarCloud scan requires access to `SONAR_TOKEN` secret.

It should still be possible to run the scan, without forks being given access to the `SONAR_TOKEN` secret. There are some details in this blog post https://securitylab.github.com/research/github-actions-preventing-pwn-requests/ , and an example implementation here https://community.sonarsource.com/t/how-to-use-sonarcloud-with-a-forked-repository-on-github/7363/30 that could be useful. But its important to follow the steps and warnings in the blogpost so the secrets are not accidentally exposed.

### Additional Information

This is the error message you will get when its trying to run
![image](https://user-images.githubusercontent.com/2082481/174328414-1e47205c-525b-4e7b-8eed-da92d36491fa.png)

haakemon:
I looked into this, and while I got things to work (a build workflow that collects code coverage, stores it as a workflow artifact, then triggers a privileged workflow which downloads the coverage artifacts and ships it off to SonarCloud), I'm not really sure its the best approach after all.

Some drawbacks
- The complexity is much greater
- The separate triggered workflow loses the connection to the PR, which makes the Actions list a bit confusing and messy to look at
- The separate triggered workflow doesnt appear in the PR "checks list" until the action is complete which can take a while
  - Its possible to manually trigger some checks here to make the feedback loop shorter, but that adds even more complexity
- To get a correct code coverage report, we would need to checkout the PR code in the privileged workflow as well. I didnt look into this in my investigation, but its certainly doable.

I think the time is better spent to implement code coverage checks with other tools, f.ex [Codecov ](https://github.com/marketplace/actions/codecov)which is [free for open source](https://about.codecov.io/pricing/), and does not require tokens to publish the report for public repos. We would lose out on some niceties with SonarCloud, f.ex code smell reports - but perhaps other tools also can supply this information?

PR #255, opened 2022-06-27
chore: move sonarcloud check to separate workflow so its possible to run on external PRs

haakemon:
## Description
External PR's cannot run the SonarCloud scan, because they do not have access to the `SONAR_TOKEN` secret. To fix this, we can separate the workflows into multiple parts, where the actual scan will run in a trusted context.

- Build will now generate an artifact with code coverage results instead of running SonarCloud directly. This workflow will no longer need or have access to `SONAR_TOKEN`.
- A separate workflow is triggered after build, which will do the SonarCloud scan. It downloads the artifacts containing the code coverage, and analyzes it in SonarCloud. This workflow will have access to the `SONAR_TOKEN`.

Related blog post https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

## Related Issue(s)
- #234 

## Verification
- [ ] All tests run green


haakemon:
I dont think this approach is a good idea after all. I've updated the PR with some comments about missing details, and updated the issue with details about why this approch probably isnt the best.

Issue #300, opened 2022-06-29
Add Codecov (or similar tool) to workflow to check coverage for fork PR's


haakemon:
### Description

See details in #234 why we need a different tool than SonarCloud for forks (TLDR: fork PRs do not have access to secrets).

Codecov is a tool that does not require any secrets for public repos, and is also free for open source projects. https://github.com/marketplace/actions/codecov

It may be valuable to add this to our workflow to ensure fork PRs also fulfill our internal test coverage requirements.

PR #301, opened 2022-06-29
chore: disable SonarCloud scan for forks

haakemon:
## Description
Builds from forks always result in a failed check, because they do not have access to the `SONAR_TOKEN` secret. This is very annoying. We investigated possibilities to be able to run this check in a separate workflow which had access to the secret in #255 , but it turned out to be a bit complex, and not worth the trouble.

We should instead disable SonarCloud scan on forks, and if we want coverage checks (or other checks) from forks, we should add other tools to our pipeline (tools which ideally do not require any secrets, because then we get into the same problem again)

## Related Issue(s)
- #234 
- #300

## Verification
- [x] All tests run green

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -41,6 +41,7 @@ jobs:
         run: yarn test --coverage
 
       - name: SonarCloud Scan
+        if: github.event.repository.fork == false
         with:
           projectBaseDir: app-frontend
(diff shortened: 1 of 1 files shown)


olemartinorg (approved):
SonarCloud seems to be happy about this change, and so am I. :wink:

PR #303, opened 2022-06-30
chore: fix unstable tests

haakemon:
## Description
We recently had some unstable tests, due to tests taking more than the allotted time to execute (10 seconds pr test). This was caused by using `getByRole` query, which is not great on a large DOM tree (~600 nodes) in a loop within a loop. The slowest test in this group now takes ~3500ms which is still not _great_, but I think good enough for now 🙂 When running the whole test suite, it takes longer (6-7000ms), which is why I have kept the increased jest timeout of 10000ms.

- Decreased test data set
	-  This is to reduce the generated DOM size, to improve the general test performance
- Changed from `getByRole` to `getByDisplayValue` in the innermost loop to improve test performance
- Fixed issue in test data where answers were added with incorrect casing in the key. This was not caught by any tests, because of a missing `expect` statement.
	- Add function to get "questions with answer", to ensure that answers are added with the proper key with correct casing (`Answer`)
	- Add extra expect to validateRadioLayout which will fail the test when answers are added with incorrect key casing (`answer`)

## Verification
- [x] All tests run green

### DIFF
--- a/src/altinn-app-frontend/setupTests.ts
+++ b/src/altinn-app-frontend/setupTests.ts
@@ -1,12 +1,12 @@
-import Enzyme from "enzyme";
-import Adapter from "@wojtekmaj/enzyme-adapter-react-17";
-import "jest";
-import "@testing-library/jest-dom/extend-expect";
+import Enzyme from 'enzyme';
+import Adapter from '@wojtekmaj/enzyme-adapter-react-17';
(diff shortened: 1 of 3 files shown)


olemartinorg (approved):
Looks great! Thanks for taking the time to make this more stable! :tada:

Issue #257, opened 2022-06-16
Display of checkbox value on summarypage is wrong when the checkbox is in the outer group of a "group in group" setup.


RogerKjaernsrod:
### Description of the bug

Display of checkbox value on summarypage is wrong when the checkbox is in the outer group of a "group in group" setup.
When you add the group that contains another group to the summary page like this:
{
  "id": "SummaryComp1",
  "type": "Summary",
  "componentRef": "MainGroup1",
  "pageRef": "7_objekt",
  **"largeGroup": true**
}
and the outer group ("MainGroup1" here) contans a checkbox component, the value is displayed like this:
![image](https://user-images.githubusercontent.com/47383888/174006846-46c08444-06f7-4265-aefa-c77fadd8f3d4.png)
It is displayed like this on the "filling out" page:
![image](https://user-images.githubusercontent.com/47383888/174006960-34ae20f4-9bcc-42da-b799-94e635a0f545.png)

The bug can be seen on the Økokrim app mt-3:
https://oko.apps.tt02.altinn.no/oko/mt-3/

### Steps To Reproduce

Create an app that has a checkbox in the outer group of a group in group setup.
Add the checkbox on the summary page.

olemartinorg:
@RogerKjaernsrod The fix for this should now have landed in the latest frontend version (v3.44.3 and later), could you test it to make sure it works? Let me know it anything unexpected happens, or close the issue. Thank you! 😊

RogerKjaernsrod:
It works.

PR #310, opened 2022-07-01
Rendering checkboxes summary as string, when string is provided

olemartinorg:
## Description
The summary page could end up rendering `Checkboxes` components by iterating each character and listing them line-by-line. This happens when the `formData` is provided as a string.

This quick-fix just renders the string instead, but we could also go the extra mile by hunting down why this happens and make sure `Checkboxes` components are always summarized with list inputs - although it is stored in `formData` as a comma-separated string.

## Related Issue(s)
- #257

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/altinn-app-frontend/src/utils/formLayout.ts
+++ b/src/altinn-app-frontend/src/utils/formLayout.ts
@@ -18,6 +18,7 @@ import type {
   ILayoutGroup,
 } from '../features/form/layout';
 import type { IDatePickerProps } from 'src/components/base/DatepickerComponent';
+import type { ICheckboxContainerProps } from 'src/components/base/CheckboxesContainerComponent';
 
 interface SplitKey {
   baseComponentId: string;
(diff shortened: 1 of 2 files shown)


src/altinn-app-frontend/src/utils/formLayout.ts
```diff
@@ -468,3 +469,11 @@ export function isDatePickerComponent(
 ): component is IDatePickerProps & ILayoutComponent {
   return component.type.toLowerCase() === 'datepicker';
 }
+
+export function isCheckboxesComponent(
```
haakemon, line 473: Add some tests? 😇

olemartinorg: Yup, I can do that, but I prefer to do it in a later PR. There are some changes to these functions in the [hackaton/dynamics](https://github.com/Altinn/app-frontend-react/compare/main...hackaton/dynamics) branch which I'd like to port over, and I've thought about possibly changing up some of the typings. The component types are a prime example of a case where [discriminated unions](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#discriminated-unions) works wonders - so with a change in that direction, these functions could be removed altogether, as TypeScript would give us the same functionality out of the box. I'll add it to my list of sabbatical TODOs.. :wink:

haakemon (approved):
LGTM (with the promise of added tests/simplified code which doesnt require tests to come during a sabbatical 😅)

Issue #312, opened 2022-07-04
Altinn 3 apps does not work on Safari


haakemon:
### Description of the bug

We recently had an issue where we introduced some [regex lookbehind, which is not supported in Safari](https://caniuse.com/js-regexp-lookbehind) (both mobile and on mac). This caused all apps to not work on Safari.

### Steps To Reproduce

1. Go to any application in a Safari browser
2. App will not load

PR #314, opened 2022-07-04
Fixing code using lookbehind + lookahead, for Safari

olemartinorg:
## Description
The offending code caused Safari to crash on load.

## Related Issue(s)
- #312

## Verification
- [x] All tests run green

## Documentation

### DIFF
--- a/src/altinn-app-frontend/src/utils/databindings.test.ts
+++ b/src/altinn-app-frontend/src/utils/databindings.test.ts
@@ -8,6 +8,7 @@ import {
   getKeyWithoutIndex,
   mapFormData,
   removeGroupData,
+  getKeyIndex,
 } from './databindings';
 
 describe('utils/databindings.ts', () => {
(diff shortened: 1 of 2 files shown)


lorang92 (approved):
Neat 🐛

src/altinn-app-frontend/src/utils/databindings.test.ts
```diff
@@ -111,6 +112,12 @@ describe('utils/databindings.ts', () => {
     ];
   });
 
+  describe('getKeyIndex', () => {
+    it('should work', () => {
```
haakemon, line 116: Would suggest a more descriptive title 🙂

f.ex like this? Or something else if you can come up with something 😁
```suggestion
    it('should return key indexes from string', () => {
```

olemartinorg: Right, I agree! I blame the rush.. ;-) Fixing this later.

Issue #311, opened 2022-06-16
Dynamic rules is not working for repeating group in repeating group on summary page.


RogerKjaernsrod:
### Description of the bug

When you have dynamic rules on elements in a repeating group in repeating group the summary page shows hidden elements that should be hidden.
This image shows a rep group in a rep group were the fields for E-post, Nettsted, IP-adresse an Annet are hidden.
Only fields for telefon are visible:
![image](https://user-images.githubusercontent.com/47383888/174054236-3ede33a7-ade4-4b7a-a9bc-2abef12b61a7.png)
This is displayed like this on the summary page:
![image](https://user-images.githubusercontent.com/47383888/174054435-dfb99bfa-b335-4440-a6c5-37cf559080e7.png)

Another example for the same group in group:
![image](https://user-images.githubusercontent.com/47383888/174056105-647961be-66ec-42e9-98c1-b787ea8f8f79.png)

On Summary page:
![image](https://user-images.githubusercontent.com/47383888/174056214-feb128b0-1780-4990-b230-23fa9fb50127.png)

It should have been shown like this:
![image](https://user-images.githubusercontent.com/47383888/174056458-50ab1b99-1573-4ba4-b818-be8b432fc177.png)

### Steps To Reproduce

Create app with dynamic rules on elements in a repeating group in repeating group.
Show the outer group on the summary page.
Elements that should be hidden are visible.

RogerKjaernsrod:
@olemartinorg The fix is working as expected. Thanks.

PR #315, opened 2022-07-04
Bug: Dynamics/hidden fields in summaries of nested repeating groups

olemartinorg:
## Description
This fix involves three distinct changes:

1. Using a Set instead of iterating the hidden fields to reduce it to a subset. While the group children filter works, it does not filter nested children, so this didn't seem to have the desired effect on nested groups. I assume the filtering is done for performance reasons, so using a Set for this seems more ideal (lookups are cheap, at least).
2. Making sure we check for the full nested dashed path when testing if the field should be hidden. Without this, nested dynamics have no effect when showing a summary.
3. Moving the code so that loops continue earlier when hidden fields are found. Previously it would run the iteration almost to the end before scrapping the result if the field was hidden, which is a bit wasteful. This shouldn't have an effect on the core problem, but should slightly improve performance.

## Related Issue(s)
- #311

## Verification
- [x] All tests run green

## Documentation

### DIFF
--- a/test/cypress/e2e/integration/app-frontend/summary.js
+++ b/test/cypress/e2e/integration/app-frontend/summary.js
@@ -91,7 +91,7 @@ describe('Summary', () => {
       });
   });
 
-  it('is possible to view summary of repeating group', () => {
+  it.only('is possible to view summary of repeating group', () => {
     cy.completeTask3Form();
     cy.get(appFrontend.group.mainGroupSummary)
       .should('be.visible').and('have.length', 1)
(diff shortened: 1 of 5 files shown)


lorang92 (approved):
LGTM. Nice catch 🐛

test/cypress/e2e/integration/app-frontend/summary.js
```diff
@@ -91,7 +91,7 @@ describe('Summary', () => {
       });
   });
 
-  it('is possible to view summary of repeating group', () => {
+  it.only('is possible to view summary of repeating group', () => {
```
haakemon, line 94: @olemartinorg This should be removed 🙂

olemartinorg: Always such a great eye, @haakemon! I added a new PR to fix this, in #333, and hopefully prevented such issues in the future (it.only is great when isolating a test issue, but I've seen it being committed to git before by mistake - should be easy to catch in a pre-commit hook).

PR #320, opened 2022-07-05
Support for custom components (web components)

nkylstad:
## Description
Support for adding custom web components to form using formLayout file.
See related issue below for documentation on how to set up a custom web component, as well as overview of what was done and what remains to be done/is not supported.

## Related Issue(s)
- https://github.com/Altinn/altinn-studio/issues/8681

## Verification
- [ ] All tests run green

## Documentation


src/altinn-app-frontend/src/components/custom/CustomComponent.tsx
```diff
@@ -0,0 +1,42 @@
...
+  framework,
+  handleDataChange,
+  ...passThroughProps
+}: ICustomComponentProps) {
+  if (framework === CustomSupportedFrameworks.React) {
+    return React.createElement(tagName, {
+      tagName,
+      handleDataChange,
+      ...passThroughProps,
+    });
```
ivarne, line 27: When would you be able to use this? If I understand `React.CreateElement` correctly, it either takes a string for a standard html tag (eg: `input`), or a react component (function that returns react elements). Is this for using standard html tags? or should you define a render function in FormLayout.json?

nkylstad: Yep, some leftovers from POC'ing, trying to import a react component from external source, that I never got around to testing. Removing this, since it's not usable at the moment, and sticking with just web components for now.

src/altinn-app-frontend/src/components/custom/CustomWebComponent.tsx
```diff
@@ -0,0 +1,94 @@
...
+    current.componentValidations = componentValidations;
+  }, [formData, componentValidations]);
+
+  if (!Tag || !textResources) {
+    return null;
+  }
+
+  const propsAsAttributes: any = {};
+  Object.keys(passThroughProps).forEach((key) => {
+    propsAsAttributes[key] = JSON.stringify(passThroughProps[key]);
```
olemartinorg, line 62: Won't this mean that simple string values from the form layout gets wrapped in double-quotes? Might get a bit annoying on the receiving end. I get that we can't pass complex objects as HTML attributes, but maybe we should filter those out (or assign them to the ref like with dataModelBindings etc above). Simple values like string/number/boolean could be passed through without JSON encoding.

olemartinorg (approved):
I added comment/question above. Great job with this one, this is really exciting stuff! :1st_place_medal:

src/altinn-app-frontend/src/components/custom/CustomWebComponent.tsx
nkylstad: Hmm yep, you're right. Didn't think of that. Will do a filtering for more complex objects/arrays, and leave the rest as-is.

Issue #225, opened 2022-06-16
Browser back button do not apply to app pages


Febakke:
### Description of the bug

When using an app with multiple pages you can not go back and forth with the browsers back/forward buttons. Im guessing this is because we are actually not actually loading in a new page. If the user press the back button they will be moved to the page they were before starting the app. Im guessing most people will expect going one page back when using that button.

Im not sure if this is directly a WCAG issue, but i'm guessing that having expected behavior on the browser button will be better for everyone. 

### Steps To Reproduce

1. Go to an app with multiple pages
2. Go to page two
3. Try the browser back button

### Additional Information

Tested with the "Oppstartsveileder" application 
On Mac with Chrome

bjosttveit:
Fixed by https://github.com/Altinn/app-frontend-react/pull/1682

PR #325, opened 2022-07-05
Bug/225 browser back button not navigating in app

mjulstein:
## Description
I changed the behavior of how the form url changes where it now puts the page id at he end of the route.
By doing so we get to listen to history-changes and re-render the Form component based on the page id inside the formLayout.

Tested in Firefox.

## Related Issue(s)
- #225 

## Verification
- [x] All tests run green

## NOTE
The below steps (in Norwegian) do reproduce am error. However, the issue seems to be related to not being able to access pages after the allowed pages after the page order has been updated or switching layout sets:

> 1. Du har tidligere instanser, velg en fra listen eller start en ny. Jeg startet en ny.
>  2. Velkommen til frontend-test, noen beskrivelser, en knapp for å starte. Jeg trykker på knappen. (Tror URL sier /formLayout på dette tidspunktet)
>  3. Skjema for å endre navn. Jeg fyller ut skjemaet.
> 4. Oppsummeringsside for endring av navn. Jeg trykker neste.
> 5. Side for repeterende grupper.
> 6. Side for å fylle ut et tekstfelt (innmelder?).
> 7. Oppsummeringsside for repeterende grupper.
> Hvis du står på steg 4 og trykker tilbake, kommer du til steg 3, men hvis du står på steg 3 og går tilbake fikk jeg en feilmelding (appen kræsjet). Samme hvis jeg stod på steg 5, 6 eller 7 og prøvde å gå tilbake til steg 4.

### DIFF
--- a/src/altinn-app-frontend/src/App.tsx
+++ b/src/altinn-app-frontend/src/App.tsx
@@ -1,5 +1,5 @@
-import React from 'react';
-import { Route, Switch } from 'react-router-dom';
+import * as React from 'react';
+import { Route, Switch, useLocation, useRouteMatch } from 'react-router-dom';
 
 import { useAppDispatch, useAppSelector } from 'src/common/hooks';
 import Entrypoint from 'src/features/entrypoint/Entrypoint';
@@ -29,9 +29,11 @@ export const App = () => {
 
   const allowAnonymousSelector = makeGetAllowAnonymousSelector();
   const allowAnonymous = useAppSelector(allowAnonymousSelector);
-
   const [ready, setReady] = React.useState(false);
-
+  const instancePath = '/instance/:partyId/:instanceGuid';
+  const match = useRouteMatch<string>(instancePath);
+  const matchUrl = match?.url || '';
+  const location = useLocation();
   React.useEffect(() => {
     function setUpEventListeners() {
       window.addEventListener('mousemove', refreshJwtToken);
@@ -92,23 +94,21 @@ export const App = () => {
   if (!ready) {
     return null;
   }
-
   return (
     <Switch>
       <Route
-        path='/'
-        exact={true}
+        path={`/`}
+        exact={!!matchUrl || location.pathname.includes('/partyselection/')}
       >
         <Entrypoint allowAnonymous={allowAnonymous} />
       </Route>
       <Route
         path='/partyselection/:errorCode?'
-        exact={true}
+        exact
         component={PartySelection}
       />
       <Route
-        path='/instance/:partyId/:instanceGuid'
-        exact={true}
+        path={instancePath}
         component={ProcessWrapper}
       />
     </Switch>
--- a/src/altinn-app-frontend/src/common/hooks/useFormLayoutHistory.ts
+++ b/src/altinn-app-frontend/src/common/hooks/useFormLayoutHistory.ts
@@ -1,0 +1,117 @@
+import { useCallback, useEffect } from 'react';
+import { useHistory, useRouteMatch } from 'react-router-dom';
+
+import { useAppDispatch } from 'src/common/hooks/useAppDispatch';
+import { FormLayoutActions } from 'src/features/form/layout/formLayoutSlice';
+import type { IUpdateCurrentView } from 'src/features/form/layout/formLayoutTypes';
+
+enum HistoryAction {
+  BackOrForward = 'POP',
+}
+
+/**
+ * @param activePageId the formLayout page id that is currently active in the app
+ * @return matchRootUrl: Matches the location pathname of the history root. If not, an empty string.
+ */
+export const useFormLayoutHistoryAndMatchInstanceLocation = ({
+  activePageId,
+}: {
+  activePageId: string;
+}): { matchRootUrl: string } => {
+  const matchRoot = useRouteMatch<string>('/instance/:partyId/:instanceGuid');
+  const matchRootUrl = matchRoot?.url || '';
+  const dispatch = useAppDispatch();
+  const doDispatch = useCallback(
+    (payload: IUpdateCurrentView) => {
+      dispatch(FormLayoutActions.updateCurrentView(payload));
+    },
+    [dispatch],
+  );
+  useFormLayoutHistory({
+    renderedPageId: activePageId,
+    renderFunc: doDispatch,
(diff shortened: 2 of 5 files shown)


src/altinn-app-frontend/src/features/form/containers/Form.tsx
```diff
@@ -143,8 +140,34 @@ export function Form() {
...
+      return history.listen((_location, action) => {
+        if (action === 'POP') {
+          const pageId = getPageId();
+          if (pageId && pageId !== currentLayout) {
+            dispatch(FormLayoutActions.updateCurrentView({ newView: pageId }));
+          }
+        }
+      });
+    }
+  }, [dispatch, currentLayout, matchUrl, history]);
```
haakemon, line 167: I would suggest to extract this to a custom hook. That should make it easier to test this in isolation, reduce the cognitive load when reading the Form component, and it might help with understanding the meaning behind this with a descriptive name.

```suggestion
  const {matchUrl} = useHistoryReplaceWhenUrlMatchInstance({currentLayout}); // This name could probably be even better, it was just what I could think of at the top of my head 😅
```


Then extract the content to a separate hook:
```typescript
	
export const useHistoryReplaceWhenUrlMatchInstance = ({currentLayout}) => {
  const dispatch = useAppDispatch();
  const match = useRouteMatch<string>('/instance/:partyId/:instanceGuid');
  const matchUrl = match?.url || '';
  const history = useHistory();

  useEffect(() => {
    if (currentLayout) {
      const pageId = history.location?.pathname.replace(`${matchUrl}/`, '');
      const isOnRootPage = !pageId || pageId === matchUrl;

      if (isOnRootPage) {
        history.replace(`${matchUrl}/${currentLayout}`);
      } else if (`${currentLayout}` !== pageId) {
        history.push(`${matchUrl}/${currentLayout}`);
      }

      return history.listen((_location, action) => {
        const isBrowserBackOrForwardPressed = action === 'POP';

        if (isBrowserBackOrForwardPressed) {
          if (pageId && pageId !== currentLayout) {
            dispatch(FormLayoutActions.updateCurrentView({ newView: pageId }));
          }
        }
      });
    }
  }, [dispatch, currentLayout, matchUrl, history]);

  return {matchUrl};
}


  

```

haakemon (changes requested):

src/altinn-app-frontend/src/App.tsx
```diff
@@ -97,6 +101,9 @@ export const App = () => {
   }
 
   if (!ready) {
+    if (hasInstanceId) {
+      return <div>Not ready, but got instanceId.</div>;
```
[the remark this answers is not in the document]
mjulstein: No, maybe I should just add an empty div with a testid instead?

src/altinn-app-frontend/src/common/hooks/useFormLayoutHistory.ts
```diff
@@ -0,0 +1,61 @@
...
+  const matchRootUrl = matchRoot?.url || '';
+  const history = useHistory();
+  const dispatch = useAppDispatch();
+  const dispatchAction = FormLayoutActions.updateCurrentView;
+  const doDispatch = useCallback(
+    (payload: IUpdateCurrentView) => {
+      dispatch(dispatchAction(payload));
+    },
+    [dispatch, dispatchAction],
+  );
```
olemartinorg, line 27: I don't see the point in declaring a new dispatch function here. You don't have to pass `dispatchAction` as a dependency to `useCallback`, it's fine just calling `dispatch(FormLayoutActions.updateCurrentView({...}))` directly. With all of this indirection it just makes the code harder to follow, IMO.

src/altinn-app-frontend/src/common/hooks/useFormLayoutHistory.ts
```diff
@@ -0,0 +1,61 @@
...
+  useEffect(() => {
+    const pageIdFromLocation = history.location.pathname.replace(
+      `${matchRootUrl}/`,
+      '',
+    );
+    const locationIsRoot =
+      !pageIdFromLocation || pageIdFromLocation === matchRootUrl;
```
olemartinorg, line 34: Could this be simplified with `history.location.pathname.startsWith(matchRootUrl)` in some way instead?

src/shared/src/utils/instanceIdRegExp.ts
```diff
@@ -0,0 +1,23 @@
+const instanceIdRegExp =
+  /(\d{1,10}\/[\d,a-f]{8}-[\d,a-f]{4}-[1-5][\d,a-f]{3}-[89ab][\d,a-f]{3}-[\d,a-f]{12})/i;
```
olemartinorg, line 2:
```suggestion
  /(\d{1,10}\/[\da-f]{8}-[\da-f]{4}-[1-5][\da-f]{3}-[89ab][\da-f]{3}-[\da-f]{12})/i;
```
I'm assuming we're matching a UUID here, and that it shouldn't be matching commas? This regex would also match the string `123456/7515,373-aed4-41f7-95b4-e5b5115c2edc` (notice the comma).

Also not quite sure where this regex came from, but I can't imagine it being important that some of these parts of the UUID only start with `[1-5]` and `[89ab]`? If not, it can be simplified further:

```suggestion
  /(\d{1,10}\/[\da-f]{8}-[\da-f]{4}-[\da-f]{4}-[\da-f]{4}-[\da-f]{12})/i;
```

olemartinorg (changes requested):
Great job in cleaning up cognitive complexity and such along the way! :1st_place_medal: This looks good, but I still think we should find a way to solve the usability issue you noted in the initial description - I'm not sure we should be releasing this early at a stage where it causes (apparent) app crashes when clicking the browser back button in some apps.

src/altinn-app-frontend/src/common/hooks/useFormLayoutHistory.ts
mjulstein: rewriting.
pageIdFromLocation is either a pageId (all will be false), `''`og the `matchRootUrl`. 
So `startsWIth` would no really help.

olemartinorg:
Closing this, but keeping the branch for a while. The changes here are similar to #451, which we intend to keep (but have to rewrite to keep up with the ever-changing main).

PR #328, opened 2022-07-06
implement text field from design system

Magnusrm:
Implement Input component from design system in place of material-UI input component.
Should not be merged until input component is merged and released in a package.

## Todo
- [x] Merge input component and create new release of altinn-design-system. https://github.com/Altinn/altinn-design-system/pull/61
- [x] Update to the new altinn-design-system version in altinn-frontend-react.
- [x] Discuss look of AdressComponent fields (especially postplace look)
	- [x] There is also some missing margin-bottom after updating, this needs to be added somewhere

## Verification
- [x] All tests run green

## Documentation


src/altinn-app-frontend/src/components/advanced/AddressComponent.tsx
```diff
@@ -289,35 +288,35 @@ export function AddressComponent({
...
+          ? renderValidationMessagesForComponent(
+              allValidations[AddressKeys.address],
+              `${id}_${AddressKeys.address}`,
+            )
+          : null}
+      </div>
+
       {!simplified && (
-        <>
+        <div>
```
olemartinorg, line 319: Any reason for converting this fragment to a div? I prefer using fragments instead of divs (to avoid cluttering the DOM) unless we're actually styling the div somehow. And if we're styling it, we should be adding a class to it. Maybe it's just a question of personal taste..

src/altinn-app-frontend/src/components/advanced/AddressComponent.tsx
```diff
@@ -326,12 +325,9 @@ export function AddressComponent({
...
             labelSettings={labelSettings}
           />
-          <input
+          <TextField
             id={`address_care_of_${id}`}
-            className={cn('form-control', {
-              'validation-error': allValidations.careOf.errors.length,
-              disabled: readOnly,
-            })}
+            isValid={allValidations.careOf.errors.length === 0}
```
olemartinorg, line 330: Did we lose the `readOnly` property in the process of rewriting this?

src/altinn-app-frontend/src/components/advanced/AddressComponent.tsx
olemartinorg: Same thing on line 403

olemartinorg (approved):
LGTM! :partying_face: Make sure the `readOnly` parameter gets passed that one place, other than that my comments are of no particular importance.. :wink:

src/altinn-app-frontend/src/components/advanced/AddressComponent.tsx
haakemon: Yes, the reason is to group the items properly so we can take advantage of the `gap` css property that is set on the common parent. This makes it easier to lay out the items with correct spacing between them. The old implementation used `margin-bottom` instead, which caused extra space before the next content. This could f.ex be solved by setting no margin on the last item instead, but using `gap` is a lot cleaner. The only "downside" is that the items need to be grouped properly.

src/altinn-app-frontend/src/components/advanced/AddressComponent.tsx
haakemon: Its still there, just a few lines down (not modified) 🙂

Its just the classname that was removed, which is correct.

olemartinorg: Haha, I didn't notice there were non-modified lines! Thanks! :see_no_evil:

src/altinn-app-frontend/src/components/advanced/AddressComponent.tsx
olemartinorg: Ah - so you'd see a visual difference if you'd try removing them, then. Fine by me! :+1:

Febakke:
@haakemon 
Testet på inputfelter og adressekomponenten. Er det andre komponenter som også burde testes her?

Fant en WCAG feil 🚨
Vi har ikke spesifisert dette i skissene, men oppdaget en feil her. Read-only felter har ingen focus visning. Spesielt når feltene er tomme vil det ikke er være noe indikasjon på hvor du er i skjema når du navigerer med tastatur. Dette er et brudd på 2.4.7: Focus visible. 

Foreslår at vi bruker den samme blå rammen som andre komponenter får når de får fokus. Det er jo også en diskusjon om slike tekstfelter skal få fokus i tastaturnavigasjonen, men så lenge de får fokus må de ha fokusvisning. 

**Hva som er testet:** 
Tastaturnavigasjon: Ok 
Skjermleser: Ok
Kontraster: Ok
Små skjermer: Ok
Text-spacing: Ok
Forstørret skjermen: Ok
Generell styling: Som spesifisert, men ser vi mangler å designspesifikasjon fokus visning av felter som feiler. Design tar en runde på dette.

haakemon:
@Febakke 
Godt poeng - jeg tror de bør være "focusable", det er slik de oppfører seg som standard. Tror ikke det er lurt å tukle med det.

- [ ] Legg til ramme ved focus.

Issue #196, opened 2020-11-26
Replace existing redux setup


nkylstad:
## Description
In our applications, we use a "traditional" setup for redux, with actions, actionTypes, dispatchers, reducers, all defined explicitly in separate files. This results in a lot of boilerplate code, and also increases the complexity of the applications greatly, especially with regards to new developers who want to get started.

About a year ago, redux introduced redux-toolkit, which is a library that can be used to simplify the setup of redux. I have tested using this with an existing feature that has the traditional setup. The result was that a single file could replace the files for actions/actionTypes/dispatchers/reducers (which are all mostly boilerplate). In addition, the new setup makes it much clearer how what the code flow is, and what the actions actually do.

I suggest we replace all of our existing redux setup with a setup using redux-toolkit. 

## Screenshots
> Screenshots or links to Figma (make sure your sketch is public)

## Considerations
We should make this change in altinn studio first, to get used to the setup and iron out any wrinkles that might appear. Then we can make the same change to app frontend.

As this new setup can perfectly well work in parallel with the existing setup, we suggest that all new features (where actions/reducers, etc. have to be set up from scratch) are implemented using redux-toolkit.

## Acceptance criteria
> Describe criteria here (i.e. What is allowed/not allowed (negative tesing), validations, error messages and warnings etc.)
- [ ] Redux is set up using redux-toolkit
- [ ] There is no functional effect/change to the apps

## Specification tasks

- [ ] Development tasks are defined
- [ ] Test design / decide test need

## Development tasks
- [ ] Upgrade redux setup in studio frontend
  - [ ] app-development
  - [ ] dashboard
  - [ ] ux-editor
- [ ] Upgrade redux setup in app frontend
- [ ] Update developer handbook documentation

## Definition of done
Verify that this issue meets [DoD](https://confluence.brreg.no/display/T3KP/Definition+of+Done#DefinitionofDone-DoD%E2%80%93utvikling) (Only for project members) before closing.

- [ ] Documentation is updated (if relevant)
  - [ ] Technical documentation (docs.altinn.studio)
  - [ ] User documentation (altinn.github.io/docs)
- [ ] QA
- [ ] Manual test is complete (if relevant)
- [ ] Automated test is implemented (if relevant)
- [ ] All tasks in this userstory are closed (i.e. remaining tasks are moved to other user stories or marked obsolete)

olemartinorg:
Moving this back to Studio - app-frontend has solved this using our "saga slices".

PR #336, opened 2022-07-13
Wrapping createSlice() to include saga references

olemartinorg:
## Description
For my summer sabbatical, I've made an attempt at solving (or at least helping) a problem I encountered often when I started a few months ago. As I was not familiar with neither redux or sagas, it got very confusing following code around when an action got defined in one file, dispatched in another, with its type a third place, with a pointer here and some calls there. In the wealth of usages for actions, there were also some `yield takeLatest(...)` which didn't look like much to me, but were actually the listeners for each action, calling a separate function to 'do all the work'.

The goal with this change is to gather more of this logic (effects caused by dispatching an action) to one place. At least as much as possible. I created a wrapper, `createSagaSlice()`, which behaves like the redux-toolkit `createSlice()`, but it can be used to assign saga functions to an action as well as just reducers. The implementation of `createSagaSlice()` itself is fairly straight-forward, and the challenge of creating the wrapper was in getting the TypeScript magic to work such that generated types for the global reducers list (and `IRuntimeState`) could continue being inferred from the slice declarations.

When defining a SagaSlice, you have to specify the type of the `mkAction` helper function as ex. `MkActionType<IFormDataState>`, pulling in the explicit type of your state object. Along with another generic argument when calling `mkAction<ISubmitDataAction>({...})`, both reducers and saga functions will automatically get the correct payload argument when called, and dispatching your action will require an argument with a payload of the same type.

As an example of how this (hopefully) makes code easier to follow, let's take a look in `GenericComponent.tsx` when dispatching `FormDataActions.update({...})`. If you ctrl+click on `update` it leads you to the definition in `createSagaSlice()`, where you can immiediately tell that dispatching this action will run a reducer, setting some flags on the formData state. Further, a saga function is defined that takes from an actionChannel and calls the `updateFormDataSaga` function. Following that function definition shows you what it does, which actions it might in turn dispatch, and among them it might put the `FormDataActions.updateFulfilled({...})` action. Following that definition, we can see what the reducer does, and that dispatching that action will in turn call multiple other saga functions.

One downside I see by this approach is that not all sagas can be directly tied to one action. Simple sagas that `takeLatest` or `takeEvery` have their own shortcuts in `createSagaSlice()` and are easy to follow, but some more complex watcher sagas often have intricate conditions to be met before they run the 'worker' function - and it's not directly caused by dispatching an action. In theory, even though you can define any number of saga functions to the `saga` key for a given action, these sagas don't need to be tied to the action in any way. In fact, `createSagaSlice()` does not particularly care which action definition defined a saga function - the all get bundled and registered to the root saga the same way regardless where they came from. I would still argue that bundling code together where it logically belongs works in most cases. Those few places where it doesn't work at all, I created a generic `extraSagas` key outside the action definitions. 

## Related Issue(s)
- #196 (not directly, but this change required rewriting to redux-toolkit slices)
- #329

## Verification
- [x] All tests run green

## Documentation


haakemon (approved):
I would still like us to explore if RTK can help out with some of these cases (f.ex RTK Query for the simplest cases, and RTK createListenerMiddleware for more complex ones) - but regardless I think this PR is a good step making things a bit easier to follow.

I also stumbled upon similar efforts https://github.com/taoyingsong/redux-toolkit-with-saga and https://github.com/saga-slice/saga-slice , but neither seems to be very actively maintained.

Great work on this 🚀

olemartinorg:
@haakemon Yup, it was a good way to get to know the sagas better. There are definitely some of them more complex than others, but I'm not entirely sure they really need to be that complex. Simplifying this is correct way forward, I think.

Nice find with the other variants of saga slices! I looked around a bit before I started making this, but never found those. Still, it wasn't really all that hard to write it (and the code is fairly short). By writing this ourselves, we can also extend it (or morph it) to whatever we need in the future (if, for example, createListenerMiddleware needs a way to put the listeners closer to the action definitions).

PR #341, opened 2022-07-14
Minor eslint cleanup and new rules

olemartinorg:
## Description
Adding eslint rules:
- `eslint-plugin-sonarjs` so that we get local warnings about things sonarcloud would complain about either way. Converted most of these to warnings, so that we are still allowed to commit code without these passing.
- Cleaned up eslint configuration so that plain `.js` files do not fail anymore, as `eslint` complained they didn't belong to a project (according to the `paths` spec in `tsconfig.json`)
- Cleaning up some files (either automatically or manually, to get rid of some sonarjs-warnings)
- Adding the `prefer-template` rule, so that string concatenation is automatically rewritten to template strings. At least on my keyboard, the backtick key is hard to reach, so I prefer just hitting `ctrl+s` to let `eslint` do that for me. If you, as me, find string concatenation easier to write, but template strings easier to read, this rule gives you the best of both worlds.

## Verification
- [ ] All tests run green

## Documentation


src/altinn-app-frontend/src/components/base/FileUpload/FileUploadWithTag/FileListComponent.tsx
```diff
@@ -302,12 +230,17 @@ export function FileList(props: FileListProps): JSX.Element {
...
+                                  style={
+                                    props.mobileView
+                                      ? { marginLeft: '10px' }
+                                      : null
+                                  }
                                 />
                               </div>
-                            )}
+                            ) : null}
                             {!attachment.uploaded && (
```
haakemon, line 241: This could be simplified even further - the previous condition is checking the opposite, so instead of rendering `null` here, you can render this condition instead.

```suggestion
                            ) : (
```

Would need to change the rest of the condition as well though on ca line 255

src/.eslintrc.json
```diff
@@ -55,6 +55,27 @@
...
-    "react-hooks/exhaustive-deps": ["error"]
-  }
+    "react-hooks/exhaustive-deps": ["error"],
+    "sonarjs/no-duplicate-string": ["off"],
+    "sonarjs/cognitive-complexity": ["warn", 15],
+    "sonarjs/no-collapsible-if": ["warn"],
+    "sonarjs/prefer-single-boolean-return": ["warn"],
+    "sonarjs/no-identical-functions": ["warn"],
+    "sonarjs/no-small-switch": ["warn"],
+    "sonarjs/no-nested-template-literals": ["warn"]
```
haakemon, line 65: My one worry is that we are starting to get a more and more complex eslint rule, and I just dont want to end up in the scenario we had before, where the code was littered with `eslint-ignore` statements all over the place. If we introduce these rules, we need to stick by them and fix the issues. 

That said, I am all for enabling this, since we anyway already have them (just hidden in Sonarcloud) 😅 I wasnt aware eslint extensions for these existed.

src/altinn-app-frontend/src/components/base/FileUpload/FileUploadWithTag/FileListComponent.tsx
haakemon: Yeah it was not that easy to read the diff for this 😅 I think this component needs some love and simplification. Maybe Ill just do generic code cleanup for my sabbaitcal 🤔🤓

olemartinorg: Well, turns out this function has a cognitive complexity of 22 - so it's par for the course! :laughing:

src/.eslintrc.json
haakemon: Agree on disabling the `sonarjs/no-duplicate-string`. Could be valuable in some cases I guess, but would surely be some noise we cant really fix (and then we're back to adding eslint-disable statements which I really dont want).

PR #342, opened 2022-07-14
chore: remove invalid eslint-disable statements

haakemon:
## Description
- ~I propose to allow unused arguments named `_action`. This should of course be used only when needed, as the example in a saga where its used for TS infering types (see commit [`00ba52d` (#342)](https://github.com/Altinn/app-frontend-react/pull/342/commits/00ba52dfadee0c52931d48710e8e2a1d243d5a62) for details). By allowing this, we dont need to add `eslint-disable-next-line` whenever doing this, explaining why its needed.~ Not going to include this after all, as it will not be needed (ref: https://github.com/Altinn/app-frontend-react/pull/342#commitcomment-78491931). We can always add something like this back if we see the need for it.
- Removal of all `eslint-disable` statements, except one, [where a console.log is used](https://github.com/Altinn/app-frontend-react/blob/v3.46.0/src/altinn-app-frontend/src/features/instantiate/containers/PartySelection.tsx#L196-L197). Perhaps we can remove this (and the console.log) as well? Not really sure what value this statement gives.
- `filterOutInvalidData` function was rewritten a bit as a result of this. Tests for this did not exist, but are now added and was tested with old code before rewrite.

## Verification
- [x] All tests run green

### DIFF
--- a/src/.eslintrc.json
+++ b/src/.eslintrc.json
@@ -52,10 +52,7 @@
     "import/no-unresolved": ["off"],
     "@typescript-eslint/no-unused-vars": [
       "error",
-      {
-        "argsIgnorePattern": "^_action$",
-        "ignoreRestSiblings": true
-      }
+      { "ignoreRestSiblings": true }
     ],
     "@typescript-eslint/consistent-type-imports": ["warn"],
     "react-hooks/exhaustive-deps": ["error"]
--- a/src/.eslintrc.json
+++ b/src/.eslintrc.json
@@ -52,7 +52,10 @@
     "import/no-unresolved": ["off"],
     "@typescript-eslint/no-unused-vars": [
       "error",
-      { "ignoreRestSiblings": true }
+      {
+        "argsIgnorePattern": "^_action$",
+        "ignoreRestSiblings": true
+      }
     ],
     "@typescript-eslint/consistent-type-imports": ["warn"],
     "react-hooks/exhaustive-deps": ["error"]
--- a/src/altinn-app-frontend/src/features/form/data/formDataSlice.ts
+++ b/src/altinn-app-frontend/src/features/form/data/formDataSlice.ts
@@ -66,9 +66,6 @@ const formDataSlice = createSlice({
       state.isSaving = false;
       state.ignoreWarnings = true;
     },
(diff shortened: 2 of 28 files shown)


src/.eslintrc.json
```diff
@@ -52,7 +52,10 @@
     "import/no-unresolved": ["off"],
     "@typescript-eslint/no-unused-vars": [
       "error",
-      { "ignoreRestSiblings": true }
+      {
+        "argsIgnorePattern": "^_action$",
```
olemartinorg, line 56: Nice! Didn't know about that option. I've heard it's common to allow unused arguments when prefixed with an underscore, and in fact WebStorm (with the built-in inspection that looks for unused arguments) did silence its warning when I prefixed with an underscore - but I added the ignore comment just for the remaining eslint warning.

(Btw, I wrote this comment directly in the linked commit, but seems like it got posted to the void. I'll try again here!)

src/altinn-app-frontend/src/features/form/data/formDataSlice.ts
```diff
@@ -66,9 +66,6 @@ const formDataSlice = createSlice({
       state.isSaving = false;
       state.ignoreWarnings = true;
     },
-    // The _action parameter is unused, but these parameters are used by TypeScript to infer the payload type for
-    // this action, so we need to keep it here even if we only read it in the saga.
-    // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
olemartinorg, line 71: This was actually meant to be short-lived, as it gets removed again in #336. In `createSagaSlice`, the API allows for entirely empty/non-reducer actions (without having to declare them elsewhere with `createAction()`), and it's also allowed for a reducer to ignore the payload even if the action is declared to expect a payload.

haakemon: I reverted this change already, based on your previous comments 🤓

olemartinorg (approved):
Superb! :100:

src/altinn-app-frontend/src/features/form/data/formDataSlice.ts
olemartinorg: Oh no! My github incompetence is showing! :laughing:

PR #343, opened 2022-07-17
Discriminated union in layout component types

olemartinorg:
## Description
This change mostly affects TypeScript definitions, and should be the last code contribution from me during this sabbatical period.

I had previously made some type guard functions (like `isGroupComponent(...)` and `isFileUploadComponent(...)`) that could be used to check if a given `ILayoutComponent` is a specific component, so that you can access the additional properties on it (for a `Group` this might mean the `.children` property, and for a `FileUpload` component that might mean the `.maxNumberOfAttachments` property). I never considered these type guard functions to be the optimal solution, so with this PR I've changed the definition of `ILayoutComponent` to be much smarter - it's now a union type of all possible components. Effectively it performs just like a [discriminated union](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#discriminated-unions).

An important part of this change was to make sure `fetchFormLayoutSagas` cleans up component `.type` properties, so they are always in the correct case (`group` is converted to `Group`),

These changes means we can write code like this, without any type guard functions:

```typescript
const component: ILayoutComponent | ILayoutGroup = {} as any;

// Gives a TypeScript error. We do not know if this component is a Group yet.
alert(component.children);

if (component.type === 'Group') {
  // At this point the component type has been narrowed to ILayoutGroup, so .children is accessible
  component.children.forEach((child) => child);
} else if (component.type === 'FileUpload' && component.hasCustomFileEndings) {
  // Component has now been narrowed to a FileUpload component.
  // We can access properties unique to the FileUpload component,
  // and we got auto-completion when comparing the .type property, since
  // the editor knows all the possible component types.
  alert(component.validFileEndings);
}

// We can specify that we only allow certain components, not all of them:
function onlyTakesInputComponents(component: ILayoutComponent<'Input'>) {
  alert('Not implemented');
}

// This gives a TypeScript error. Not all components are assignable to an Input component.
onlyTakesInputComponents(component);

if (component.type === 'Input') {
  // This works, we know the component type now
  onlyTakesInputComponents(component);
}
```

In addition to this, the mapping from component types to React components in `src/components/index.ts` has been greatly simplified (flattened into a simple dictionary), with typings enforcing you to define the custom component properties that can be defined in the layout as soon as you implement a new component. It will also give you TypeScript errors if you define a new component in `layout/index.ts` without specifying a React component for `GenericComponent.tsx` in `src/components/index.ts`. 

## Verification
- [x] All tests run green

## Documentation


src/altinn-app-frontend/src/components/base/RadioButtons/RadioButtonsContainerComponent.tsx
```diff
@@ -1,17 +1,13 @@
...
-import type { IMapping, IOption, IOptionSource } from 'src/types';
 import { useRadioButtons } from 'src/components/base/RadioButtons/radioButtonsUtils';
 import { ControlledRadioGroup } from 'src/components/base/RadioButtons/ControlledRadioGroup';
+import type { ILayoutCompRadioButtons } from 'src/features/form/layout';
 
-export interface IRadioButtonsContainerProps extends IComponentProps {
+export interface IRadioButtonsContainerProps
+  extends IComponentProps,
+    ILayoutCompRadioButtons {
   validationMessages?: any;
```
haakemon, line 10: Why is `validationMessages` not also moved to `ILayoutCompRadioButtons` ?

src/altinn-app-frontend/src/features/form/layout/fetch/fetchFormLayoutSagas.test.ts
```diff
@@ -4,13 +4,37 @@ import {
...
 import { FormLayoutActions } from '../formLayoutSlice';
+import type {
+  ILayoutGroup,
+  ILayoutCompSummary,
+  ILayoutCompFileUploadWithTag,
+} from 'src/features/form/layout';
 
 describe('fetchFormLayoutSagas', () => {
+  describe('cleanLayout', () => {
+    it('should convert lower-case types to the correct case', () => {
```
haakemon, line 21:
```suggestion
    it('should convert incorrectly cased types to the correct case', () => {
```

src/altinn-app-frontend/src/components/presentation/NavigationButtons.test.tsx
```diff
@@ -93,6 +93,7 @@ describe('NavigationButton', () => {
     render(
       <Provider store={mockStore}>
         <NavigationButtons
+          type={'NavigationButtons'}
```
haakemon, line 96: Seems a bit redundant to have to specify the type on the component in tests - any easy way to avoid having to do this?

src/altinn-app-frontend/src/components/summary/SummaryComponent.test.tsx
```diff
@@ -28,7 +28,7 @@ describe('SummaryComponent', () => {
             (t) =>
               ({
                 id: t,
-                type: t,
+                type: t === 'group' ? 'Group' : t,
```
haakemon, line 31: I dont think really gives any value - the test now just fixes the incorrectly cased implementation, and this would not affect any production code. 

I would suggest to just remove the `group` entry on line 22, and leave `t` here instead.

haakemon (approved):
This is some really good stuff 🤓 Left some minor comments to consider

src/altinn-app-frontend/src/components/base/RadioButtons/RadioButtonsContainerComponent.tsx
olemartinorg: I'm not exactly sure where that value is supposed to come from, but it's neither `IComponentProps` (properties from `GenericComponent` when rendered from there) nor `ILayoutCompRadioButtons` (properties defined for that component in the layout file). There were a few cases like this. The component apparently expects to sometimes get a value with that property name, but I'm not sure exactly where it is supposed to come from - other than it surely is some internal state that might not belong to either of the two interfaces it inherits from.

src/altinn-app-frontend/src/components/presentation/NavigationButtons.test.tsx
olemartinorg: Yeah, you're right. I added `Omit<..., 'type'>` to these prop types instead.

src/altinn-app-frontend/src/components/summary/AttachmentWithTagSummaryComponent.test.tsx
```diff
@@ -30,7 +29,7 @@ describe('AttachmentWithTagSummaryComponent', () => {
   const extendedState = {
     attachments: {
       attachments: {
-        [typeName]: [
```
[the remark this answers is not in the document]
olemartinorg: Yep, I went a bit overboard when getting this to work. Reverting back to using a const here. Since it is a const, the inferred type is `"FileUploadWithTag"`, not `string`, so it works perfectly fine that way as well.

src/altinn-app-frontend/src/components/summary/SummaryComponent.test.tsx
olemartinorg: To be fair, it introduces a new case in the test where `id` and `type` actually differs... :wink: Kidding aside, I agree - I'll remove the useless test.

Issue #100, opened 2022-03-17
Validation: Move error message box to the bottom 


mrosvik:
## Description
_As a user I want to see why I can´t click "Next", so that I can complete the form._ 

- The error message box thats shows when you try to hit "Next"/"Send" should be connected with the "Next"/"Send"-button, since the summary explain why the next button is not working. 
- The error messages should link to the field where the error is (set focus).
- WCAG: The current error message use H2 and H4 as texts. This is not compliant with WCAG. Remember to think about what header we use on the new component

## Design
https://www.figma.com/file/Pvti3aRcOwH0k5Z7DrLznk/Arbeidsomr%C3%A5de%3A-Altinn-Studio-APPS?node-id=2545%3A50628

<img width="1847" alt="image" src="https://user-images.githubusercontent.com/1464915/158781837-f24304be-4f57-4bf4-9251-ca1f853a612e.png">

## Development tasks
- [ ] Move error box from the top, into the bottom of the form. 
- [ ] Link the error to the connected field (set focus).

## Definition of done

- [ ] Documentation (docs.altinn.studio) is updated (if relevant)
  - [ ] Technical documentation
  - [ ] User documentation
- [ ] QA
- [ ] Manual test is complete (if relevant)
- [ ] Automated test is implemented (if relevant)
- [ ] All tasks in this userstory are closed (i.e. remaining tasks are moved to other user stories or marked obsolete)

olemartinorg:
Innspill fra @FinnurO: Mulig vi bør endre teksten fra "Du må *fikse* disse feilene før du kan gå videre" til "Du må *rette* disse feilene før du kan gå videre".

Febakke:
@olemartinorg Is this the correct place for test feedback

**Screen reader test:** Ok (VoiceOver on chrome)
**Keyboard navigation:** Ok

**Smaller screens:** 
On smaller screens the text fall down and the icon remains. It would be better to have the text linebreak and always have the start of the text aligned with the icon.
![Skjermbilde 2022-08-02 kl  10 59 26](https://user-images.githubusercontent.com/33222679/182355568-8b325c4f-8ae6-4f0b-82d1-913d238feb47.png)

**Contrast**
In the Figma sketches it looks the pink color is set to 50% opacity. With a white background this would equal to #FCE4E9. This is probably easy to fix with tokens at a later stage?

[...]

PR #346, opened 2022-07-20
Moving ErrorReport component to the bottom of the form

olemartinorg:
## Description
This change moved the `ErrorReport` component away from over the form modal to inside the modal, at the bottom of the form. Noteworthy changes:

- This location is (usually) where navigation buttons are located, and these `NavigationButtons` will now be incorporated into the `ErrorReport` should they be located at the end of the currently active layout when there are any validation errors.
- The ErrorReport now uses a Panel (currently the `warning` variant, as the `error` variant is not implemented yet)
- When clicking the `next` navigation button, if validation messages are added to previous components in the form, you could end up seemingly being force-scrolled up a bit in the form (because added validation messages push the navigation buttons down). Now that validation messages are also displayed around these buttons, clicking the navigation buttons will maintain the position of the button(s) in the viewport when the `ErrorReport` shows up.
- Errors listed in the `ErrorReport` are clickable (and tab-able). Clicking them sends you to the component with the validation error (if any).
- The new functionality that navigates you to the correct page and scrolls the correct component into view (and focusing input/textarea fields) also works when clicking 'change' from a summary page.  

## Related Issue(s)
- #100
- Altinn/altinn-design-system/pull/69

## Verification
- [x] All tests run green

## Documentation


src/altinn-app-frontend/src/components/GenericComponent.tsx
```diff
@@ -154,6 +155,20 @@ export function GenericComponent<Type extends ComponentExceptGroup>(
     );
   }, [componentValidations]);
 
+  React.useLayoutEffect(() => {
+    if (!hidden && shouldFocus && gridRef.current) {
+      gridRef.current.scrollIntoView();
```
haakemon, line 160: Should we perhaps use smooth scrolling? Personally I find it annoying when the page jumps somewhere. Smooth scrolling helps visualize what happens.

There are probably some rules around this though, and I'm not really sure what they are 😅 

@Febakke Do we have any guidelines/rules about automatically scrolling the page?

src/altinn-app-frontend/src/components/GenericComponent.tsx
```diff
@@ -154,6 +155,20 @@ export function GenericComponent<Type extends ComponentExceptGroup>(
...
+
+      const maybeInput = gridRef.current.querySelector('input,textarea') as
```
haakemon, line 162: We probably should cater for other focusable items as well? Selects, radio, checkbox, etc...

src/altinn-app-frontend/src/components/message/ErrorReport.tsx
```diff
@@ -1,128 +1,145 @@
...
-  const validations = useAppSelector(
-    (state) => state.formValidations.validations,
+export interface IErrorReportProps {
+  components: ILayout;
+}
+
+const ArrowForwardIcon =
+  '<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" width="16" style="position: relative; top: 2px">' +
+  '<path d="M12 4l-1.41 1.41L16.17 11H4v2h12.17l-5.58 5.59L12 20l8-8z"></path>' +
+  '</svg>';
```
haakemon, line 23: Would suggest using a template literal here

```suggestion
const ArrowForwardIcon = `<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" width="16" style="position: relative; top: 2px">
  <path d="M12 4l-1.41 1.41L16.17 11H4v2h12.17l-5.58 5.59L12 20l8-8z"></path>
</svg>`;
```

That should give a bit nicer syntax highlighting (at least in some editors) and you can easily write with the correct indentation. Though its not really important 😅

src/altinn-app-frontend/src/components/message/ErrorReport.tsx
```diff
@@ -1,128 +1,145 @@
...
-    }
-  }, [hasSubmitted, unmappedErrors]);
+  const hasErrors = errorsUnmapped.length > 0 || errorsMapped.length > 0;
 
-  if (!formHasErrors) {
+  if (!hasErrors) {
     return null;
   }
 
+  const OnClickError =
```
haakemon, line 63: This is handling the event, and thus should be named `handleErrorClick`. It is common to put the "type" (Click) last as well. That way when listing all events, they would be grouped by "name" first, then "type". Hope that makes sense 🙂

```suggestion
  const handleErrorClick =
```

src/altinn-app-frontend/src/components/message/ErrorReport.tsx
```diff
@@ -1,128 +1,145 @@
...
+              <ul className={classes.errorList}>
+                {errorsUnmapped.map((error: React.ReactNode, index: number) => {
+                  return <li key={`unmapped-${index}`}>{error}</li>;
+                })}
+                {errorsMapped.map((error, index) => {
+                  return (
+                    <li key={`mapped-${index}`}>
+                      <button
+                        className={classes.buttonAsInvisibleLink}
+                        tabIndex={0}
```
haakemon, line 124: Buttons are tabbable by default, so I dont think this is needed?

```suggestion
```

src/altinn-app-frontend/src/components/message/ErrorReport.tsx
```diff
@@ -1,128 +1,145 @@
...
-        </div>
-      </div>
-    </div>
+              <ul className={classes.errorList}>
+                {errorsUnmapped.map((error: React.ReactNode, index: number) => {
+                  return <li key={`unmapped-${index}`}>{error}</li>;
+                })}
+                {errorsMapped.map((error, index) => {
+                  return (
+                    <li key={`mapped-${index}`}>
```
haakemon, line 121: Using indexes in `map` is [generally not a good idea](https://robinpokorny.medium.com/index-as-a-key-is-an-anti-pattern-e0349aece318), and even if it may be fine here, it leaves something that can be copied later to a place it could be problematic.

Could we use `componentId` instead? I guess that will be unique?

```suggestion
                {errorsUnmapped.map((error: React.ReactNode) => {
                  return <li key={`unmapped-${error.componentId}`}>{error}</li>;
                })}
                {errorsMapped.map((error) => {
                  return (
                    <li key={`mapped-${error.componentId}`}>
```

src/altinn-app-frontend/src/features/form/components/FullWidthWrapper.tsx
```diff
@@ -19,14 +20,25 @@ const useStyles = makeStyles({
...
+    '@media (min-width: 768px)': {
+      marginBottom: '-36px',
+    },
+  },
 });
 
-export function FullWidthWrapper({ children }: IFullWidthWrapperProps) {
+export function FullWidthWrapper({
+  children,
+  onBottom,
```
haakemon, line 33: This looks like an event - can we name it something else? 🙂 `isOnBottom` f.ex (or would it be `isAtBottom`? 🤔)

Should probably also have `false` as default value

src/altinn-app-frontend/src/features/form/components/FullWidthWrapper.tsx
```diff
@@ -19,14 +20,25 @@ const useStyles = makeStyles({
...
+}: IFullWidthWrapperProps) {
   const classes = useStyles();
 
   return (
     <div
-      className={classes.fullWidthWrapper}
+      className={`${classes.fullWidthWrapper} ${
+        onBottom ? classes.consumeBottomPadding : ''
+      }`}
```
haakemon, line 41: Would suggest to use `classnames` helper for this conditional class

```jsx
import cn from 'classnames';
```

```suggestion
      className={cn(classes.fullWidthWrapper, {
        [classes.consumeBottomPadding]: isOnBottom
      }}
```

haakemon (approved):
This looks great! 🥳 Left some minor comments to consider.

src/altinn-app-frontend/src/features/form/components/Panel.test.tsx
```diff
@@ -66,7 +72,7 @@ const render = (
   const allProps = {
     title: 'Panel Title',
     children: 'Panel Content',
-    variant: 'info',
+    variant: 'info' as const,
```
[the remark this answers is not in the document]
olemartinorg: Starting at line 72; since allProps does not have a specified type, it is inferred. Since it is a constant (`const allProps`), the property types will be inferred as well. However, it only goes so far - so `variant` is inferred as type `string`. When I tightened the variant property type to be either `'info' | 'warning' | 'success' | 'error'`, giving it a `string` is too broad. By adding 'as const', TypeScript will infer the type of `allProps` to be `{..., variant: 'info'}` not simply `{..., variant: string}`.

In other words, it fixes a TypeScript error.. :-)

src/altinn-app-frontend/src/features/form/containers/Form.test.tsx
```diff
@@ -152,6 +152,75 @@ describe('Form', () => {
...
+    const recurseUp = (el: HTMLElement) => {
+      if (!el.parentElement || el.parentElement === el) {
+        return el;
+      }
+      if (el.parentElement.dataset['testid'] === 'ErrorReport') {
+        return el.parentElement;
+      }
+      return recurseUp(el.parentElement);
+    };
+    expect(recurseUp(navButtons)).toEqual(errorReport);
```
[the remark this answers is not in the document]
olemartinorg: Hmm, This code isn't just testing that `NavigationButtons` exist, but that it is a child of the `ErrorReport` component. If you look at `extractBottomButtons` in `Form.tsx`, it will remove the `NavigationButtons` button from the layout and render it as a child of `ErrorReport` when there are errors in the form.

src/altinn-app-frontend/src/components/GenericComponent.tsx
olemartinorg: Doesn't really look like it in this piece of code, but this happens as a direct consequence of user action. I would probably have used hash anchors instead, if we didn't already use `window.location.hash` for routing.. :wink: Smooth scrolling is nice (when the animation isn't taking too long, at least), but I think users should be fairly used to the 'clicking this button sends you to the correct part of the document' pattern.

src/altinn-app-frontend/src/components/GenericComponent.tsx
olemartinorg: Yup - I never tested any other items. I'd prefer if there were ways to force these components themselves to implement focusing. For example, while the `DatePicker` contains an `input`, it might be best to put focus on the button to open the calendar instead. For `select`, I think it's important the dropdown doesn't actually open when setting focus, and radio/checkbox should default to focus the chosen value (if any). I'll look into adding more of the easy ones, at least.

src/altinn-app-frontend/src/components/message/ErrorReport.tsx
olemartinorg: Good point! I don't think we can avoid it in the first one (unmapped errors), as they (by being unmapped) don't have `componentId`. This also raises a point that I'll have to check to make sure this works with components inside repeating groups (including clicking the error to scroll the component into view).

src/altinn-app-frontend/src/features/form/components/FullWidthWrapper.tsx
olemartinorg: You're right! Renaming it and setting a default value.

src/altinn-app-frontend/src/components/message/ErrorReport.tsx
olemartinorg: Tested this out, cleaned up the code and fixed an issue I found where `FormLayoutActions.updateFocus` didn't work with component IDs in repeating groups. Also reported issue #350, which I found when cleaning this up.

src/altinn-app-frontend/src/components/GenericComponent.tsx
olemartinorg: Done! `select` does not open the dropdown by default. Radio and checkbox already worked (they are `input`), but there is no visual indication these are focused.

src/altinn-app-frontend/src/features/form/containers/Form.tsx
```diff
@@ -110,45 +110,33 @@ export function Form() {
...
-      const componentsToRender = layout.filter(
-        (component) => !renderedInGroup.includes(component.id),
-      );
-      setFilteredLayout(componentsToRender);
+  const [mainComponents, errorReportComponents] = React.useMemo(() => {
+    if (layout && hasErrors) {
+      const topLevel = topLevelComponents(layout);
+      return extractBottomButtons(topLevel);
+    } else if (layout) {
+      return [topLevelComponents(layout), []];
```
mjulstein, line 133:
```suggestion
if(!layout){
  return [[], []];
}
const topLevel = topLevelComponents(layout);
if (hasErrors) {
  return extractBottomButtons(topLevel);
}
return [topLevel, []];
```

src/altinn-app-frontend/src/features/form/containers/Form.tsx
```diff
@@ -161,14 +149,44 @@ export function Form() {
...
+  for (const component of [...layout].reverse()) {
+    if (extract.has(component.type) && toMainLayout.length === 0) {
+      toErrorReport.push(component);
+    } else {
+      toMainLayout.push(component);
+    }
+  }
+
+  return [toMainLayout.reverse(), toErrorReport.reverse()];
+}
```
mjulstein, line 192: could these be extracted to a utils.ts (or similar) and have unit-tests that actually describe what they are for?

src/altinn-app-frontend/src/features/form/containers/Form.tsx
```diff
@@ -161,14 +149,44 @@ export function Form() {
...
+  const toErrorReport: ILayout = [];
+  for (const component of [...layout].reverse()) {
+    if (extract.has(component.type) && toMainLayout.length === 0) {
+      toErrorReport.push(component);
+    } else {
+      toMainLayout.push(component);
+    }
+  }
+
+  return [toMainLayout.reverse(), toErrorReport.reverse()];
```
mjulstein, line 191: Not really sure why these are reversed?

src/altinn-app-frontend/src/features/form/containers/Form.tsx
```diff
@@ -161,14 +149,44 @@ export function Form() {
...
+function extractBottomButtons(layout: ILayout) {
+  const extract = new Set<ComponentTypes>([
+    'NavigationButtons',
+    'Button',
+    'PrintButton',
+  ]);
+
+  const toMainLayout: ILayout = [];
+  const toErrorReport: ILayout = [];
+  for (const component of [...layout].reverse()) {
```
mjulstein, line 183: Was the intent to break before scanning the whole layout?

src/altinn-app-frontend/src/features/form/containers/Form.tsx
olemartinorg: Good point! Will do that.

src/altinn-app-frontend/src/features/form/containers/Form.tsx
olemartinorg: Because we only want to extract the buttons etc from the bottom/end of the layout. In the for-loop above the layout is copied and reversed when we loop it so that we can easily extract the bottom-most components, but skip a `Button` located near the top of the layout. Uhh, not sure if that was more understandable, but a unit-test would surely help here - I'll get on it!

src/altinn-app-frontend/src/features/form/containers/Form.tsx
olemartinorg: If I understood you correctly; yes. Earlier code iterated the layout from the bottom-up to extract the bottom-most of these components (`NavigationButtons`, `Button` and `PrintButton`) and break the iteration when it finds other components. This is basically the same, but gives you two outputs (one for the regular layout, another with these components split into its own layout so they can be injected into `ErrorReport` as children).

src/altinn-app-frontend/src/components/GenericComponent.tsx
haakemon: Ah, makes sense. Nevermind 🙂

src/altinn-app-frontend/src/features/form/containers/Form.test.tsx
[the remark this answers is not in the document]
haakemon: Ah, I got fooled from the comment and didnt read thoroughly enough - comments are evil 😂

Anyway, if you want to ensure the buttons are rendered inside something else, you could use [`within`](https://testing-library.com/docs/dom-testing-library/api-within/) for this instead of the custom code. IMO that would be preferred, as it would be a bit easier to understand what is going on, and its an established pattern when using RTL 🙂

Something like this should work I think: 
```jsx
expect(within(errorReport).getByTestId('NavigationButtons')).toBeInTheDocument();
```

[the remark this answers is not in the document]
olemartinorg: Nice! I believe I googled for this quickly, but didn't find anything. I'll rewrite the test to use this instead. Comments aren't evil, but they can't replace reading the code.. :wink:

src/altinn-app-frontend/src/features/form/components/Panel.test.tsx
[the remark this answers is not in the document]
haakemon: Nice, learned something new today. Thanks for the explanation.

PR #347, opened 2022-07-21
Smart un-uppercasing for required field validations

olemartinorg:
## Description
After the introduction of better error messages for required fields (`You have to fill out {0}`), I've been slightly irked by these error messages falling back to the `title` text binding (if no `shortName` exists), but keeping the uppercase first letter these titles usually have. This produces error messages like `You have to fill out First name` instead of the more natural `You have to fill out first name`.

It might not be advisable to simply un-uppercase the first letter all the time, as the title text might be an all-uppercase abbreviation, etc. This impulsive PR nobody asked for tries to solve this by looking at the first 5 non-symbol letters in the string in order to decide if it should uncapitalize the first letter.

Demo of what this does (here inside the unfinished #346, with warning colors):

ucfirst

Note: This was an early draft of this PR, and the behaviour has slightly changed after recording. The error `Du må fylle ut dato for navneendring` now actually reads `Du må fylle ut Dato for navneendring` (still capitalized), because this field has a `shortName` key defined (although it is still capitalized on the first letter for some reason).

Very open to discussion on this stunt! I'm not sure if we want to touch these strings, but I was hoping it is slightly better in most cases than just keeping the `title` as-is.

## Related Issue(s)
- #230 
- #130

## Verification
- [x] All tests run green

## Documentation

### DIFF
--- a/src/altinn-app-frontend/src/utils/formComponentUtils.ts
+++ b/src/altinn-app-frontend/src/utils/formComponentUtils.ts
@@ -521,7 +521,7 @@ export function getFieldName(
   fieldKey?: string,
 ): string {
   if (fieldKey) {
-    return smart_unucfirst(
+    return smartLowerCaseFirst(
       getTextFromAppOrDefault(
         `form_filler.${fieldKey}`,
         textResources,
@@ -537,7 +537,7 @@ export function getFieldName(
   }
 
   if (textResourceBindings.title) {
-    return smart_unucfirst(
+    return smartLowerCaseFirst(
       getTextResourceByKey(textResourceBindings.title, textResources),
     );
   }
@@ -548,7 +548,7 @@ export function getFieldName(
 /**
  * Un-uppercase the first letter of a string
  */
-export function unucfirst(text: string, firstLetterIndex = 0): string {
+export function lowerCaseFirst(text: string, firstLetterIndex = 0): string {
   if (firstLetterIndex > 0) {
     return (
       text.substring(0, firstLetterIndex) +
@@ -563,7 +563,7 @@ export function unucfirst(text: string, firstLetterIndex = 0): string {
  * Un-uppercase the first letter of a string, but be smart about it (avoiding it when the string is an
  * uppercase abbreviation, etc).
  */
-export function smart_unucfirst(text: string): string {
+export function smartLowerCaseFirst(text: string): string {
   const uc = text.toUpperCase();
   const lc = text.toLowerCase();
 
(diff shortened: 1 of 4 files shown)


src/altinn-app-frontend/src/utils/formComponentUtils.ts
```diff
@@ -535,8 +537,64 @@ export function getFieldName(
...
+    );
+  }
+  return text[firstLetterIndex].toLowerCase() + text.substring(1);
+}
+
+/**
+ * Un-uppercase the first letter of a string, but be smart about it (avoiding it when the string is an
+ * uppercase abbreviation, etc).
+ */
+export function smart_unucfirst(text: string): string {
```
haakemon, line 566: Should perhaps use `smartUnucFirst` as name instead to follow the same naming pattern we have for other functions? Personally I would probably prefre something like `smartLowerCaseFirst` 🙂

test/cypress/e2e/fixtures/texts.json
```diff
@@ -15,7 +15,7 @@
   "missingRights": "Du mangler rettigheter for å se denne tjenesten",
   "next": "Neste",
   "requiredField": "Feltet er påkrevd",
-  "requiredFieldLastName": "Du må fylle ut Nytt etternavn",
+  "requiredFieldLastName": "Du må fylle ut nytt etternavn",
   "requiredFieldDateFrom": "Du må fylle ut Dato for navneendring",
```
haakemon, line 19: Shouldnt Dato have lowercase `D` ? We should perhaps fix the shortname (as you mentioned in the description) as well? Or maybe I have misunderstood 😅

haakemon (approved):
I like it! Seems like a good approach to me 🚀

src/altinn-app-frontend/src/utils/formComponentUtils.ts
olemartinorg: You're absolutely right! My [years of programming in PHP are shining through](https://news-web.php.net/php.internals/70691).. :laughing:

test/cypress/e2e/fixtures/texts.json
olemartinorg: Yeah, I agree - but that's defined in the `frontend-test` app, and I don't really see the need to fix that just to update this test (it would break other tests for a while until this is merged as well). Since the `shortName` was introduced because of this change to new error messages, I'm assuming app developers will make the smarter decision to use lowercase names most of the time (since it looks nicer).

haakemon: Ah, I didnt read the full file path here, and didnt realise this was just testdata. Not too worried about it then, fine to leave as is 🙂

Issue #344, opened 2022-07-18
Switch from hash-router to browser-router


mjulstein:
### Description
Currently we are using HashRouter as our React-Router option in Apps (and other solutions.)
The drawback of this is that we loose the ability to comply with modern accessibility requirements such as landmarks.
HashRouter is no longer recommended by the React-Router either.

## Upgrade to React-Router v6
React Router 6 has changes that would require re-write of the changes required to switch to BrowserRouter.
- [x] Therefore it makse sense to upgrade to v6 first:  #353

## Related issues:
- [x] #353 
- [ ] #446
- [ ] #373 
- [x] #225 
- [ ] Altinn/app-template-dotnet#121
- [ ] #374 

## Requirements:
The applications no longer use HashRouter and the following App-Configurations work as expected (before):
- [ ] State-full apps
- [ ] State-less apps
- [ ] Anonymous State-less apps

## Additional tasks:
- [ ] Edit the LandmarkShortcut component to make use of the router switch.
- [x] Update to latest version of react-router

mjulstein:
Moved this issue out of current sprints as it's not prioritized for now. Some of the subtasks need to be done though.

PR #349, opened 2022-07-21
refactor: remove all usages of withStyles HOC in favor of hooks approach

### LINKED ISSUE #348: cleanup: Remove withStyles HOC
# Description

When creating components in the design system, it is often valuable to try and integrate them in app-frontend at the same time to verify its API. This is easily done by symlinking the altinn-design-system dist folder into node_modules of app-frontend. However, this causes some odd typescript errors:
image

And I am not really sure why these appear, but it seems to be caused by wrapping some components in the `withStyles` HOC from material-ui: 
image

~Rewriting this to use `makeStyles`/`useStyles` hook instead seems to resolve the problem.~ Turns out this was only part of the problem - there are several other typescript errors not related to this:
image

So not really sure whats going on here... Perhaps upgrading react-router to the latest version could help, but that should be done in a separate task (#344).

## Tasks:

- [x] Remove `withStyles` HOC usage

haakemon:
## Description
I initially thought updating this would make symlinking from altinn-design-system repo to app-frontend would work, so it would be faster to iterate on new design system components, but theres still some missing pieces. The other errors are related to react-router, so updating that may fix the remaining issues.

Even if this doesnt fix all of the issues in #348, it fixes most of them. React-router upgrade will be done in #344 

## Related Issue(s)
- fixes #348 

## Verification
- [x] All tests run green


olemartinorg (approved):
Superb! :100:

Issue #319, opened 2022-07-05
Repeating lists/groups: Entries don't disappear when removed programatically


allinox:
### Description of the bug

We are currently trying to populate a repeating list/group based on a choice the user has made previously in the form. The user can then add additional information to these choices previously made in the repeating group.
Populating the list works fine, and is done through DataProsessingHandler. However, removing them does not.

For some reason, the front-end still displays entries that are removed, even though the blob-storage has correctly removed them from the list. And in some cases, it does remove the data from the removed entires, but refills and displays the entry as a copy of one of the actual valid ones. A refresh of the page seems to fix the issue.

What's strange is the fact that the C#-model seems to be remembering the same (wrong) state of the list as the front-end does, even though the blob-storage contains the correct list and values.

### Steps To Reproduce

1. Use an app which populates and depopulates a repeating group based on some previous question in the app.
2. Choose multiple options so as to cause multiple entries to show up in the repeating group
3. See that these show up in the repeating group
4. Go back and de-select one or more of the previous choices
5. Observe that the deselected choices have not been removed, and/or still remain but with the wrong data.

### Example

Say I check all available options in which child I want to apply for here:
![image](https://user-images.githubusercontent.com/22012997/177298584-15b6dbda-3d8e-4238-bb8f-e2351c570a60.png)

This will (correctly) load prefilled values over to the repeating list so that I can add the additional required information to each child.
![image](https://user-images.githubusercontent.com/22012997/177298770-7e0f5596-d748-487b-a484-b839b14cc0cd.png)

But then, say I change my mind, and go back to uncheck the two first children.
![image](https://user-images.githubusercontent.com/22012997/177298943-258eb63b-f01c-46d6-9e51-bb785fa2d93a.png)

One would expect this to cause the data and entries of the two first children to be removed from the list and for only the third (Falsk Talltegn) to remain. And indeed, this is what is stored in the blob. But when I move on to the next page, it is not what I see as a user. Instead, I get an incorrect list, with two entries of the one child I actually checked off for, and one unchecked child.
![image](https://user-images.githubusercontent.com/22012997/177299292-a056f518-0aad-447d-8f13-f5a067e23f19.png)

olemartinorg:
[...]

PR #354, opened 2022-07-28
Redux fixes for #319

olemartinorg:
## Description
This fixes two minor bugs found when debugging and fixing #319, and also adds Cypress tests for this bug. Cypress tests will fail until the linked PR in `app-lib-dotnet` is merged, a new version of `Altinn.App.Common` is released, and the changes to `frontend-test` are merged from the `bug/frontend-319` branch into master (along with upgrading to the new `Altinn.App.Common` package and deploying to tt02). All of this has to happen in sequence before this PR will pass its Cypress tests.

The two minor bugs found here:
1. When attempting to open the app where this bug was first found, I had trouble prying it open to be able to instantiate it. At one point instantiation caused an error, the action got rejected and instantiating got set to false. This caused the `useEffect()` to create an infinite request loop, continually requesting a new instance. Making sure we don't loop infinitely when there are errors so we don't DoS ourselves.
2. The code in `handleCalculationUpdate` could end up dispatching redux actions too slowly, ending up with the `formData.unsavedChanges` state to remain as `true` after form data updates from server-side calculations. This might also have been caused by the saga in `FormDataActions.update` blocking, such that it could miss some events when multiple are dispatched at about the same time. The effect of this was that you can't navigate to the next page, since the form thinks it has unsaved changes. The page hangs until the next time the form is saved successfully, then it would navigate to the next page.

## Related Issue(s)
- #319
- Altinn/app-lib-dotnet/pull/8

## Verification
- [x] All tests run green

## Documentation


haakemon (approved):
LGTM 🙂

Issue #327, opened 2022-07-06
Change repeating group "lagre" button to the secondary button design


Febakke:
### Description

User tests show that user where confused by the icon button with the green checkmark. Thinking it was an indication that the page was saved. The secondary button has a more typical button design and should not compete with the primary (next) button on attention. 

## Sketches
![Skjermbilde 2022-07-06 kl  10 35 00](https://user-images.githubusercontent.com/33222679/177507343-5164107f-08b1-4071-908d-904bdbb5d6a4.png)

![Skjermbilde 2022-07-06 kl  10 35 13](https://user-images.githubusercontent.com/33222679/177507390-b6d216ef-3ab5-404e-8460-5f611a12d485.png)

SimenRokaas:
Fare for at noen vil tro at knappen lukker hele skjemaet... hva med bare "Lagre"?

Febakke:
**Tastaturnavigasjon: Feil fokusvisning**
Rammen rundt knappen forsvinner og det blir ikke vist vår standard fokusvisning.
![Skjermbilde 2022-08-09 kl. 12.54.27.png](https://images.zenhubusercontent.com/5c8b643d6e79732923030a9c/61c1ee54-7934-49d4-b841-a5285a49a493)

**Styling: Den er litt nærme elementet over. Her skulle vi gjerne hatt 24px mellomrom.**
![Skjermbilde 2022-08-09 kl  12 56 32](https://user-images.githubusercontent.com/33222679/183636923-52d9b6f4-182e-4a6d-83d1-d98962fa2570.png)

Skjermleser: Ok (VoiceOver) 
Små skjermer: Ok

Febakke:
Tested ok!
Both focus and spacing is fixed 👌
![Skjermbilde 2022-08-18 kl. 16.12.05.png](https://images.zenhubusercontent.com/5c8b643d6e79732923030a9c/44861c6f-5ab7-40de-899a-d6e31ed2c06f)

PR #362, opened 2022-08-02
Change save button for repeating groups

Magnusrm:
Change done button in repeating groups with the secondary button from altinn-designsystem.

Also add a new textResource as the needed text was not already present.

## Related Issue(s)
- #327 

## Verification
- [ ] All tests run green

## Documentation


src/altinn-app-frontend/src/features/form/containers/GroupContainer.test.tsx
```diff
@@ -244,7 +244,9 @@ describe('GroupContainer', () => {
     const store = render({ container: mockContainerInEditModeWithTrigger });
     const user = userEvent.setup();
 
-    const editButton = screen.getAllByText('Ferdig')[0].closest('button');
+    const editButton = screen
+      .getAllByText('Lagre og lukk')[0]
+      .closest('button');
```
haakemon, line 249: I think you can use something like this instead, which IMO looks a bit better:

```jsx
screen.getByRole('button', {
  name: /lagre og lukk/i
})
```

A good tool to help finding good selectors is https://testing-playground.com/ . You can use `screen.debug()` to print out the DOM to the console during the testrun, and paste it into testing-playground.com, and then click on items to get suggestions for selectors.

haakemon (approved):

PR #367, opened 2022-08-08
Cypress: Fuzzing and redux logging

olemartinorg:
## Description
Multiple adjustments to Cypress testing.

1. Adding an option called `responseFuzzing`. When enabled, all API responses from the backend will be artificially delayed a random amount of time, simulating a slow network, I/O load on the backend, etc. This also causes dispatched events in the frontend to sometimes be fulfilled out-of-order, and can thus be used to bring to light any issues where our functionality implicitly relied on server response times in order to update the redux state correctly (like what happened in #365 and #87). Running this right now causes multiple cypress tests to fail, so it's evident we have more problems like this.
2. (Not included in code here, but still relevant to these changes.) I added `crossorigin` to the `<script>` tags in all tested apps, so that Cypress is able to gather error details from the app when something fails.
3. Adding a way to extract the log of redux actions and save it a test artifact we can download from github after the action runs. This can be imported into redux devtools and help us figure out what went wrong in the order of actions. This is especially useful when the fuzzing tests fail, as they can be very difficult to reproduce otherwise. 

## Related Issue(s)
- #365
- #87

## Verification
- [x] All tests run green

## Documentation

### DIFF
--- a/test/cypress/e2e/fixtures/texts.json
+++ b/test/cypress/e2e/fixtures/texts.json
@@ -17,7 +17,7 @@
   "next": "Neste",
   "requiredField": "Feltet er påkrevd",
   "requiredFieldLastName": "Du må fylle ut nytt etternavn",
-  "requiredFieldDateFrom": "Du må fylle ut Dato for navneendring",
+  "requiredFieldDateFrom": "Du må fylle ut dato for navneendring",
   "securityReasons": "Av sikkerhetshensyn vil verken innholdet i tjenesten eller denne meldingen være synlig i Altinn etter at du har forlatt denne siden",
   "selectNewReportee": "Velg ny aktør under",
   "startingSoon": "Hold deg fast, nå starter vi!",
--- a/test/cypress/e2e/support/start-app-instance.js
+++ b/test/cypress/e2e/support/start-app-instance.js
@@ -12,6 +12,25 @@ Cypress.Commands.add('startAppInstance', (appName, anonymous=false) => {
     },
   };
 
+  if (Cypress.env('responseFuzzing') === 'on') {
+    const [min, max] = [10, 1000];
+    cy.log(`Response fuzzing on, will delay responses randomly between ${min}ms and ${max}ms`);
+
+    const rand = () => {
+      return Math.floor(Math.random() * (max - min + 1) + min);
+    };
+
+    const randomDelays = (req) => {
+      req.on('response', (res) => {
+        res.setDelay(rand());
+      });
+    };
+    cy.intercept('**/api/**', randomDelays);
+    cy.intercept('**/instances/**', randomDelays);
+  } else {
+    cy.log(`Response fuzzing off, enable with --env responseFuzzing=on`);
+  }
+
   cy.visit('/', visitOptions);
   if (Cypress.env('environment') === 'local') {
     if (anonymous) {
--- a/.github/workflows/cypress-altinn-app-frontend.yml
+++ b/.github/workflows/cypress-altinn-app-frontend.yml
@@ -60,6 +60,7 @@ jobs:
         working-directory: test/cypress
 
       - uses: actions/upload-artifact@v3
+        if: failure()
         with:
           name: redux-state-history
           path: test/cypress/redux-history/*.json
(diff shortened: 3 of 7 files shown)


test/cypress/e2e/fixtures/texts.json
```diff
@@ -17,7 +17,7 @@
   "next": "Neste",
   "requiredField": "Feltet er påkrevd",
   "requiredFieldLastName": "Du må fylle ut nytt etternavn",
-  "requiredFieldDateFrom": "Du må fylle ut Dato for navneendring",
+  "requiredFieldDateFrom": "Du må fylle ut dato for navneendring",
```
olemartinorg, line 20: I updated `frontend-test` to use a lowercase label for this field. Just a minor annoyance I had, that I thought I'd fix when I had to deploy a new version of `frontend-test` on friday.

test/cypress/e2e/support/start-app-instance.js
```diff
@@ -12,6 +12,25 @@ Cypress.Commands.add('startAppInstance', (appName, anonymous=false) => {
     },
   };
 
+  if (Cypress.env('responseFuzzing') === 'on') {
+    const [min, max] = [10, 1000];
+    cy.log(`Response fuzzing on, will delay responses randomly between ${min}ms and ${max}ms`);
```
haakemon, line 17: Just wondering if this will "lead to more flaky tests". What I mean by that is, since its random, the tests will pass sometimes (when delay is 10ms) and fail other times (when delay is 1000ms). Would it be better/less noisy to just use a slow variant (always 1000ms) all the time?

If we have some conditions where a slow response causes issues, we should be able to reliably reproduce and fix it.

olemartinorg: Heh, introducing flakiness is the whole point.. :wink: In principle, our app should work the exact same regardless of the backend using more or less time to handle an API request, but that's not always the case.

Let's say you have some code firing off request `A` to do something simple, then it fires off request `B` to do something more complex. Every time you run this locally, `A` will finish before `B`, so you end up unintentionally writing your code in a way where the reducer for `BFulfilled` depends on data from the `AFulfilled` reducer which is already in the store. If we use a random delay, `B` will sometimes finish before `A`, and this fuzzing would then show us an actual bug in our code we'd need to fix. If, however, we just use a fixed delay of 1000ms, `A` will still always finish before `B`, so no problems are found.

So, seemingly random response times from the server is what happens in the real world, and it should not affect our app at all. If/when it does, that shouldn't be considered a flaky test - it's actually a bug in our code, and we need to write more defensive code to make sure everything works as expected even when some requests (but not all) take long to complete.

olemartinorg: Btw, I don't think we're ready to run these nightly. I ran the `frontend-test` suite with fuzzing once on friday, and multiple tests failed - indicating we still have some of these bugs left. I'd like to clean those up before starting to run fuzzing tests periodically.

haakemon: Right I see what you mean. Makes sense to keep some randomisations then. But perhaps the values should be a bit higher in general(f.ex 500ms for min and 2500 for max?) so that cases where any slow response in general will fail consistenly?

haakemon (approved):

test/cypress/e2e/support/start-app-instance.js
olemartinorg: Yep, that's a good point! It will take a looong time to run tests on github actions then, though - there are many requests, so this quickly adds up. I'll do a few test runs so I can get some numbers on the effects of this.

haakemon: Not rush to change this until we start running this nightly though, more important to get this merged to unblock other PRs 🙂 We can always adjust this timeout later when we get some experience with it

olemartinorg: Results are in!

Without fuzzing: **10min 39sec** (all tests successful)
Fuzzing (10-1000ms): **22min 54sec** (`components.js`, `group.js` and `validation.js` each failed once)
Fuzzing (500-2500ms): **40min 54sec** (`components.js`, `group.js`, `validation.js` and `wcag.js` each failed once)

As this is supposed to be random, the timings are, of course, approximate. And again, as it's random, which tests fail when may also be random. Maybe I should just make the min/max configurable..

PR #368, opened 2022-08-08
Implement error message from design system

Magnusrm:
Implementing the new ErrorMessage component from the design system. 

This implementation removed the final use case for the MessageComponent, as the other functionality in the message component has been taken over by the SoftValidations component using the Panel component from design system. 

I therefore chose to remove the MessageComponent and its tests as it no longer was in use.

TODO:
- [x] Merge new ErrorComponent to the design system
- [x] Create new release of design system
- [x] Update the package version of design system in this PR

## Related Issue(s)
#99 

## Verification
- [ ] All tests run green

## Documentation


src/altinn-app-frontend/src/utils/render.tsx
```diff
@@ -74,14 +70,14 @@ export function renderValidationMessages(
...
-      style={messageComponentStyle}
-      key='error'
-      id={id}
-    >
-      <ol>{messages.map(validationMessagesToList)}</ol>
-    </MessageComponent>
+    <div style={{ paddingTop: '0.375rem' }}>
+      <ErrorMessage
+        id={id}
+        key='error'
```
haakemon, line 76: Key should be removed here, ref docs about how they are used https://reactjs.org/docs/lists-and-keys.html#keys

test/cypress/e2e/integration/app-frontend/validation.js
```diff
@@ -22,13 +22,12 @@ describe('Validation', () => {
       .blur()
       .focus()
       .clear()
-      .blur();
+      .blur()
     cy.get(appFrontend.fieldValidationError.replace('field', appFrontend.changeOfName.newFirstName.substring(1)))
       .should('exist')
       .should('be.visible')
-      .should('have.text', texts.requiredField)
-      .find(appFrontend.errorExclamation)
```
haakemon, line 30: I dont think `appFrontend.errorExclamation` is used anywhere else, so would be good to remove that definition as well

haakemon (approved):
Nice! ✨

Issue #371, opened 2022-08-09
Objects in redux state for validations crashes the app


olemartinorg:
### Description of the bug

In [this discussion on slack](https://altinn.slack.com/archives/C02ESDSC929/p1659963042393879), it was reported that simple operations on a group could cause the app to crash.

### Steps To Reproduce

1. In a repeating group with required fields, add a new row
2. Do not fill out the required fields
3. Try to save the row, observe there are validation errors since these fields have not been filled [note 1]
4. Delete the row with the validation errors present
5. Observe the crash

### Additional Information

Submitting this error report after triaging and fixing the bug.

This turned out to be an onion revealing multiple bugs. The root cause of this seems to be because some objects persisted in the validation state (originally React nodes) fail to render as react nodes, as React would crash with a cryptic message. When I cleaned up the redux state to not include react nodes/objects, that fix seem to have corrected the problem - but I was still unable to reproduce this in `frontend-test` and failed to write a regression test for it. It turns out the validation code did not correctly handle the repeating group in `frontend-test`, as it contains multiple pages. So, originally this bug failed to reproduce at [note 1] in `frontend-test`.

Magnusrm:
Tested following the reproduction steps and the issue seems to be fixed.

olemartinorg:
Thanks, @Magnusrm! I had some trouble reproducing this at first. (Mainly because our `frontend-test` app had a `multiPage` repeating group. For unrelated reasons, that skipped validation of required fields even with validation triggers.) I'll also contact the person who reported this to me on Slack as soon as this is released.

Issue #350, opened 2022-07-25
Markdown not supported in error messages


olemartinorg:
### Description of the bug

In our `frontend-test` app, setting the first fields inside the repeating group as required exposes a bug:

![20220725_12h20m39s_grim](https://user-images.githubusercontent.com/700139/180755169-9e0ef990-8c5e-40ab-8138-ee62d3c3c6cb.png)

These error messages should handle (or remove) markdown formatting. This is further complicated by #347, in case these texts are not just strings.

olemartinorg:
Should be fixed now, as a part of #372. Closing this and tracking this in #371 instead.

Issue #76, opened 2022-05-04
Repeating groups: multiPage functionality might breed bugs


olemartinorg:
## Introduction
For repeating groups, it is possible to enable `multiPage: true` and structure group components on multiple pages by prefixing the component IDs with page numbers ([as documented here](https://docs.altinn.studio/app/development/ux/fields/grouping/alternatives/#flere-sider-innad-i-gruppe-visning)). As proved by [#7478](https://github.com/altinn/altinn-studio/issues/7478) (and the fix for it in #73), the decision to mix multiple configuration properties into the strings in `group.children` broke some code that assumed `group.children` is just a list of component IDs.

#### Examples where this is handled

While searching for code using `group.children`, I found that about half of (non-test) usage already contain code branches to remove this page number:

https://github.com/Altinn/app-frontend-react/blob/871e4e58dee11eabbc9e95063dfa2bcb07fecaa8/src/altinn-app-frontend/src/utils/formLayout.ts#L29-L39

https://github.com/Altinn/app-frontend-react/blob/871e4e58dee11eabbc9e95063dfa2bcb07fecaa8/src/altinn-app-frontend/src/utils/formLayout.ts#L63-L72

https://github.com/Altinn/app-frontend-react/blob/871e4e58dee11eabbc9e95063dfa2bcb07fecaa8/src/altinn-app-frontend/src/features/form/containers/RepeatingGroupsEditContainer.tsx#L166-L178

https://github.com/Altinn/app-frontend-react/blob/871e4e58dee11eabbc9e95063dfa2bcb07fecaa8/src/altinn-app-frontend/src/features/form/containers/Form.tsx#L38-L44

https://github.com/Altinn/app-frontend-react/blob/871e4e58dee11eabbc9e95063dfa2bcb07fecaa8/src/altinn-app-frontend/src/features/form/containers/Form.tsx#L94-L100

https://github.com/Altinn/app-frontend-react/blob/871e4e58dee11eabbc9e95063dfa2bcb07fecaa8/src/altinn-app-frontend/src/features/form/layout/update/updateFormLayoutSagas.ts#L73-L75

#### Examples where this is NOT handled

The rest of `group.children` usage do not have any special branches for `multiPage: true`, and while I haven't created any proofs for it, I believe some of these code paths may contain bugs were `multiPage: true` enabled for these groups:

[...]

PR #372, opened 2022-08-09
Removing complex objects from redux store

olemartinorg:
## Description
This removes `React.Node` objects from the redux store, making our states properly serializable. In addition to this, it fixes a problem where validation behaviour for groups with `edit.multiPage` would lead to required fields inside the group not being validated when saving the group row.

## Related Issue(s)
- #371 (the main bug report this aimed to fix)
- #350 (this should also be fixed as a consequence)
- #76 (related bug report detailing the over-arching issue causing part of this bug)

## Verification
- [x] All tests run green

## Documentation


nkylstad (approved):
Fantastic! 🏅 This simplifies so much ❤️

PR #383, opened 2022-08-15
Fix bug with repeating groups only showing 9 rows

lasseklovstad:
## Description
Fixed a bug with repeating groups where it would only show max 9 rows, even if you had a much larger datamodel. 
We discovered the bug when implementing the Likert component with rows above 10.

I reproduced the bug in a test and fixed it!

## Verification
- [x] All tests run green


olemartinorg (approved):
Great fix! :100: Thank you!

Issue #58, opened 2022-04-27
User icon and logout feature missing when logged in through Feide


allinox:
## Description

In regular apps, you can log out by clicking on the user icon in the top right corner while in app-view or on the receipt page.
![image](https://user-images.githubusercontent.com/22012997/165465369-42e07c90-d41b-4438-a2ec-5ae25c5f9c14.png)

However, when logged in through Feide, this icon is missing
![image](https://user-images.githubusercontent.com/22012997/165465453-e0135bab-8016-4c0e-b533-ec7499875cc0.png)
This means that you can't log out through this feature.

## To Reproduce

I would expect that the icon is present in the top right corner, like is the case in regular apps.

lasseklovstad:
This issue should be closed as the problem is fixed in the above pull request. @allinox could you test it and see if Feide login has the icon?

allinox:
Tested in TT02 - it works!🥳
![image](https://user-images.githubusercontent.com/22012997/207551291-14e4a662-640e-4b1f-a974-840de2c9bf87.png)

PR #384, opened 2022-08-15
BugFix: Show menu icon when user doesn't have ssn or org number

lasseklovstad:
## Description
Shows the private person icon as a fallback if ssn or orgNumber is not present in user party.

## Related Issue(s)
- #58 

## Verification
- [x] All tests run green

### DIFF
--- a/src/shared/src/components/organisms/AltinnAppHeader.test.tsx
+++ b/src/shared/src/components/organisms/AltinnAppHeader.test.tsx
@@ -17,7 +17,7 @@ describe('organisms/AltinnAppHeader', () => {
     name: 'Bedrift',
   } as IParty;
 
-  const anonymousUser = {
+  const selfIdentifiedUser = {
     childParties: null,
     isDeleted: false,
     name: 'uidp_brxzt8pt992',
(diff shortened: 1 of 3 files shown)


olemartinorg:
Thanks for your contribution, @lasseklovstad! :partying_face: I've made inquiries to make sure we correctly handle anonymous stateless apps here, etc. This is not the part of the app I know best, at least. You'll hear from us soon!

src/shared/src/components/organisms/AltinnAppHeader.test.tsx
```diff
@@ -16,6 +16,21 @@ describe('organisms/AltinnAppHeader', () => {
     partyId: '54321',
     name: 'Bedrift',
   } as IParty;
+
+  const anonymousUser = {
```
nkylstad, line 20: This might be nitpicking, but I would prefer it if you could rename this to something like `selfIdentifiedUser`. We use the term `anonymous` elsewhere in the app frontend to denote a user who is _not logget in at all_, and even though this is just a variable in a test it might be confusing to use the anonymous naming here.

nkylstad (changes requested):
Nice! 👍  One comment on naming that you can take a look at.

nkylstad:
> Thanks for your contribution, @lasseklovstad! 🥳 I've made inquiries to make sure we correctly handle anonymous stateless apps here, etc. This is not the part of the app I know best, at least. You'll hear from us soon!

Pretty sure that anonymous (not logged in) users do not have a party object at all, so this change should not affect that. The cypress tests for stateless anonymous pass, which seems to confirm this 🙂

nkylstad (approved):
🎉

PR #386, opened 2022-08-16
chore: upgrade to latest design system package

haakemon:
## Description
- new version is published on npm feed instead of github, so we no longer need to set up a PAT


olemartinorg (approved):
Fantastic! :tada:

PR #387, opened 2022-08-16
chore: attempt to skip SonarCloud run on external PR's

haakemon:
## Description
Currently, fork PR's still try to run SonarCloud scan, which will not work since they will not have access to the required secret. This should skip it for external PRs.


olemartinorg (approved):
Hope it works! :+1:

Issue #322, opened 2022-07-05
Delete-row-button should be moved out of the group and into the table.


Febakke:
### Description

Today our delete button is inside each element in the repeating group. We should move it out of the group and place it next to the edit button.

## Sketches
![Skjermbilde 2022-07-05 kl  14 49 37](https://user-images.githubusercontent.com/33222679/177331477-42f6ef5c-61c7-4295-b783-770ab6f14cf4.png)
![Skjermbilde 2022-07-05 kl  14 50 17](https://user-images.githubusercontent.com/33222679/177331597-a9314dc5-b75c-4c9e-9e92-e5c07ec05a3a.png)

[Figma link](https://www.figma.com/file/wnBveAG2ikUspFsQwM3GNE/Altinn-Studio-Apps?node-id=13550%3A9415)

## Acceptance criterea
- [ ] the edit and delete buttons should show as icons only when screen is smaller than 768. 
- [ ] The button is moved outside of the group

lorang92:
@Magnusrm functionality tested ok. I see that our icons in the table header cells 
![image](https://user-images.githubusercontent.com/10416981/187204429-c19babe5-a119-497c-83c8-1452d352fc41.png)
 does not have aria-labels 
![image](https://user-images.githubusercontent.com/10416981/187204279-9dc87dd8-971b-4006-8cc5-c555ba97a002.png)

I guess this should be added before we close this issue.

Magnusrm:
Thanks! I will add the aria-labels then 👍

PR #400, opened 2022-08-22
Move delete-row button from edit-group into table

Magnusrm:
## Description
Move delete-functionality from RepeatingGroupsEditContainer to RepeatingGroupTable.
Move delete button besides edit button in the table.
Add styling for the edit and delete buttons for mobileview.

## Related Issue(s)
- #322 

## Verification
- [x] All tests run green

## Documentation


src/altinn-app-frontend/src/features/form/containers/RepeatingGroupsTable.test.tsx
```diff
@@ -14,21 +17,25 @@ import type { ILayoutState } from 'src/features/form/layout/formLayoutSlice';
...
-  let mockRepeatingGroupDeepCopyComponents: Array<
-    Array<ILayoutComponent | ILayoutGroup>
-  >;
+import type { ILanguage } from 'altinn-shared/types';
+
+const user = userEvent.setup();
 
+let mockContainer: ILayoutGroup;
+let mockLanguage: ILanguage;
+let mockTextResources: ITextResource[];
```
lorang92, line 26: I know you probably just continued on existing code here, but is there really a need to declare these variables before each run? I'm assuming they never actually change.

Could we not just declare them as const and remove the `beforeEach` all together? :)

src/altinn-app-frontend/src/features/form/containers/RepeatingGroupsTable.test.tsx
```diff
@@ -127,28 +145,99 @@ describe('RepeatingGroupTable', () => {
...
+
+    it('should render edit and delete buttons as icons for screens smaller thnn 786px', () => {
+      const onClickRemove = jest.fn();
+      render({ onClickRemove: onClickRemove });
+
+      const iconButtonsDelete = screen.getAllByTestId(/delete-button/i);
+      const iconButtonsEdit = screen.getAllByTestId(/edit-button/i);
+
+      expect(iconButtonsDelete).toHaveLength(4);
+      expect(iconButtonsDelete[0]).toContainHTML(
```
lorang92, line 207: What are we really interested in testing with these `toContainHTML` tests? Seems like they are almost like snapshot tests to me. Prone to breaking if we change some css classes for instance.

If you don't have a good reason for testing this I suggest just removing them.

lorang92 (changes requested):
Great stuff! 

Some minor comments 🦖

src/altinn-app-frontend/src/features/form/containers/RepeatingGroupsTable.test.tsx
Magnusrm: I wanted to test that the buttons were rendered with only the icon and no text for the mobile view. I guess a other way of doing it could be to query for elements with the button texts and assert the result to null? The main reason i added this test was for code coverage for the SonarQube gods.

lorang92: Yeah, a negative query with the button text sounds better 👍

PR #403, opened 2022-08-24
BugFix: Close button not hidden when specified in layoutsettings

lasseklovstad:
## Description
You can now add page settings in layout-sets.json:
```json
{
  "sets": [
    {
      "id": "student-survey",
      "dataType": "survey",
      "tasks": [
        "Task_1"
      ]
    },
    {
      "id": "info-page",
      "dataType": "statelessData",
      "tasks": [
        "Task_2"
      ]
    }
  ],
  "uiSettings": {
    "hideCloseButton": true,
    "showLanguageSelector": true
  }
}

```

Relevant properties are: triggers, hideCloseButon, showLanguageSelector and showProgress. These wil act as the initial settings for the whole app. Page settings defined in Settings.json will override the initial settings when loaded.

## Related Issue(s)
- #401

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/altinn-app-frontend/src/features/form/layout/formLayoutSlice.ts
+++ b/src/altinn-app-frontend/src/features/form/layout/formLayoutSlice.ts
@@ -22,7 +22,7 @@ import { createSagaSlice } from 'src/shared/resources/utils/sagaSlice';
 import type { ILayouts } from 'src/features/form/layout';
 import type * as LayoutTypes from 'src/features/form/layout/formLayoutTypes';
 import type { MkActionType } from 'src/shared/resources/utils/sagaSlice';
-import type { ILayoutSets, IUiConfig } from 'src/types';
+import type { ILayoutSets, IPagesSettings, IUiConfig } from 'src/types';
 
 export interface ILayoutState {
   layouts: ILayouts;
@@ -45,9 +45,6 @@ export const initialState: ILayoutState = {
     layoutOrder: null,
     pageTriggers: [],
     keepScrollPos: undefined,
-    hideCloseButton: true,
-    showProgress: false,
-    showLanguageSelector: false,
   },
   layoutsets: null,
 };
@@ -89,7 +86,12 @@ const formLayoutSlice = createSagaSlice(
       fetchSetsFulfilled: mkAction<LayoutTypes.IFetchLayoutSetsFulfilled>({
         reducer: (state, action) => {
           const { layoutSets } = action.payload;
-          state.layoutsets = layoutSets;
+          if (layoutSets.sets) {
+            state.layoutsets = { sets: layoutSets.sets };
+          }
+          if (layoutSets.uiSettings) {
+            updateCommonPageSettings(state, layoutSets.uiSettings);
+          }
         },
       }),
       fetchSetsRejected: mkAction<LayoutTypes.IFormLayoutActionRejected>({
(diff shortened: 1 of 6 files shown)


mjulstein (changes requested):
I would suggest that this is not really a issue with the application itself, that the close button returns to Altinn is more likely the problem.
Could you maybe change it so that you can set the on-close re-direct instead to some other route than Altinn?
By changing the rendering conditions for the close button the cypress-tests looking it.
I think the main issue is that we do not load the settings of the app on deep linking to a receipt.
It would make more sense that the app-settings will be loaded and read on deep links as well.

lasseklovstad:
> I would suggest that this is not really a issue with the application itself, that the close button returns to Altinn is more likely the problem. Could you maybe change it so that you can set the on-close re-direct instead to some other route than Altinn? By changing the rendering conditions for the close button the cypress-tests looking it. I think the main issue is that we do not load the settings of the app on deep linking to a receipt. It would make more sense that the app-settings will be loaded and read on deep links as well.

That sounds like a good idee! Were do you suggest putting this setting-variable? Maybe same place as `hideCloseButton` is?

lasseklovstad:
@mjulstein I started checking out how to override the on-close redirect. If I put the settings in same file as `hideCloseButton` we still have the same problem as before because the close button is visible when the app is loading the form layout settings. So I would have to have this settings in `applicationmetadata.json`. 

But to include a new value in applicationmetadata.json i would have to change the endpoint implementation, because it filters out values not specified. That means a PR in this repo aswell: https://github.com/Altinn/app-lib-dotnet 

Any thoughts? Should I continue this aproach?

mjulstein:
@lasseklovstad Could you maybe make it so that you add a flag to the url you are sharing or typing instead?
Ie. `?closable=false` 
`https://org.app?closeable=false/#/instanceid/somethingBlaBla/receipt`
Then you can simply share any url you want and still hide that close-button regardless of the config.

lasseklovstad:
@mjulstein In our case we would like the url to be as clean as possible. Because in some cases the survey url would be shared via feks. the teacher writing it on the blackboard to school children down to 5th grade.

lasseklovstad:
@nkylstad @mjulstein Tror jeg har implementert det slik beskrevet av Nina nå. Og alle testene ser ut til å bli grønne. Litt usikker på hva slags tester jeg børe skrive rundt dette. Noen forslag?

src/altinn-app-frontend/src/features/form/layout/formLayoutSlice.ts
```diff
@@ -325,5 +324,25 @@ const formLayoutSlice = createSagaSlice(
...
+  >,
+) => {
+  const {
+    hideCloseButton = state.uiConfig.hideCloseButton,
+    showLanguageSelector = state.uiConfig.showLanguageSelector,
+    showProgress = state.uiConfig.showProgress,
+    triggers = state.uiConfig.pageTriggers,
+  } = page;
+
+  state.uiConfig.hideCloseButton = hideCloseButton;
```
mjulstein, line 341: Could this cause a condition where the global state is not set, but when going through a form it might be set by the form, and is hidden on the next process?

mjulstein (approved):
Looks great :)
Disregard the other comment. It is not likely that it is a problem.

mjulstein:
> @nkylstad @mjulstein Tror jeg har implementert det slik beskrevet av Nina nå. Og alle testene ser ut til å bli grønne. Litt usikker på hva slags tester jeg børe skrive rundt dette. Noen forslag?

I guess you can write some tests to check the residence of the settings.
As long as your revision gives the same experience i any app that did not have a "hideCloseButton" setting then i don't think there is much to test for.

Issue #388, opened 2022-08-16
PDF-generator needs a "page render completed" indicator 


SandGrainOne:
### Description

The new PDF-generator currently being planned will be using puppeteer and chrome to convert an app-page to PDF. To do this as flexibly as possible, puppeteer needs a "hook" on the page(s) being rendered that indicate that the page is ready.

In puppeteer this is called the [waitForSelector](https://pptr.dev/api/puppeteer.page.waitforselector/).

Ideally the PDF-generator should be able to wait for the same element regardless of page being rendered. There might already be such a DOM element, but the platform team will need help identifying what to use. 

We would like to find such a DOM element on all pages instead of restricting what pages that can be used to create a PDF.

olemartinorg:
Closing, as this has been tested OK by @SandGrainOne

PR #423, opened 2022-08-30
Adding an element for the PDF generator to know when the page has loaded

olemartinorg:
## Description
This element should be injected everywhere in our app at the points where all content has been loaded and the DOM is ready for the user (i.e., after all our loading placeholders). This element (`#readyForPrint`) can be listened for by the PDF generator to know when the page can be printed to a PDF.

As it's notoriously hard to test these things, and I'm not sure how (or even if) we should include this in our Cypress test suite, I opened up the debugger in Chrome and pasted this code inside a file to make sure we trigger a breakpoint (which freezes rendering) when the `readyForPrint` element appears on the page. It's not pretty, but it's what made me find the issue with images that have to be loaded.

```typescript
let iterations = 0;
let found = false;
setInterval(() => {
  iterations++;
  const elmFound = !!document.getElementById('readyForPrint');
  if (!found && elmFound) {
    found = true;
    console.log(
      'Found readyForPrint after',
      iterations,
      'iterations',
      document.readyState,
    );
    debugger;
  } else if (found && !elmFound) {
    found = false;
    iterations = 0;
  }
}, 5);
```

I'm not sure about the naming - maybe we should use a more generic name like `altinnAppReady` instead?

## Related Issue(s)
- #388

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/altinn-app-frontend/src/shared/components/ReadyForPrint.tsx
+++ b/src/altinn-app-frontend/src/shared/components/ReadyForPrint.tsx
@@ -1,0 +1,48 @@
+import React from 'react';
+
+/**
+ * This element only serves to let our PDF generator know the app is ready and have rendered its content.
+ * It should be included in the app DOM for every possible execution path, except those where we're showing
+ * loading indicators to the user while waiting for content to get ready.
+ */
+export function ReadyForPrint() {
+  const [imagesLoaded, setImagesLoaded] = React.useState(false);
+
+  React.useLayoutEffect(() => {
+    const promises = [];
+
+    const imageLoadPromise = (img: HTMLImageElement) => {
+      return new Promise((res) => {
+        img.addEventListener('load', res);
+        img.addEventListener('error', res);
+      });
+    };
+
+    document.querySelectorAll('img').forEach((image) => {
+      image.complete || promises.push(imageLoadPromise(image));
+    });
+
+    Promise.all(promises).then(() => {
+      // All images have loaded (or failed to load), but they haven't always been painted/caused a re-render. A request
+      // for an animation frame reserves you a slot to execute some code _before the next repaint_, but that might be
+      // the repaint where the image was supposed to render. When we call it twice, we're guaranteed to output our
+      // element after all images have been rendered.
+      requestAnimationFrame(() => {
+        requestAnimationFrame(() => {
+          setImagesLoaded(true);
+        });
+      });
+    });
+  }, []);
+
+  if (!imagesLoaded) {
+    return;
+  }
+
+  return (
+    <div
+      style={{ display: 'none' }}
+      id='readyForPrint'
+    />
+  );
+}
--- a/src/altinn-app-frontend/src/features/confirm/containers/Confirm.tsx
+++ b/src/altinn-app-frontend/src/features/confirm/containers/Confirm.tsx
@@ -9,6 +9,7 @@ import {
 } from 'src/common/hooks';
 import { ValidationActions } from 'src/features/form/validation/validationSlice';
 import { selectAppName } from 'src/selectors/language';
+import { ReadyForPrint } from 'src/shared/components/ReadyForPrint';
 import { InstanceDataActions } from 'src/shared/resources/instanceData/instanceDataSlice';
 import { ProcessActions } from 'src/shared/resources/process/processSlice';
 import { getValidationUrl } from 'src/utils/appUrlHelper';
@@ -191,6 +192,7 @@ const Confirm = () => {
             pdf={getInstancePdf(instance.data)}
           />
           <SubmitButton />
+          <ReadyForPrint />
         </>
       )}
     </div>
--- a/src/altinn-app-frontend/src/features/feedback/Feedback.tsx
+++ b/src/altinn-app-frontend/src/features/feedback/Feedback.tsx
@@ -3,6 +3,7 @@ import * as React from 'react';
 import { createTheme, MuiThemeProvider, Typography } from '@material-ui/core';
 
 import { useAppDispatch, useAppSelector } from 'src/common/hooks';
+import { ReadyForPrint } from 'src/shared/components/ReadyForPrint';
 import { ProcessActions } from 'src/shared/resources/process/processSlice';
 import { getTextFromAppOrDefault } from 'src/utils/textResource';
 
@@ -34,6 +35,7 @@ export default function Feedback() {
           {getTextFromAppOrDefault('feedback.body', textResources, language)}
         </Typography>
       </MuiThemeProvider>
+      <ReadyForPrint />
     </div>
   );
 }
--- a/src/altinn-app-frontend/src/features/form/containers/Form.tsx
+++ b/src/altinn-app-frontend/src/features/form/containers/Form.tsx
@@ -10,6 +10,7 @@ import { DisplayGroupContainer } from 'src/features/form/containers/DisplayGroup
 import { mapGroupComponents } from 'src/features/form/containers/formUtils';
 import { GroupContainer } from 'src/features/form/containers/GroupContainer';
 import { PanelGroupContainer } from 'src/features/form/containers/PanelGroupContainer';
+import { ReadyForPrint } from 'src/shared/components/ReadyForPrint';
 import {
   extractBottomButtons,
   hasRequiredFields,
@@ -150,6 +151,7 @@ export function Form() {
         )}
         <ErrorReport components={errorReportComponents} />
       </Grid>
+      <ReadyForPrint />
     </>
   );
 }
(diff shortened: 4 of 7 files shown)


src/altinn-app-frontend/src/shared/components/ReadyForPrint.tsx
```diff
@@ -0,0 +1,48 @@
...
+        img.addEventListener('load', res);
+        img.addEventListener('error', res);
+      });
+    };
+
+    document.querySelectorAll('img').forEach((image) => {
+      image.complete || promises.push(imageLoadPromise(image));
+    });
+
+    Promise.all(promises).then(() => {
```
haakemon, line 25: Can we be sure that after all images are loaded/painted, all other content has been loaded? 

What about pages where there are no images?

src/altinn-app-frontend/src/shared/components/ReadyForPrint.tsx
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
+
+/**
+ * This element only serves to let our PDF generator know the app is ready and have rendered its content.
+ * It should be included in the app DOM for every possible execution path, except those where we're showing
```
haakemon, line 5: Could it be moved to the root instead, so we only have to include it once? Just thinking it might be difficult to remember to add this component to new areas.

src/altinn-app-frontend/src/shared/components/ReadyForPrint.tsx
olemartinorg: No, we can't really be sure of that - but I can't think of anything else that is added to the DOM but won't be displayed until some later stage (when it is loaded). We're not loading external CSS at any point either. The only thing I could think of is we should make sure we're not waiting for any sagas currently not fulfilled/rejected, but I can't think of a way to know that for sure (not all sagas have to result in anything either).

For pages with no images, there are no promises to await, so `Promise.all()` resolves immediately - so that shouldn't be a problem.

src/altinn-app-frontend/src/shared/components/ReadyForPrint.tsx
olemartinorg: I know, yes, maintainability is the biggest problem... Maybe I could hack together something that makes sure to wait until there are no `AltinnLoader` elements in the page - that could probably work! Let me research that for a bit. Thanks for the idea!

src/altinn-app-frontend/src/shared/components/ReadyForPrint.tsx
haakemon: > For pages with no images, there are no promises to await, so Promise.all() resolves immediately - so that shouldn't be a problem.

Wont this be a problem, since the promises resolves immediately, and the content may not have loaded yet? :) 

Any app could also potentially load external css (or js for that matter)

src/altinn-app-frontend/src/shared/components/ReadyForPrint.tsx
haakemon: We are not guaranteed to use AltinnLoaders when loading content either, so not sure this approach will be any more bulletproof than the current solution.

src/altinn-app-frontend/src/shared/components/ReadyForPrint.tsx
olemartinorg: By using `useLayoutEffect`, this runs after the DOM mutations, so all `img` elements should be loaded into the DOM at that point - ready for us to wait until they load.

And yes, any other resources may also be loaded - but I don't think that's a problem here. I'm assuming Puppeteer will wait for the document `readyState` to settle - but I guess I could add a state check for that as well. The core problem here is that while the page looks to be ready (i.e. the `readyState` indicates the page is ready, and the DOM has finished painting), the Altinn app is still not ready (because our initializing sagas are still running, fetching data, and the page is showing you the loading indicator).

src/altinn-app-frontend/src/shared/components/ReadyForPrint.tsx
olemartinorg: Right, but it might be better than the alternative, though? So, in principle, either:
1. Hope we'll remember to include this element everywhere in future execution paths, after content has been loaded. If we don't, the PDF generation might time-out (resulting in either PDFs not being generated at all, or the processing taking longer than they need to).
2. Hope we'll remember to update the implementation of this element every time we add/change components used to display "we're loading your data" indicators (such as spinners). The consequence if we don't is PDFs being generated too early, sometimes simply just showing the loading indicator instead of actual data.

I prefer the second alternative here, as I think it's slightly more maintainable. I'm basing that assumption on my other assumptions that:
- We'll probably add branching paths in the app loading procedure more often than we're adding loading indicator components
- Users/app developers will be more likely to let us know if a PDF shows them something entirely useless rather than just the PDF taking a bit longer than usual.

Also, I haven't started experimenting with a hack, but I think I can implement some console warnings/errors should we render the `ReadyForPrint` element a bit too early (by detecting near-immediate DOM changes after the render not caused by user interaction - usually indicating the `ReadyForPrint` element was not rendered late enough).

haakemon: I think the suggestion is good, lets try it :)

olemartinorg: @ivarne That's a good point - no, I guess that would be hard (retroactively replacing data in instances) to impossible (replacing files users have downloaded).

In this issue, the working assumption is that the PDF generator is 'blind', in that it loads a page, doesn't know exactly what it is looking at, and just prints it. I'm also assuming this is a stop-gap for a future solution where `app-frontend-react` gets some kind of indication that an instance should be presented in a printable manner, and thus generates the proper DOM for a ready-to-print presentation. In that case, the PDF generator could also be made to perform some checks to make sure it 'sees' something similar to what it is supposed to see (making it 'non-blind'). @SandGrainOne can probably give us some insights here on the future PDF plans (and the way it will work alongside `app-frontend-react`). If it is blind, and we only have a simple high-level element to indicate the page is (or hopefully is) ready to print, the risk of future changes to `app-frontend-react` breaking PDFs is quite high, IMO.

SandGrainOne: This is indeed a stop-gap and a work in progress. The immediate goal it to collect som experience with the new PDF solution. What kind of performance are we getting and what type of issues can occur.

olemartinorg: @haakemon I looked at a more generic and maintainable solution, but I firmly believe there's a time limit to how long it should be worth spending on this. I went down a route using `MutationObserver` on the root node with some heuristics to guess at which point the app has started rendering its content, and look for loading-indicator components inside it. Could be cool, but I'm not having any luck getting `MutationObserver` to actually do my bidding.

In summary, I think we should go ahead with this (suboptimal) solution, and rather create an issue for us to remove it later (when a proper interaction with the PDF generator is implemented).

Some details for those who want to know why this seemingly simple task could be this difficult to implement: In principle, we want an indicator to know when the page is "ready", and content has rendered. We don't know which content is rendered, and we want to implement it in a way (unlike the current implementation) where any future content alternatives also get covered - so developers don't have to know to include this indicator. If the page is displaying a loader/spinner, we should not add the indicator to the DOM, and if we'll show a loader/spinner in the future we also cannot add the indicator to the DOM *yet*. The React app includes code paths (for example in `App.tsx`) that will short-circuit the rendering and return a `null` (in practice defer rendering and show a white page for a little bit longer), and we need to detect that as a non-ready state. We also have wrappers around `App` that outputs empty `div` elements - so even though those are HTML elements, they shouldn't be considered content. We can solve a lot with heuristics (i.e. defining a 'ready' page to be one with more than 3 elements inside the root container, where none of them are loaders/spinners), but even that is not a guaranteed solution (a few more `App` wrappers producing empty `div`s, and that thing breaks). We could simply move the indicator inside `App.tsx`, but then it would break again the same way when `PartySelection.tsx` loads (as it will return `null` until it has a language). We could replace all `null` values with an element that actively blocks the ready-indicator, but then we're still not guaranteed a developer will not just add a new `return null;` somewhere else in the future and break the whole thing. </rant>

TL;DR: This thing is not great, but it's good enough for a temporary solution. Let's make sure it's temporary, and move to make a regression-testable proper PDF 'mode' in the future.

haakemon: I agree - I think the suggestion you have added is a good start to get some more experience with the new PDF generator. We will likely need to make adjustments as we go.

haakemon (approved):

Issue #429, opened 2022-09-01
Extra files added by the build - process in the dist folder referenced by altinn-app-frontend.js are not deployed to cdn


chrisnruud:
### Description of the bug

The deploy - to cdn step in the build process is designed to only deploy altinn-app-frontend.js and altinn-app-frontend.css , and after implemenmting the Map - component in altinn-design-system, new icon files for the map has been generated in the dist - folder that are NOT deployed, causing the map to render improperly.

Se live demo at https://dsb.apps.tt02.altinn.no/dsb/schema-with-map-component that renders the marker as "marker" not icon because the files are getting a 404 from the cdn

### Steps To Reproduce

moved from https://github.com/Altinn/altinn-design-system/issues/99

chrisnruud:
Testet OK

PR #430, opened 2022-09-02
Uploading all files to CDN on release

olemartinorg:
## Description
This should add all files from the `dist` folder to CDN, not just the js/css file. The only obvious weakness is that it will not delete previous files no longer part of the `dist` folder.

## Related Issue(s)
- #429

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -60,9 +60,11 @@ jobs:
           cd ..
           echo Copy Major Version
           mkdir -p ./cdn/toolkits/altinn-app-frontend/${{ steps.prepare.outputs.APP_MAJOR }}
+          git rm ./cdn/toolkits/altinn-app-frontend/${{ steps.prepare.outputs.APP_MAJOR }}/*
           cp -fr ./app-frontend/src/altinn-app-frontend/dist/* ./cdn/toolkits/altinn-app-frontend/${{ steps.prepare.outputs.APP_MAJOR }}/
           echo Copy Minor Version
           mkdir -p ./cdn/toolkits/altinn-app-frontend/${{ steps.prepare.outputs.APP_MAJOR_MINOR }}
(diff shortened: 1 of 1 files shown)


olemartinorg:
Converting to draft: There are outstanding issues I should be figuring out here first:
1. What to do about the `altinn-app-frontend.js.LICENSE.txt` file
2. Could we easily clean up removed files to avoid cluttering the major/minor versions?

haakemon:
> Converting to draft: There are outstanding issues I should be figuring out here first:
> 
> 1. What to do about the `altinn-app-frontend.js.LICENSE.txt` file
> 2. Could we easily clean up removed files to avoid cluttering the major/minor versions?

I think the proposed solution looks like a good solution. I dont think theres any reason to not publish the entire dist folder? There may be other assets included later as well that will need to be deployed, and this will avoi similar issues like the leaflet issue in the future.

1. Shouldnt we just publish the license file as well?
2. Not sure I understood this question

chrisnruud:
2. the action does a mkdir that seems to be unique for the current run anyway, so the dir should be clean?

olemartinorg:
1. @haakemon Yes, I agree.
2. @haakemon Say you have `some.artifact.js` which is introduced in v3.15, but removed again in v3.16. In the previous code it would still be present in the folder when v3.16 is copied into the v3 folder. I tested out removing previous files now, and the git log looks clean (git correctly deduces that you wanted a `M` for existing files, not a `D` and `A`, if that makes sense)
@chrisnruud Yup, but we run `mkdir -p`, so it's just to make sure the folder existed if it didn't exist already - it won't empty the folder.

haakemon:
Ah, I get it now. I didnt think about the fact that its actually checking out the CDN repo as well. Thanks for the clarification. The workaround you added seems like a good approach to me.

haakemon (approved):

PR #438, opened 2022-09-06
Fix: Nullcheck layoutSets

mjulstein:
This variable did not get checked and causes some apps to crash.


olemartinorg (approved):
Looks good! (And harmless.)

altinnadmin (approved):
LGTM

PR #447, opened 2022-09-07
bug: race condition conditional rules

lorang92:
## Description
Solves a race condition [reported by SSB](https://altinn.slack.com/archives/C02EVE4RU82/p1662552278562939). For dynamics for repeating groups to work the repeating group state must be set before running the rules. This fix prevents the rules running before this init is completed.

## Verification
- [x] **Your** code builds clean without any errors or warnings
- [x] Manual testing done (required)
- [x] Relevant automated test added (if you find this hard, leave it and we'll help out)
- [x] All tests run green

## Documentation
- [x] User documentation is updated with a separate linked PR in [altinn-studio-docs.](https://github.com/Altinn/altinn-studio-docs) (if applicable)

PR #450, opened 2022-09-08
Implementing missing loading state for app tasks

olemartinorg:
## Description
The `isLoading` state/slice was made to wait for sagas to complete, marking the app as ready when resources had been loaded. Sadly, no such state existed for the actions dispatched in `startInitialAppTaskQueue` (such as fetching text resources, language, etc). This could lead to the app rendering its content before the text resources it needed had been fetched (easily reproducible by using a proxy such as [Charles](https://www.charlesproxy.com/) to halt the request for a while - causing the app to render without the correct language resources).

I consider this fix to be somewhat *risky*, as it could lead to an app loading forever if any of the actions are supposed to be rejected. It looks fine to me when reading through the actions themselves, but I'm wary of yet again introducing show-stopping problems like we've seen recently.

## Related Issue(s)
- #447 (found in the org/ssb app after this problem had been fixed)

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/altinn-app-frontend/__mocks__/initialStateMock.ts
+++ b/src/altinn-app-frontend/__mocks__/initialStateMock.ts
@@ -50,6 +50,7 @@ export function getInitialStateMock(
       instantiating: null,
     },
     isLoading: {
+      appTask: null,
       dataTask: false,
       stateless: null,
     },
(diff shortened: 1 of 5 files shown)


olemartinorg:
This failed spectacularly in our Cypress tests (thank you, Cypress!). Holding it back for now while I try to figure out the real root cause for this problem (this probably wasn't it, because loading the text resources a bit later should make React re-render the components).

PR #462, opened 2022-09-12
Move edit-container into table

Magnusrm:
Move edit-container into table for repeating groups with edit.mode.showTable and update styling for edit-container. 

## Description
- Editing an element in a repeating group will now open the edit-container in a row directly below the chosen element. 
- Creating a new element will add an element to the bottom of the group and open the edit-container for the new element.
- This is also implemented for the mobile-view component for repeating groups.

The edit-container is placed within a row in the table that is placed directly below the row of the element being edited. In order to match the design in figma, the currently edited row and the one above it are being applied style classes when the edit-container is open.

There are no changes to the functionality of the edit-container for repeating groups using `edit.mode: 'showAll'` or `edit.mode: 'hideTable'` other than style changes.

## Related Issue(s)
- #243 

## Verification
- [ ] All tests run green

## Documentation


src/altinn-app-frontend/src/features/form/containers/RepeatingGroupTable.tsx
```diff
@@ -136,8 +141,39 @@ const useStyles = makeStyles({
...
+  editContainerRow: {
+    '&:hover': {
+      background: 'unset !important',
+    },
+  },
+  editingRow: {
+    backgroundColor: 'rgba(227, 247, 255, 0.5)',
+    '& td': {
+      borderBottom: 0,
+      '&:nth-child(1)': {
```
olemartinorg, line 157: I just glanced at this, but could we have a more specific class here? :thinking: It's hard for me to understand what 'the second child of every `td`' means while reading this, at least. I also noticed below that you add some padding for every `span` inside here - that seems dangerous if this container can include anything.

Magnusrm: This whole block of css is in order to get a dotted top-border for the row being edited. In tables, the `dotted` border-style behaves in a way that makes it uneven when adding it to a row consisting of several table-cells. 

A solution i found was to create a psuedo-element to the first cell of the current row being edited and add the border to that instead. In order to align the border correctly i had to remove the padding from the cell itself. Then the content would be placed all the way to the left, so i added a span inside each cell around its content and added the same padding that was removed from the cell itself to the span of the first cell in the row currently being edited.

I could definitely try to create a specific class for adding the padding and add it to the `containter-span` for the cells.

Any other improvements to this fix or alternative solutions are very welcome 😄

olemartinorg (approved):
Looks fantastic! :rocket:

src/altinn-app-frontend/src/features/form/containers/RepeatingGroupTable.tsx
olemartinorg: Ahhh, I see! :exploding_head: I inspected this in devtools now, and I absolutely get it.. Clever fix, and I'm happy to sign off on it! Adding code to inject a class here (and not just do it in CSS) might just make it harder to maintain, so I vote for just keeping it as-is.

src/altinn-app-frontend/src/features/form/containers/RepeatingGroupTable.tsx
```diff
@@ -239,6 +278,17 @@ export function RepeatingGroupTable({
     }
   };
 
+  const onClickSave = () => {
```
haakemon, line 281: I would prefer to name this `handleSaveClick` instead, as its handling the event. That makes it easy to spot where its used, if the event is just passed further down, for some other component to handle.

Related blog post https://javascript.plainenglish.io/handy-naming-conventions-for-event-handler-functions-props-in-react-fc1cbb791364

```suggestion
  const handleSaveClick = () => {
```

src/altinn-app-frontend/src/features/form/containers/RepeatingGroupTable.tsx
```diff
@@ -451,22 +452,24 @@ export function RepeatingGroupTable({
...
+                                  ? getFormDataForComponent(component, index)
+                                  : null}
                               </span>
                             </TableCell>
                           );
                         })}
                         <TableCell
                           align='left'
-                          style={{ padding: 0 }}
+                          style={{ width: '110px', padding: 0 }}
```
olemartinorg, line 464: I haven't tested, but how does this affect different labels/languages for the button text? See #469 for an example. Not sure if a fixed width improves anything here.

Magnusrm: The reason i added the hardcoded width in this instance was to make the cells for the buttons behave the same way when the table-header is not displayed. The width is inherited from the column-header and needed to be set on the cells in order to retain the width when the header is not displayed. This will unfortunately not improve the issue in #469.

olemartinorg: Ah - I get it. The way I see it, this problem has been documented well enough in #469 then.

Just a quick comment, though - could `min-width` (or `minWidth`) work better here instead of a fixed `width`? That way it can theoretically scale to become wider, at least.

Issue #452, opened 2022-09-09
Race condition leading to text resources not loading


olemartinorg:
### Description of the bug

[Originally reported on Slack](https://altinn.slack.com/archives/C02EVE4RU82/p1662625754077169?thread_ts=1662552278.562939&cid=C02EVE4RU82). On rare occasions, possibly in this particular app only, the text resources fail to load for some reason, leaving the user without proper texts.

![bug-no-text-resources](https://user-images.githubusercontent.com/700139/189286455-669c43ef-f6d8-4257-974e-b6a83e5b380a.png) 

### Steps To Reproduce

1. Open the app
2. Observe raw text resource identifiers instead of resolved ones

olemartinorg:
The fix for this is now in production. Calling out @StianVestli to report back if he notices the problem again.

Issue #392, opened 2022-08-17
Expose redux devtools in production


olemartinorg:
### Description
When making apps, it could be helpful to view the source redux state in the frontend while manipulating input fields in your app (and observing the effects in the `formData.formData` state). Exposing the redux devtools in production builds [shouldn't be problematic](https://medium.com/@zalmoxis/using-redux-devtools-in-production-4c5b56c5600f), and in fact, some big players seem to be doing it.

### Related issues
- Altinn/altinn-studio/issues/8369

### Tasks
- [x] Add documentation for this here: https://docs.altinn.studio/app/testing/local/debug/
  - Altinn/altinn-studio-docs/pull/665

PR #465, opened 2022-09-13
Enabling redux devtools in production

olemartinorg:
## Description
This change enables redux devtools in the production build, which brings with a few hundred bytes extra code, ~~as well as a 7.1mb `altinn-app-frontend.js.map` and a 556kb `altinn-app-frontend.css.map` to be published on altinncdn.no.~~

Adding these to the production build gives us a few advantages when debugging and reproducing strange bugs:
- We can record the redux state history, saving them to a file, making it possible to save and replay the state history for race condition bugs
- ~~Thrown exceptions in our own code will include references to the actual source files, lines and code that caused the error (instead of line 1, column 489984)~~

## Related Issue(s)
- #392
- #452

## Verification
- [x] All tests run green

## Documentation

### DIFF
--- a/src/altinn-app-frontend/src/store/index.ts
+++ b/src/altinn-app-frontend/src/store/index.ts
@@ -22,7 +22,7 @@ export const setupStore = (preloadedState?: PreloadedState<RootState>) => {
 
   const innerStore = configureStore({
     reducer: reducers,
-    devTools: isDev,
+    devTools: true,
     middleware: (getDefaultMiddleware) =>
       getDefaultMiddleware({
         serializableCheck: isDev,
--- a/src/altinn-app-frontend/webpack.config.production.js
+++ b/src/altinn-app-frontend/webpack.config.production.js
@@ -5,7 +5,7 @@ const common = require('./webpack.common');
 module.exports = {
   ...common,
   mode: 'production',
-  devtool: 'source-map',
+  devtool: false,
   performance: {
     hints: false,
   },
--- a/src/altinn-app-frontend/webpack.config.production.js
+++ b/src/altinn-app-frontend/webpack.config.production.js
@@ -5,7 +5,7 @@ const common = require('./webpack.common');
 module.exports = {
   ...common,
   mode: 'production',
-  devtool: false,
+  devtool: 'source-map',
   performance: {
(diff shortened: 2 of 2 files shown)


olemartinorg:
I spoke with @lorang92 and @altinnadmin about this, and we decided against the 7mb source map this time around. Having it checked into the git repo for cdn, hosted on github spaces, etc, would quickly balloon into much more disk usage than just 7mb.

We also considered:
- Hosting the source maps elsewhere
- Only copying over the source map for the major version to CDN (skipping source maps inside every minor/patch folder)
- Gzipping (1.8mb) or compressing with brotli (1.4mb), however it seems devtools are not ready for a `.js.map.gz` or `.js.map.br` yet - it didn't load correctly (might have been an issue with the wrong HTTP response headers being generated by my static file web server).

TL;DR: Skipping source maps for now

src/altinn-app-frontend/src/store/index.ts
```diff
@@ -22,7 +22,7 @@ export const setupStore = (preloadedState?: PreloadedState<RootState>) => {
 
   const innerStore = configureStore({
     reducer: reducers,
-    devTools: isDev,
+    devTools: true,
```
lorang92, line 25: Just a note: this [defaults](https://redux-toolkit.js.org/api/configureStore#devtools) to `true`, so strictly not needed config. But better to be explicit I guess! 

Neat 🥇

lorang92 (approved):

src/altinn-app-frontend/src/store/index.ts
olemartinorg: Oh, didn't know that! But yeah, I don't mind that being explicit.

olemartinorg:
@ivarne I agree they would be useful, especially for more experienced developers, but as of now they might be problematic to add to the CDN. As the CDN is backed by a git repo, every new file needs to be checked into git, grows the history, and might affect our ability to serve the CDN at all (github static pages have storage limitations). I'm hoping for this to easier in the future, possibly after moving the [CDN to azure](/Altinn/altinn-studio/issues/7566). Right now, the risks outweighs the benefits, IMO.

Issue #473, opened 2022-09-15
Support Right to left languages (Arabic)


lasseklovstad:
### Description

At UDIR (Elevundersøkelsen) we need support for right to left translations.

### Additional Information

1. Add new optional property in translation file `languageDirection: "rtl" | "ltr"` (defaults to ltr)
2.  In app-frontend-react changes the layout direction accordingly to the above new property.
3. Arrow icons must be flipped
4.  Because Altinn doesn't support arabic the page will contain two languages. The top html element will be the altinn language. The Form inside `<main/>` will be the right to left language.

LTR view:
![MicrosoftTeams-image](https://user-images.githubusercontent.com/43168787/190333329-ca1b6a97-c8ef-4110-994f-e8cf05c06403.png)

RTL view:
![MicrosoftTeams-image (1)](https://user-images.githubusercontent.com/43168787/190333365-75a28c51-1529-48a7-9a69-f3c0674876e9.png)

Issue #378, opened 2022-08-12
Change text resources to have a flat key/value structure


haakemon:
### Description

Our current structure is a bit verbose, and also doesn't support reuse of variables:

# Current structure:

```json5
// App/config/texts/resources.nb.json
{
  "language": "nb",
  "resources": [
  {
    "id": "optionsFromRepeatingGroup",
    "value": "Endre fra: {0}, til: {1}",
    "variables": [
      {
        "key": "Endringsmelding-grp-9786.OversiktOverEndringene-grp-9788[{0}].SkattemeldingEndringEtterFristOpprinneligBelop-datadef-37131.value",
        "dataSource": "dataModel.nested-group"
      },
      {
        "key": "Endringsmelding-grp-9786.OversiktOverEndringene-grp-9788[{0}].SkattemeldingEndringEtterFristNyttBelop-datadef-37132.value",
        "dataSource": "dataModel.nested-group"
      }
    ]
  },
  {
    "id": "confirm.body",
    "value": "Vennligst vent, endrer navn til {0}. Du har startet instance: {1} for app: {2}. Trykk [her]({3}) for å starte på nytt.",
    "variables": [
      {
        "key": "Innledning-grp-9309.NavneendringenGjelderFor-grp-9310.SubjektFornavnFolkeregistrert-datadef-34730.value",
        "dataSource": "dataModel.ServiceModel-test"
      },
      { "key": "instanceId", "dataSource": "instanceContext" },
      { "key": "appId", "dataSource": "instanceContext" },
      { "key": "appUrl", "dataSource": "applicationSettings" }
    ]
  }
 ]
}
```

# New structure:
```json5
// App/config/texts/nb.texts.json
{
  "optionsFromRepeatingGroup": "Endre fra: ${{dmv.nested-group.oldName}}, til: ${{datamodelvariables.nested-group.newName}}",
  "confirm.body":  "Vennligst vent, endrer navn til ${{dmv.ServiceModel-test.updatedName}}. Du har startet instance: ${{ic.instanceId}} for app: ${{instanceContext.appId}}. Trykk [her](${{as.appUrl}}) for å starte på nytt.",
}
```
- `[as|applicationSettiings].<variable> `points to the **applicationSettings** dataSource in current structure
- `[ic|instanceContext].<variable>` points to the **instanceContext** dataSource in current structure
[...]

PR #474, opened 2022-09-15
Support for right to left language

lasseklovstad:
## Description
Support for right to left languages. Supported languages:

```
 'ar', // Arabic
  'arc', // Aramaic
  'dv', // Divehi
  'fa', // Persian
  'ha', // Hausa
  'he', // Hebrew
  'khw', // Khowar
  'ks', // Kashmiri
  'ku', // Kurdish
  'ps', // Pashto
  'ur', // Urdu
  'yi', // Yiddish
```
Ref: https://lingohub.com/academy/best-practices/rtl-language-list

## Manual tests:
I have tested some, but not all components. All components used by Elevundersøkelsen is tested. Other apps should not be impacted by this upgrade. Some screen shots from tests (Some translation are not added):

Left to right example 1:
image

Right to left example 1:
image

Left to right example 2:
image

Right to left example 2:
image

## Related Issue(s)
- #473 

## Verification
- [ ] All tests run green

## Documentation


src/altinn-app-frontend/package.json
```diff
@@ -30,6 +30,7 @@
     "axios": "0.27.2",
     "dot-object": "2.1.4",
     "jsonpointer": "5.0.1",
+    "jss-rtl": "0.3.0",
```
haakemon, line 33: I'm not too sure about introduction of this dependency. It seems to be a relatively little used project (~80 stars on github, not updated in 3 years, ~50k weekly downloads on npm). We are also in general moving away from css-in-js.

Do you have examples of any specific rules that broke without this plugin? I think this is also valuable info for @Febakke and @mrosvik moving forward, that we probably need to also think about RTL languages when making our designs.

haakemon:
The proposed extra metadata in the `resources.xx.json` file is a bit unfortunate. We are intending to make a change to the structure of this file, so that the end result is a flat key/value structure with the content of the `resources` array.

```json
{
  "language": "ar",
  "languageDirection": "rtl",
  "id": "udir-elevundersokelsen-ar",
  "org": "udir",
  "resources": []
}
```

We already should know the language code, so from that info we should be able to tell if the language is RTL or LTR, unless I have missed something. I dont really have any experience with RTL languages, so that may very well be the case 😅

haakemon:
As a slightly unrelated sidenote/observation/question, I cannot remember having seen the `id` or `org` fields in a resource file before - what are those fields used for?

src/altinn-app-frontend/package.json
lasseklovstad: It was the recommended way to handle rtl in mui v4: https://v4.mui.com/guides/right-to-left/#right-to-left

lasseklovstad:
> The proposed extra metadata in the `resources.xx.json` file is a bit unfortunate. [We are intending to make a change to the structure of this file](https://github.com/Altinn/app-frontend-react/issues/378), so that the end result is a flat key/value structure with the content of the `resources` array.
> 
> ```json
> {
>   "language": "ar",
>   "languageDirection": "rtl",
>   "id": "udir-elevundersokelsen-ar",
>   "org": "udir",
>   "resources": []
> }
> ```
> 
> We already should know the language code, so from that info we should be able to tell if the language is RTL or LTR, unless I have missed something. I dont really have any experience with RTL languages, so that may very well be the case 😅

Da kan jeg endre til at frontend finner ut om det rtl basert på en liste med predefinert liste av språk. Fant denne listen online: https://lingohub.com/academy/best-practices/rtl-language-list

lasseklovstad:
> As a slightly unrelated sidenote/observation/question, I cannot remember having seen the `id` or `org` fields in a resource file before - what are those fields used for?

This could be some extra fields we by mistake have added 😅

src/altinn-app-frontend/package.json
lasseklovstad: The plugin will flip for example left padding to right padding in all mui components. It will basically flip everything mui related. And since it is alot of mui stuff here it was the easiest solution currently available.

lasseklovstad:
@haakemon I am not sure why the cypress tests are failing and i don't have the cypress test app locally to debug it (where can i clone the test app?). I see that the formatting test is testing text-align property, maybe that is an issue 😟

src/altinn-app-frontend/package.json
haakemon: I'm just wondering if this is nescessary at all? We usually have equal padding on right and left sides, so there really shouldnt be much need to swap them. There might be edge cases here though, thats why I would also love some input from our UX group to see if they know more.

I did some quick checks with just enabling 

```css
direction: rtl;
text-align: right;
```
and from what I can tell it looks fine with just this? But I havent been able to really dig into this and look at everything either, so there may be things you have noticed, that doesnt work with just these settings.

haakemon:
@lasseklovstad regarding the cypress tests, its located here `https://github.com/Altinn/app-frontend-react/tree/main/test/cypress`, and by following the readme you should hopefully get it up and running. If not, let me know what you get stuck on and we can improve the docs together 🙂

Looking at the summary of the cypress testrun, it seems to be failing when checking formatting of numbers, so hopefully that may give you a clue: https://github.com/Altinn/app-frontend-react/actions/runs/3061478165/jobs/4941293976#step:7:458

If you get totally stuck, just let me know and I'll try and help out

src/altinn-app-frontend/package.json
lasseklovstad: When removing this library the form looks like this. As you can see there are three problems. Right align text in table, Extra margin on right sides og RadioButtons(Will also apply for Checkboxes), Margin between NavigationButtons.
image

lasseklovstad: I have pushed a new commit where I remove the library and make the margin between Navigation and table cell right align text work without this library. The right margin on the Radiobutton isn't that important to fix imo. SO maybe we can live with the extra margin on the right side until these components are implemnted in the design system?

lasseklovstad:
@haakemon Da har jeg refakturert såpass bra at cypress testen fikk lyst til å fungere uten at jeg trengte å gjøre noe 😄 

Det er som nevnt noe margin og padding som ikke reverseres, men er ikke krise i første omgang (tenker jeg). Så lenge man husker på å kunne støtte RTL så vil dette ordne seg med nytt design system og mui blir fjernet i fremtiden.

haakemon (approved):
Looks great! 🥳

Issue #316, opened 2022-07-04
WCAG issues with the likert component


Febakke:
### Description of the bug

There are a couple of WCAG issues with the "likert component" that should be fixed. 

## 1. Title element is a h5. 
We should avoid headers in our components, and if we need them they should probably be h2 by default and possible to edit by the app developer. As it is now it will in most cases cause the app to fail the [1.3.1. info and relationships](https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships) by jumping over some elements in the header order.
[Example of how header structure should be (Norwegian)](https://www.uutilsynet.no/veiledning/tekst-og-struktur/226#overskriftsniver)

![Skjermbilde 2022-07-04 kl  15 36 10](https://user-images.githubusercontent.com/33222679/177170525-f52d1350-6fa1-47b5-93cd-178eb7e24c4e.png)

## 2. Empty `<th>` element
There is an empty header element in the table. This is breaking [1.3.1. info and relationships](https://www.w3.org/WAI/WCAG21/Understanding/info-and-relationships)
![Skjermbilde 2022-07-04 kl  15 37 42](https://user-images.githubusercontent.com/33222679/177171317-fe6dfce5-5f42-45c9-9167-2b23fe3cf7d0.png)

## 3. Empty labels
The testing tool Wave is giving a warning that the component contain labels, but that they do noy have any content. I tested it with a screen reader and every bit of information in relation to the radio button was read. (Tested with VoiceOver) Maybe look into if the labels should be structured in another way? 

One weird bug was that the label was read twice by Voice Over
![Skjermbilde 2022-07-04 kl  15 25 19](https://user-images.githubusercontent.com/33222679/177172044-c5b3d176-93ec-40b1-985e-b6be0613af26.png)

### Steps To Reproduce

Go to an app with a likert component
Run wave or another testing tool

Magnusrm:
Could be looked at together with #507

mijohansen:
This issue is partly unsolveable with the current use of material-UI which  doesn't support a two-header setup that the likert component needs. The problem with empty headers is probably solveable as we could just provide a text there to indicate that the column underneath is a column of questions.

https://www.w3.org/WAI/tutorials/tables/two-headers/

Issue #217, opened 2022-06-08
Summary view for Likert component (in group) shows option value instead of display value/label


nkylstad:
## Description
Bug from likert component implementation (#72)
Summary view  ([Summary](https://docs.altinn.studio/app/development/ux/pages/summary/)) for Likert is rendered, but the _value_ of the selected option is shown and not the _display value/label_. I believe we have implemented this for regular radio buttons (so that the label of the option is shown) making the summary view more readable. This should be the case for Likert as well.

## To Reproduce
1. Go to an app with likert component (w/group) and summary for likert component.
2. Select some values in likert component
3. See that summary component is updated with the selected values, but showing the stored value of the selected option rather than the display value/label.

## Expected behavior
Should render the display value (label) of the selected option rather than the stored value.

## Screenshots
![image](https://user-images.githubusercontent.com/1636323/172343585-5f0e37f4-1771-4454-9750-aa47050469e1.png)

## Additional info
Check out the SummaryComponent.tsx and see how this has been solved for f.ex. regular RadioButtons.
Tested in Chrome.

nkylstad:
@lasseklovstad Somehow we completely missed another bug here, which is that the question text does not appear in the summary. `{0}` should be replaced by the relevant question text for the row, but it seems like variables in text is not working here. Could you take a look?

lasseklovstad:
@nkylstad It takes some time for me to test this so I won't have time right now. But I have done some investigating and it looks like this line 118 in `SummaryComponent` should change the implementation to what `GenericComponent` is doing. As you can see the `selectComponentTexts` has a third param `tryNesting` which is key for the Likert component. 

SummaryComponent (118):
```
  const label = useAppSelector((state) => {
    const titleKey = formComponent.textResourceBindings?.title;
    if (titleKey) {
      return getTextFromAppOrDefault(
        titleKey,
        state.textResources.resources,
        state.language.language,
        [],
        false,
      );
    }
    return undefined;
  });
```

GenericComponent (128):
```
  const texts = useAppSelector((state) =>
    selectComponentTexts(
      state.textResources.resources,
      props.textResourceBindings,
      props.type === 'Likert',
    ),
  );
```

lasseklovstad:
It is a bit confusing cause it is multiple implementations of getting texts from text resources. It should maybe be done a cleanup around this.

JohanGuntherKrogevoll:
[...]

PR #478, opened 2022-09-20
Likert bugs

lasseklovstad:
Fix Summary component not displaying likert values correct.
Fix validation message not displaying correct for likert.
Fix summary component when using filter property on repeating group.
Fix Wcag issue with empty column header. (Added textResourcebinding)

## Related Issue(s)
- #316
- #217

## Verification
- [x] All tests run green

## Documentation


olemartinorg:
Thanks for the contribution! :partying_face: As I mentioned in #479, maybe the cypress test could be expanded to assert the values shown/selected in the summary component here? :blush:

src/shared/src/utils/language.ts
```diff
@@ -108,9 +108,18 @@ export function getTextResourceByKey(
...
   );
+
+  if (textResource) {
+    // Checks if this text resource is a reference to another text resource.
+    // This is a common case when using likert component
+    textResource =
+      textResources.find((resource) => resource.id === textResource.value) ||
+      textResource;
+  }
   return textResource ? textResource.value : key;
```
mjulstein, line 123:
```suggestion
const textResource = textResources.find(...);
if (!textResource){ return key; }

// Checks if this text resource is a reference to another text resource.
// This is a common case when using likert component
const resource = textResources.find((resource) => resource.id === textResource.value) || textResource;
return resource.value;  
```

test/cypress/e2e/fixtures/texts.json
```diff
@@ -19,7 +19,7 @@
   "requiredFieldLastName": "Du må fylle ut nytt etternavn",
   "requiredFieldDateFrom": "Du må fylle ut dato for navneendring",
   "requiredFieldFromValue": "Du må fylle ut 1. endre fra",
-  "requiredFieldToValue": "Du må fylle ut 2. endre verdi {0} til ",
+  "requiredFieldToValue": "Du må fylle ut 2. endre verdi  til ",
```
mjulstein, line 22: Shouldn't this actually render some text in this test?
This would mean that the field is not working properly.
I agree that `{0}` doesn't make much sense either, but this looks like the substitution is not working properly.

lasseklovstad: Here is a screenshot of the test if I fill inn the first textfield and click save: 
image

lasseklovstad: I guess the value of the first textfield should actually be in the error message.

lasseklovstad: Yes I can see that the errormessage is correctly updated after clicking "Lagre og lukk". I Can change the test so it will reflect this.

src/altinn-app-frontend/src/features/form/containers/GroupContainerLikert.test.tsx
```diff
@@ -24,6 +24,22 @@ describe('GroupContainer', () => {
...
+            title: 'Test title',
+            description: 'Test description',
+            leftColumnHeader: 'Test left column header',
+          },
+        },
+      });
+      screen.getByText('Test title');
+      screen.getByRole('table', { name: 'Test title' });
+      screen.getByText('Test description');
+      screen.getByRole('columnheader', { name: 'Test left column header' });
```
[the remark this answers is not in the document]
lasseklovstad: getByRole will throw an error if it is not found so I don't see the point in adding an extra assertion to something that is allready found in th document? Or is it something I am missing here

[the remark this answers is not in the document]
haakemon: The advantage is that the purpose of the code will be more explicit, instead of implicit, and we could have avoided this discussion altogether 🙂 

Future potential errors could also be avoided, if f.ex the `getByRole` is changed (or if the implementation is changed to f.ex `queryByRole`, which does not throw when no items are found) to not throw an error. If we dont add the expects, the test will pass even if the elements are no longer found.

haakemon (approved):
LGTM 🙂

src/altinn-app-frontend/src/features/form/containers/GroupContainerLikert.test.tsx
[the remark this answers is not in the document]
haakemon: Did a bit digging on this, and seems like there are several others that prefer your approach 🙂, and there even is a [eslint rule for it](https://github.com/testing-library/eslint-plugin-testing-library/blob/main/docs/rules/prefer-explicit-assert.md). Personally, I would prefer sticking with `expect` though.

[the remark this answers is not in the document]
haakemon: [Using `expect` is also recommended](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-get-variants-as-assertions) by Kent C. Dodds, the original author of testing-library.

PR #480, opened 2022-09-20
Fixing race condition preventing language from loading

olemartinorg:
## Description
After struggling to get this to reproduce, I finally managed it. The following code was the culprit:

```typescript
yield all([
  take(FormLayoutActions.fetchSetsFulfilled),
  take(ApplicationMetadataActions.getFulfilled),
  take(LanguageActions.fetchLanguage),
]);

const allowAnonymous = yield select(makeGetAllowAnonymousSelector());
if (!allowAnonymous) {
  yield take(ProfileActions.fetchFulfilled);
}
```

In order for `makeGetAllowAnonymousSelector` to work, it needs the layout sets and application metadata, so we wait for those first in the `yield all([...])` on top. The profile, however, is already being fetched, so if any of the requests on the top are slow to respond (I tested this by artificially delaying the request for layout sets), we end up hanging on the `take(ProfileActions.fetchFulfilled)`. The profile fetching won't fulfill, as it has been fetched already.

The solution is quite simple; re-use the `waitFor()` effect implemented in [#6611](/Altinn/altinn-studio/pull/6611) to wait until the *effect* of fetching the profile has been observed (i.e. having the profile state). This makes sure to return immediately if the profile is in redux.

## Related Issue(s)
- #452

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/altinn-app-frontend/src/shared/resources/language/fetch/fetchLanguageSagas.test.ts
+++ b/src/altinn-app-frontend/src/shared/resources/language/fetch/fetchLanguageSagas.test.ts
@@ -10,7 +10,7 @@ import {
   watchFetchLanguageSaga,
 } from 'src/shared/resources/language/fetch/fetchLanguageSagas';
 import { LanguageActions } from 'src/shared/resources/language/languageSlice';
-import { ProfileActions } from 'src/shared/resources/profile/profileSlice';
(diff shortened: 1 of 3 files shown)


haakemon (approved):
LGTM 🙂

Issue #432, opened 2022-09-02
Navigate from confirmation step to other steps in a process (BPMN)


mjulstein:
### Description
As a user I need to be able to navigate back to a previous process task when I have reached the confimation task, so that I can fix what blocks me from submitting.

This is a specification of the frontend tasks from Altinn/app-frontend-react#262 
In issue Altinn/app-template-dotnet#129 a new endpoint is discussed.
The gist is that this endpoint can be called on the start of any given processes-task and the backend will respond with tasks that are available to the client at a given time in the process.

The API should return enough information about the process so that a call to `process/next/<requiredInfo:takID?>` will cause the backend to update the process state and the user can be redirected to the root of the App.

### Additional Information

In addition the enabling custom navigation to other processes from a given process we need:
- [ ] Cypress tests to validate that navigating to a given process-task works as intended

A new receipt page is in development by @lorang92 and should work with the overriding button layout.
- <Link to the issue>

### Tasks

In order for this to work we need:
- [ ] Designs on how the overriding button layout should be presented (if the user has access or not) @Febakke  we need to discuss conditional states/access
- [ ] Find a suitable file or create a new one to hold the configuration needed to override the 'submit-button-only' design
- [x] Decide what is actually needed to set up in the new configuration.
   * Is a config actually needed, can we find enough information from the API from Altinn/app-template-dotnet#129 to render additional buttons?
   * If that is the case can we make a unique text-id-prefix for process-next-api buttons that can be localized? WYDT @RonnyB71 ?
   * Can we avoid requiring App developers to edit their configs to utilize the new feature
   * Should we avoid adding this feature unless explicit settings have been applied in the App (the available-next-tasks api could possibly be enough, as it can return 404 if the feature is not available.)
   * Other things

### Acceptance Criterias

- [ ] Buttons are rendered on the receipt-page(s) per the final design
- [ ] The buttons that can be used are functioning as intended: If disabled, you won't change processes etc.
- [ ] Document how to override the current submit-button functionality with this feature
- [ ] Unchanged apps should appear as before
- [ ] When a user is sent to a different process the browser history should be replaced with the page they are currently on while navigating backwards.

PR #488, opened 2022-09-24
Feat/navigate to process from confirm task

mjulstein:
## Description
Adds support to set a `mode` property on the `ButtonComponent` to create a secondary button to switch to an available BPMN task. The mode is called `go-to-task` and requires a `taskId` prop with the name of the task to be set.
Adding the following to formLayout or will render the button:
```json
{
	"id": "Task-1-Button",
	"type": "Button",
	"mode": "go-to-task",
	"textResourceBindings": {
	  "title": "Gå til Task-1"
	},
	"dataModelBindings": {},
	"taskId": "Task_1"
}
```
https://dev.altinn.studio/repos/ttd/frontend-test/src/branch/feat/support-navigation-to-other-route

## Related Issue(s)
- #432 

## Verification
- [ ] All tests run green

## Documentation


src/altinn-app-frontend/webpack.common.js
```diff
@@ -28,11 +28,7 @@ module.exports = {
         },
       },
       {
-        test: /\.scss$/,
-        use: ['style-loader', 'css-loader'],
-      },
-      {
-        test: /\.css$/,
+        test: /\.(sa|sc|c)ss$/,
```
haakemon, line 31: This can be changed back, and we can remove the `scss` test entirely. It requires additional loaders (f.ex sass-loader) which we havent set up. And we're not using sass, so no need to keep it here.

```suggestion
        test: /\.css$/,
```

src/altinn-app-frontend/src/components/base/ButtonComponent/ButtonComponent.tsx
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  const saveFormData = () => {
+    dispatch(FormDataActions.submit({ componentId: 'saveBtn' }));
+  };
+
+  const submitForm = ({ componentId }: { componentId: string }) => {
+    const { org, app, instanceId } = window as Window as IAltinnWindow;
+    dispatch(
+      FormDataActions.submit({
```
nkylstad, line 53: I'm pretty sure sure this won't actually work in navigating from the confirm page to completion (receipt). In the static confirm view, we had a custom submit button, which called `process.next`. But here, in order to display the same "submit" button, the layout must use this button, which submits the _form data_, which is not actually connected to the confirm task. The user will typically not have access to write to the data task from the confirm task, so this will cause a 403 response from the server. 
A quick fix might be to add a check in `submitFormDataSaga` and only PUT form data if the current task is a data task. Thoughts @lorang92 @TheTechArch ?

lorang92: Ah, I see that we got a `GoToTaskButton` which only calls `ProcessActions.complete`. I'm guessing this can be used in sted of the submit-button when we are in receipt going to completion?

lorang92: Can you confirm that this works @tjololo? I'm imagining we tried the whole flow from Task_1 to the end event?

nkylstad: Yep that might work. But the styling won't be the same as the static confirm page where we have the green button.

lorang92: Good point. And I agree we should probably just fix so that submit-button works in confirm :)

tjololo: I have tested with the latest commit from this branch. There are some errors presented when I run `yarn start`
Other than that I have tested the whole process in an application with custom confirm page (and buttons) and an app with the default confirm page. Both seemed to work fine locally.
The "confirm" buttons on the custom confirm page is not styled as a submit button like the one on the static page, but that might my fault.
Error presented from yarn start:
```
ERROR in ./__mocks__/validationStateMock.ts:[32m[1m2:10[22m[39m

TS2459: Module '"../src/utils/textResource"' declares 'getTextResourceByKey' locally, but it is not exported.
    1 | import type { IValidations } from '../src/types';
  > 2 | import { getTextResourceByKey } from '../src/utils/textResource';
      |          ^^^^^^^^^^^^^^^^^^^^
    3 |
    4 | export function getMockValidationState(withFixed = false): IValidations {
    5 |   const fixed = withFixed ? [] : undefined;
 ```
 
 Screenshot of custom confirm page:
 
image

Comment (tjololo) [reply src/altinn-app-frontend/src/components/base/ButtonComponent/ButtonComponent.tsx:53]: from the formlayout of the confirm page:
```json
{
    "id": "submit",
    "type": "Button",
    "mode": "go-to-task",
    "taskId": "EndEvent_1",
    "textResourceBindings": {
      "title": "button.confirm"
    },
    "grid": {
       "xs": 2
     }
 }
```
Errors are gone after the latest changes 🥳

mjulstein: The go-to-task mode is intentionality  secondary by the way.
I thought the button was supposed to be in addition to other buttons on the page. And then the submit and confirm buttons where supposed to work as before.

nkylstad: Yes, that was the idea @mjulstein. But the submit button from the custom confirm page didn't actually work as it saved the form data, as described in my original comment. I pushed a fix for this, so the submit button should be working on the custom confirm page now.

src/altinn-app-frontend/src/utils/appUrlHelper.ts
```diff
@@ -11,67 +11,59 @@ export const profileApiUrl = `${appPath}/api/v1/profile/user`;
...
-  );
-}
+export const fileUploadUrl = (attachmentType: string) =>
+  `${appPath}/instances/${altinnWindow.instanceId}/data?dataType=${attachmentType}`;
 
-export function dataElementUrl(dataGuid: string) {
-  return `${appPath}/instances/${altinnWindow.instanceId}/data/${dataGuid}`;
-}
+export const fileTagUrl = (dataGuid: string) =>
+  `${altinnWindow.instanceId}/data/${dataGuid}/tags`;
```
nkylstad, line 30: Most of the changes in this file feel pointless, and cause a lot of noise so it's almost impossible to see any _actual_ changes. In addition it introduces lots of opportunities for errors, like this case here where part of the uri is missing.

src/altinn-app-frontend/package.json
```diff
@@ -102,7 +103,7 @@
     ],
     "moduleNameMapper": {
       "\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$": "<rootDir>/__mocks__/fileMock.js",
-      "\\.(css|less)$": "<rootDir>/__mocks__/styleMock.js",
+      "\\.(css|less|scss)$": "identity-obj-proxy",
```
lorang92, line 106: If we are no longer using the `styleMock.js` I guess this file can be deleted?

src/altinn-app-frontend/src/utils/appUrlHelper.test.ts
```diff
@@ -1,163 +1,226 @@
...
-        origin: 'https://ttd.apps.altinn.no',
-        hash: '#/datamodelling',
-        pathname: '/ttd/jesttest/',
-        host: 'https://ttd.apps.altinn.no',
-        href: 'https://ttd.apps.altinn.no/ttd/test',
-      };
-      const oidcProvider = 'idporten';
-      expect(getEnvironmentLoginUrl(oidcProvider)).toContain(
-        'https://platform.altinn.no/authentication/api/v1/authentication?goto=https%3A%2F%2Fttd.apps.altinn.no%2Fttd%2Ftest&iss=idporten',
+  window['instanceId'] = 'instance/id';
```
nkylstad, line 28: I find this a little confusing, its challenging to see if this is part of the base url or if its a variable with `instance/id`. Would prefer to use something explicit to make it more obvious that this is a variable part of the url.

nkylstad (commented):
Noticed a few things:
- Spacing around the submit button on the (default) confirm page is too small, the button is too close to the texts.
- The submit button on the "default" static confirm page was blue (primary button), is now green (submit). 
@Febakke any input?
Screenshot 2022-09-29 at 10 34 55

Screenshot 2022-09-29 at 10 35 20

UPDATE: Confirmed with @Febakke that we will keep the button green for now. Added back the missing spacing.

src/altinn-app-frontend/src/components/base/ButtonComponent/InstantiationButton.tsx
```diff
@@ -2,40 +2,23 @@ import * as React from 'react';
...
-  marginLeft: '40px',
-  marginTop: '2px',
-  height: '45px',
-};
-
 export type IInstantiationButtonProps = IComponentProps &
   Omit<ILayoutCompInstantiationButton, 'type'>;
 
-export function InstantiationButtonComponent(props: IInstantiationButtonProps) {
+export type Props = Omit<ButtonProps, 'onClick'> &
```
nkylstad, line 18: We should use a more descriptive name here.

src/altinn-app-frontend/src/components/base/ButtonComponent/ButtonComponent.tsx
```diff
@@ -0,0 +1,96 @@
...
+          <div className={css['button-row']}>
+            <GenericButton {...props}>{props.text}</GenericButton>
+          </div>
+        </div>
+      </div>
+    );
+  }
+
+  const saveFormData = () => {
+    dispatch(FormDataActions.submit({ componentId: 'saveBtn' }));
```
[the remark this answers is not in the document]
nkylstad: Never mind, I misunderstood the "save" button in this context.

nkylstad (approved):
I've made a few small updates, great if @lorang92 could take a quick look as well :)

lorang92 (approved):
@nkylstad there is a lot of stuff going on here, but I have glanced at this. I think this is fine for now, and don't find anything I consider blocking. We also have cypress tests that should cover regression testing for the changed files rather well (`data` => `data` => `data` => `confirm` static => `end`), so I'm pretty confident that existing features should work as expected. 

I would like to have some cypress tests for a gateway flow introduced as this is hard to unit-test. Not blocking, but if you have some spare time that would be nice.

nkylstad:
> @nkylstad there is a lot of stuff going on here, but I have glanced at this. I think this is fine for now, and don't find anything I consider blocking. We also have cypress tests that should cover regression testing for the changed files rather well (`data` => `data` => `data` => `confirm` static => `end`), so I'm pretty confident that existing features should work as expected.
> 
> I would like to have some cypress tests for a gateway flow introduced as this is hard to unit-test. Not blocking, but if you have some spare time that would be nice.

Yep, I agree that we should have some explicit cypress tests that handle the concept of going back and forth between tasks. Will try to add some.

Issue #461, opened 2022-09-12
Readable summary of data 


lorang92:
### Description

For BRG we have a need to give a readable summary of data in the confirm step. 

![image](https://user-images.githubusercontent.com/10416981/189592488-1b8e138d-6ccc-44cf-acb5-c61cecfd6f60.png)

### Additional Information

How does this relate to our summary component? What is different about this need? Should these needs be combined in some way?

### Acceptance Criterias

- Can be used for repeating structures
- Can be used as a standalone component

PR #489, opened 2022-09-26
add display options to summary components

nkylstad:
## Description
Updated summary components with some display options to make it more readable with repeating groups:
- `hideBottomBorder` - to hide the blue dotted border between summary components.
- `useComponentGrid` - allows the summary component to use the layout components defined grid setup. Especially useful for summary groups with `largeGroup: true`.
- `hideChangeButton` - Useful when using summary components f.ex. in a custom confirm view, when it is not possible to navigate directly back to the form to change things.

In addition, updated some styling as discussed with @Febakke:
- Added a pale grey border around repeating group items in summary to make it more readable
- Swapped styling of label/data so that data is bold while label is normal.

Screenshot 2022-09-26 at 13 17 02

Apart from the small styling fixes, existing apps using summary components should not be affected.

## Related Issue(s)
- #461

## Verification
- [x] All tests run green

## Documentation

### DIFF
--- a/src/altinn-app-frontend/src/components/summary/SummaryBoilerplate.tsx
+++ b/src/altinn-app-frontend/src/components/summary/SummaryBoilerplate.tsx
@@ -8,7 +8,8 @@ import type { ILayoutCompSummary } from 'src/features/form/layout';
 
 import appTheme from 'altinn-shared/theme/altinnAppTheme';
 
-export interface SummaryBoilerplateProps extends ILayoutCompSummary {
+export interface SummaryBoilerplateProps
+  extends Omit<ILayoutCompSummary, 'type' | 'id'> {
   hasValidationMessages?: boolean;
   onChangeClick: () => void;
   changeText: string;
--- a/src/altinn-app-frontend/src/components/summary/SummaryBoilerplate.tsx
+++ b/src/altinn-app-frontend/src/components/summary/SummaryBoilerplate.tsx
@@ -43,7 +43,7 @@ export default function SummaryBoilerplate({
 }: SummaryBoilerplateProps) {
   const classes = useStyles();
   const shouldShowChangeButton =
-    !readOnlyComponent && !display.hideChangeButton;
+    !readOnlyComponent && !display?.hideChangeButton;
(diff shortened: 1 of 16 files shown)


src/altinn-app-frontend/src/components/summary/SummaryBoilerplate.tsx
```diff
@@ -13,6 +14,7 @@ export interface SummaryBoilerplateProps {
   changeText: string;
   label: any;
   readOnlyComponent?: boolean;
+  display?: SummaryDisplayProperties;
```
olemartinorg, line 17: This should probably be added to `ILayoutCompSummary` instead, and this should extend from that interface. That would give you intellisense on `component.display...` after checking `if (component.type === 'Summary')` :smiley:

src/altinn-app-frontend/src/components/summary/SummaryComponent.tsx
```diff
@@ -14,13 +15,18 @@ import {
...
+  ILayoutComponent,
+  SummaryDisplayProperties,
+} from 'src/features/form/layout';
 import type { IComponentValidations, IRuntimeState } from 'src/types';
 
 export interface ISummaryComponent {
   id: string;
   componentRef?: string;
   pageRef?: string;
+  display?: SummaryDisplayProperties;
```
olemartinorg, line 29: Same here. There are probably also some other options here not yet documented in the layout jsonschema..? :thinking:

olemartinorg (changes requested):
:godmode:
I'll put up a 'request changes' because of the failing unit tests, otherwise this LGTM!

src/altinn-app-frontend/src/components/summary/SummaryBoilerplate.tsx
nkylstad: Ah, great tip! Was able to get rid of several props that were already defined in `ILayoutCompSummary`. Will update jsonschema and make sure that all props are defined there.

olemartinorg (approved):

PR #492, opened 2022-09-27
Docker compose

ivarne:
## Description
If you can't/won't install node on your computer, you can also run frontend in docker.

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,14 @@ React SPA used by applications developed in [altinn-studio](https://github.com/A
 
 ## Prerequisites
 
+### Docker
+If you can't/won't install node on your computer, you can also run frontend in docker using the command.
+
+```bash
+docker compose up
+```
+This is really slow to start and rebuild, but sometimes better than getting someone to install node if you just want to test if a new branch fixes an issue.
+
 ### Node and Corepack
 
 - Latest [Node LTS release](https://nodejs.org/en/)
@@ -24,11 +32,20 @@ git clone https://github.com/Altinn/app-frontend-react
 cd app-frontend-react
 ```
 
+The development server can be started by following these steps:
+
+- Navigate to `./src/altinn-app-frontend`
+- `yarn --immutable` (only needed when `package.json` has changed)
+- `yarn start` (to start the development server)
+
 ### Developing app-frontend
 
-You need an Altinn app to effectively make changes to the app-frontend codebase. To serve the development version of app-frontend code, you need to make some changes to `views/Home/Index.cshtml` in the app repo you are using:
+You need an Altinn app to effectively make changes to the app-frontend codebase.
 
-Change
+Localtest now includes a way to avoid editing your app to use different frontends [see docs](https://docs.altinn.studio/app/testing/local/debug/#using-other-frontend-versions)
+
+
+If you want to test in `tt02` or other places, you need to make some changes to `views/Home/Index.cshtml` in the app repo you are using:
 
 ```html
 <link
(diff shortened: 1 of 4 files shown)


src/altinn-app-frontend/webpack.config.docker.js
```diff
@@ -0,0 +1,13 @@
...
+
+module.exports = {
+  ...common,
+  devServer: {
+    ...common.devServer,
+    host: '0.0.0.0', // docker needs to bind to public interface in container
+  },
+  watchOptions: {
+    ignored: ['**/node_modules', '**/.git', '**/dist'],
+    poll: 60000,
```
olemartinorg, line 11: Wait, so this polls files every 60s instead of watching files? :thinking: Is this needed because of some side-effect with Windows as the host?

README.md
```diff
@@ -24,11 +32,21 @@ git clone https://github.com/Altinn/app-frontend-react
...
+- `yarn --immutable` (only needed when `package.json` has changed)
+- `yarn start` (to start the development server)
+
 ### Developing app-frontend
 
-You need an Altinn app to effectively make changes to the app-frontend codebase. To serve the development version of app-frontend code, you need to make some changes to `views/Home/Index.cshtml` in the app repo you are using:
+You need an Altinn app to effectively make changes to the app-frontend codebase.
 
-Change
+Localtest now includes a way to avoid editing your app to use different frontends
```
olemartinorg, line 45: Instead of adding an image to the repo and describe this separately here, we could add a link to the docs instead: https://docs.altinn.studio/app/testing/local/debug/#using-other-frontend-versions

docker-compose.yml
```diff
@@ -0,0 +1,12 @@
+version: '3.6'
+
+services:
+  webpack:
+    container_name: name
+    image: node:16
+    volumes:
+      - ./:/usr/src/app
+    working_dir: /usr/src/app/src/altinn-app-frontend
+    command: bash -c "yarn --immutable && yarn docker"
```
olemartinorg, line 10: Could we instead create a Dockerfile extending from node:16 and install deps/trigger rebuilds when the yarn.lock changes? I love the suggestion in this PR, but I'd like for it to become a first-class citizen instead of something we describe as 'really slow'.. ;-) If you don't want to spend time doing it, I could make these changes after merging this as well.

olemartinorg (changes requested):
Fantastic suggestion! :partying_face: Some minor comments I'd like you to look over, so I'm marking it as 'changes requested' to keep it in my loop. Thanks a bunch!

src/altinn-app-frontend/webpack.config.docker.js
ivarne: Yes, cross platform file watchers was not recommended. I didn’t test.

README.md
ivarne: 👍

docker-compose.yml
ivarne: I haven’t (yet) found a great way to do local development in docker. This approach ensures that we always use the latest code on disk, and caches npm packages. A better way would be great!, but my current problem is that we have restrictions on installing software, and when we have docker, there is an obvious solution.

README.md
ivarne: Finally had some time in front of the computer.

olemartinorg (approved):
:rocket:

olemartinorg:
Whoops.. I just now noticed this PR was targeted towards the layout expressions feature branch(!) I'll pull it back out of there.

ivarne:
Oops indeed!

It was because I needed this to test the layout expressions, but the PR should obviously have been against “main”

PR #497, opened 2022-09-28
Cleaning up some unit test warnings in console

olemartinorg:
## Description
We had some unit tests that were producing warnings in the console, which makes it difficult to see which tests are failing and why (especially in non-color consoles, like sometimes here on github).

## Fixed
- Removing leftover `screen.debug()`
- Adding missing keys to validation messages (which React complains about). I struggled to find something unique, so serializing all the arguments should be unique enough. 

## Verification
- [x] All tests run green

## Documentation

### DIFF
--- a/src/altinn-app-frontend/src/utils/render.tsx
+++ b/src/altinn-app-frontend/src/utils/render.tsx
@@ -61,13 +61,11 @@ export function renderValidationMessages(
   id: string,
   variant: 'error' | 'warning' | 'info' | 'success',
 ) {
-  const key = JSON.stringify({ messages, id, variant });
-
   if (variant !== 'error') {
     return (
       <SoftValidations
         variant={variant}
-        key={key}
+        key={id}
       >
         <ol id={id}>{messages.map(validationMessagesToList)}</ol>
       </SoftValidations>
@@ -77,7 +75,7 @@ export function renderValidationMessages(
   return (
     <div
       style={{ paddingTop: '0.375rem' }}
-      key={key}
+      key={id}
     >
       <ErrorMessage id={id}>
         <ol>{messages.map(validationMessagesToList)}</ol>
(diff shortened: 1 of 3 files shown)


src/altinn-app-frontend/src/utils/render.tsx
```diff
@@ -61,16 +61,24 @@ export function renderValidationMessages(
...
   variant: 'error' | 'warning' | 'info' | 'success',
 ) {
+  const key = JSON.stringify({ messages, id, variant });
+
   if (variant !== 'error') {
     return (
-      <SoftValidations variant={variant}>
+      <SoftValidations
+        variant={variant}
+        key={key}
```
haakemon, line 70: This seems odd to me - there is no looping over values here, so why should a key be nescessary?

I think the underlying problem here is the `renderValidationMessagesForComponent` code that returns an array of items directly, and that is returned, and used as a component (rendered). 

Shouldnt this be fixed instead by actually looping over the items and add a key there?


On a sidenote, shouldnt `id` be unique?

olemartinorg: Haha, yeah, you're right - you exposed my laziness.. :wink: I'll look into rewriting this to something better.

As for the `id` not being unique - yeah, I dunno.. I printed out all the `id` arguments given during the full test suite, and there were a bunch of duplicate IDs where `undefined` had been passed in (and just converted to a string at some point). It seems fickle and buggy to me, so I agree a proper cleanup is in order.

haakemon: I mean this "fixes" the problem, and definitely improves the output from tests - so maybe its worth merging this as-is now, and do a proper cleanup later. I assume that task is a bit bigger, and will require more testing. You have my approval at least 🙂

haakemon (approved):

src/altinn-app-frontend/src/utils/render.tsx
olemartinorg: Thanks! I'll look into it, but won't spend a week on it.. :wink:

Issue #321, opened 2022-07-05
Show warning message for uses where there are large groups that are being deleted


Febakke:
When trying to delete a row in our repeating groups the user should get a warning. 

## Sketches
![Skjermbilde 2022-07-05 kl  14 41 57](https://user-images.githubusercontent.com/33222679/177329956-ae108e61-36df-4189-a284-75d1d944e4af.png)
[Figma prototype](https://www.figma.com/proto/wnBveAG2ikUspFsQwM3GNE/Altinn-Studio-Apps?page-id=2090%3A6723&node-id=13550%3A9415&viewport=1436%2C-7265%2C0.22&scaling=scale-down&starting-point-node-id=13550%3A8883&show-proto-sidebar=1)

## Criterea
- [ ] When trying to delete the user gets a warning
- [ ] The app developer can decide if this is needed. Might be overkill for small groups

## Tasks
- [x] New modal in design system using the panel?
- [ ] Implement logic in app-frontend

PR #498, opened 2022-09-28
Implement delete-warning-popover for repeating-groups

Magnusrm:
This PR implements the PopoverPanel component from the Design System to provide users with a warning before deleting a row in repeating groups. The feature is available to app developers by setting the flag `edit.alertOnDelete: true` for a repeating group in the layout.

#### The color of the panel has changed from red to yellow:
This is in order to comply with the thought from our UX-designers that red equals error and yellow equals alerts. This was discussed with @Febakke and we agreed to go forward with a yellow design to make our users accustomed to thinking warning when seeing yellow

#### The buttons are different from the design:
The designed buttons are not yet developed in the design-system. Together with @Febakke we decided to make use of the buttons currently available in the design-system until new buttons have been developed. This choice was made to save time by not needing to develop new buttons directly in app-frontend-react that would soon be replaced with new buttons from design-system.

Screen Recording 2022-09-28 at 16 12 15

## Related Issue(s)
- #321 

## Verification
- [ ] All tests run green

## Documentation
- [ ] User documentation is updated with a separate linked PR in [altinn-studio-docs.](https://github.com/Altinn/altinn-studio-docs)
- [ ] JsonSchema changes in [altinn-cdn](https://github.com/Altinn/altinn-cdn)

### DIFF
--- a/test/cypress/e2e/integration/app-frontend/group.js
+++ b/test/cypress/e2e/integration/app-frontend/group.js
@@ -346,7 +346,7 @@ describe('Group', () => {
         cy.get(table).find(mui.tableElement).find(appFrontend.group.delete).should('be.visible').click();
         cy.get(table).find(mui.tableElement).find(appFrontend.designSystemPanel)
           .find(appFrontend.group.popOverDeleteButton).should('be.visible').click();
-        cy.get(table).find(mui.tableElement).first().invoke('text').should('not.contain.text', 'automation');
+        cy.get(table).find(mui.tableElement).eq(0).should('not.contain.text', 'automation');
       });
 
     // Navigate to main group and test delete warning popoup cancel and confirm


src/altinn-app-frontend/src/features/form/containers/RepeatingGroupTable.tsx
```diff
@@ -487,16 +527,59 @@ export function RepeatingGroupTable({
...
+                                  variant={ButtonVariant.Secondary}
+                                  onClick={() => onOpenChange(index)}
+                                >
+                                  {getLanguageFromKey(
+                                    'general.cancel',
+                                    language,
+                                  )}
+                                </Button>
+                              </div>
+                            </PopoverPanel>
```
haakemon, line 582: This component is getting quite big - perhaps this should be extracted to smaller parts?

src/altinn-app-frontend/src/features/form/containers/RepeatingGroupTable.tsx
```diff
@@ -187,6 +193,19 @@ const useStyles = makeStyles({
...
   },
+  popoverButtonContainer: {
+    display: 'flex',
+    '& button': {
+      margin: '1rem 1rem 0 0',
+      '&focus': {
+        outline: 'unset',
+      },
+    },
+  },
```
haakemon, line 204: We should avoid adding styles to html elements directly. Especially when these are components coming from our Design System components. These styles may conflict with styles already set on the component, and may break in weird ways when the component is updated, f.ex if the Button component was changed to use a `div` instead of `button`, or if it had some `margin` already set. Catching these types of errors are difficult since its not easily testable. (I know this example is pretty far fetched - changing a Button component to be a `div` under the hood is extremely unlikely, but it illustrates the point, and this is a generic problem with overriding styles)

Is the intention to add a gap between the buttons? If thats the case, I would suggest to instead use the `gap` property here on the parent flex container instead (`popoverButtonContainer`).

The margin-top here is probably for spacing between the text and the buttons? That margin should IMO be moved to the `popoverButtonContainer`, and not be on the buttons.

```suggestion
  popoverButtonContainer: {
    display: 'flex',
    gap: '1rem',
    marginTop: '1rem',
  },
```

Also not sure why the focus outline is being unset here, so my suggestion does not cater for that at all.

src/shared/src/components/molecules/AltinnMobileTableItem.tsx
```diff
@@ -164,6 +178,19 @@ const useStyles = makeStyles({
...
   },
+  popoverButtonContainer: {
+    display: 'flex',
+    '& button': {
+      margin: '1rem 1rem 0 0',
+      '&focus': {
+        outline: 'unset',
+      },
+    },
+  },
```
haakemon, line 189: Same comment here as in `RepeatingGroupTable.tsx`. Perhaps this should even be extracted to a separate component, to avoid having to duplicate the styles and logic here?

src/altinn-app-frontend/package.json
```diff
@@ -84,6 +84,7 @@
     "react-styleguidist": "12.0.0",
     "redux-mock-store": "1.5.4",
     "redux-saga-test-plan": "4.0.6",
+    "resize-observer-polyfill": "1.5.1",
```
olemartinorg, line 87: Hvorfor trenger vi en polyfill? :thinking: Jeg trodde ikke vi trengte støtte for IE11: https://caniuse.com/resizeobserver

olemartinorg: Ahh, ok, brukt i testen, ser jeg. Regner med at node 16 ikke støtter den da.

olemartinorg (approved):
Godt jobba! :clap:

src/altinn-app-frontend/package.json
Magnusrm: Det var testen som streiket for resize-observer ja. Av det jeg så da jeg undersøkte litt virket det som om jest ikke definerer en resize observer så den måtte mockes. Det som bruker resize-observer er den ene dependencien til radix-ui/popover som brukes i PopoverPanel komponenten i designsystemet.

PR #506, opened 2022-09-29
feature: include what field/component that triggered save

lorang92:
## Description

Adds two header values `field` and `componentId` when calling put on a data element (which in turns calls processWrite). 
This enables app developers with some c# knowhow to get the header values from the HTTP-context when running data processing. 

## Related Issue(s)
- [Slack discussion (internal link)](https://altinndevops.slack.com/archives/CDU1S3NLW/p1664343663566769)

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/altinn-app-frontend/src/features/form/data/submit/submitFormDataSagas.ts
+++ b/src/altinn-app-frontend/src/features/form/data/submit/submitFormDataSagas.ts
@@ -132,8 +132,8 @@ export function* putFormData({
   try {
     const options: AxiosRequestConfig = {
       headers: {
-        field,
-        componentId,
+        'X-DataField': field,
+        'X-ComponentId': componentId,
       },
     };
     yield call(put, dataElementUrl(defaultDataElementGuid), model, options);
@@ -224,8 +224,8 @@ export function* saveStatelessData({
   const allowAnonymous = yield select(makeGetAllowAnonymousSelector());
   let options: AxiosRequestConfig = {
     headers: {
-      field,
-      componentId,
+      'X-DataField': field,
+      'X-ComponentId': componentId,
     },
   };
   if (!allowAnonymous) {
--- a/src/altinn-app-frontend/src/features/form/data/submit/submitFormDataSagas.ts
+++ b/src/altinn-app-frontend/src/features/form/data/submit/submitFormDataSagas.ts
@@ -1,5 +1,6 @@
 import { all, call, put as sagaPut, select } from 'redux-saga/effects';
 import type { PayloadAction } from '@reduxjs/toolkit';
+import type { AxiosRequestConfig } from 'axios';
 import type { SagaIterator } from 'redux-saga';
 
 import { FormDataActions } from 'src/features/form/data/formDataSlice';
@@ -71,7 +72,7 @@ export function* submitFormSaga({
       return yield sagaPut(FormDataActions.submitRejected({ error: null }));
     }
 
-    yield call(putFormData, state, model);
+    yield call(putFormData, { state, model });
     if (apiMode === 'Complete') {
       yield call(submitComplete, state, stopWithWarnings);
     }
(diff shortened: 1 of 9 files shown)


src/altinn-app-frontend/src/features/form/data/submit/submitFormDataSagas.ts
```diff
@@ -116,15 +117,26 @@ function* submitComplete(state: IRuntimeState, stopWithWarnings: boolean) {
...
     state.applicationMetadata.applicationMetadata,
     state.instanceData.instance,
     state.formLayout.layoutsets,
   );
   try {
-    yield call(put, dataElementUrl(defaultDataElementGuid), model);
+    const options: AxiosRequestConfig = {
+      headers: {
+        field,
+        componentId,
```
olemartinorg, line 136: Does this mean we're sending plain HTTP headers like `componentId: something`? I don't think that's standards compliant, as custom headers should have the `X-` prefix so we should probably send something like `X-ComponentId: something` instead.

olemartinorg (approved):
While I'm not opposed to this quick-fix (I know we've discussed this for a long time, and there is certainly a need for it), I still think the question is a bit misunderstood. App developers want to know "which field changed here", but with repeating groups it can be an entire row at once. There is also no guarantee there won't be changes in `app-frontend-react` causing multiple values to change at the same time. Also, what if you have a spotty internet connection and some requests to save will fail, some will be delayed - that might make the backend miss the event it was looking for. Relying on getting the specific component ID that made the change, and trusting you'll always get that event, I think it's a recipe for subtle bugs in the future. The "proper" way (I think) should be to use a diffing tool in the backend to generate a list of what changed from before, and we already have the tools for that there.

src/altinn-app-frontend/src/features/form/data/submit/submitFormDataSagas.ts
lorang92: Ah, yes, totally agree. Fixing.

lorang92:
> While I'm not opposed to this quick-fix (I know we've discussed this for a long time, and there is certainly a need for it), I still think the question is a bit misunderstood. App developers want to know "which field changed here", but with repeating groups it can be an entire row at once. There is also no guarantee there won't be changes in `app-frontend-react` causing multiple values to change at the same time. Also, what if you have a spotty internet connection and some requests to save will fail, some will be delayed - that might make the backend miss the event it was looking for. Relying on getting the specific component ID that made the change, and trusting you'll always get that event, I think it's a recipe for subtle bugs in the future. The "proper" way (I think) should be to use a diffing tool in the backend to generate a list of what changed from before, and we already have the tools for that there.

Most definitely see your point. These headers I think should be looked at as a "best effort" at telling the backend what was the cause of this put request, and might not be something we even document out a supported "feature". This solves a need now, and I agree some diffing tool would be better. But that requires more work. And more testing. And more time. 

@nkylstad do you have have any thoughts on this? I do think @olemartinorg has a great point here. Since BRG technically does not go in production until 1. january, I guess they can live with some extra requests while we implement a better way to tell developers what changed "the proper way".

Or, should we merge this and leave it undocumented? Use at your own risk kind of thing. Any ideas, anyone?

olemartinorg:
I agree that documenting this with a disclaimer (or not documenting it) will work. When there's a pressing need, this works better than nothing, I guess.

@DanRJ: I seem to remember something about you using our JSON diff tool in OED to figure out which data fields have changed in `ProcessDataWrite`? My memories of the exact code is fuzzy, but was that something that could easily be re-used or documented as a suggested way to solve the problem of "I want to know what just changed"?

lorang92:
A diff might be something we should look into in the future as a good alternative. Might need some other changes in the backend for that to be viable. I'm merging this, seems harmless and solves real-world issues.

PR #509, opened 2022-09-30
bug: fix getHostname for new local test url

lorang92:
## Description

Quick and dirty fix for `getHostname` function that failed using newest local test url (local.altinn.cloud) and latest app frontend prod build.

Some of these methods could probably be cleaned up, but I did not take the time to evaluate if the throwing of errors are used for anything. 

Fix and forget.


olemartinorg (approved):
:facepalm:

Issue #511, opened 2022-10-02
Using dynamic options in page make it crash when going back to formfilling


TheTechArch:
### Description of the bug

In relle rettigheshavere we use dynamic options (disabled now). When it is enabled for the kontroll størrelse option it crashes when returning to Data task from confirmation. The options provider is not called

altinn-app-frontend.js:2 
        
       TypeError: Cannot read properties of undefined (reading 'index')
    at t.getIndexCombinations (altinn-app-frontend.js:2:2053084)
    at altinn-app-frontend.js:2:2075576
    at Array.forEach (<anonymous>)
    at t.getOptionLookupKeys (altinn-app-frontend.js:2:2075463)
    at d (altinn-app-frontend.js:2:2017417)
    at d.next (<anonymous>)
    at m (altinn-app-frontend.js:2:1697323)
    at l (altinn-app-frontend.js:2:1697655)
    at altinn-app-frontend.js:2:1693835
    at g (altinn-app-frontend.js:2:1690101)

### Steps To Reproduce

Enable dynamic options for this field. 
Fill ot form
Go to confirm
Go back to data

See commit to disable
https://altinn.studio/repos/brg/rrh-innrapportering/commit/e228ecfee1f4f0bb1d070bdab256ff5f1ccd73ff

olemartinorg:
I suspect this has been fixed now, as a consequence of #575 and #576. @TheTechArch, could you re-test this with [v3.54.0-alpha2](https://github.com/Altinn/app-frontend-react/releases/tag/v3.54.0-alpha2)? I've made some attempts to reproduce it now, but I haven't managed to do it (but I'm unsure if I used the app correctly).

allinox:
@olemartinorg 
I tested this just now - seems to me like it works 😊

olemartinorg:
@allinox Fantastic! Sadly, the fix is a bit risky - and along with the #540 PR, the next release is scheduled to have a lot of (risky) changes. I'm wanting to delay it until I can look over #556 and clean up a little as well. You'll hear from me soon!

olemartinorg:
Closing as it is tested OK, but my point about the delayed release remains.

PR #518, opened 2022-10-05
fix to make sure page doesn't crash

nkylstad:
## Description
Add a check for empty repeating groups to avoid page crashing if repeating groups are not loaded before options start fetching - this is relevant when app uses dynamic options with mapping to field in repeating group. This also only happens when going TO a data task (using dynamic options w/mapping) FROM a confirmation task. Have not tested going FROM other task types, although it might be a problem then to 🤷‍♀️ 

The root issue still needs to be fixed

## Related Issue(s)
- #511 

## Verification
- [ ] All tests run green

## Documentation


olemartinorg:
I think this can be closed now, as #511 looks to been fixed via #575 and #576. :partying_face:

PR #524, opened 2022-10-06
Fixing flaky tests sending HTTP requests

olemartinorg:
## Description
Some tests mysteriously failed in rare cases, seemingly because they called HTTP requests.

My hypothesis is that `InstantiationButton.test.tsx` would simulate a button click, which triggered a dispatched action, and while `dispatch` was spied upon - it was not mocked. This caused the action in redux to trigger `appApi` to initialize a request via `axios` to create an instance, and it would sometimes fail on some machines. The request itself is a by-product and not really part of the test, so most of the time the test would complete before the request got started (and everything would be OK).

This change mocks axios properly and should mock all outbound requests by default when running in unit tests. This also caused some other tests to fail (these were mocking responses using `msw`).

## Verification
- [x] All tests run green

## Documentation


bjosttveit (approved):
Great 🙌

PR #529, opened 2022-10-11
Faster tests without type-checking (& github actions improvements)

olemartinorg:
## Description
I've been banging my head against the wall trying to get unit tests to run faster on my machine, without any clear revolutionary results (except for @swc/jest, which was very fast, but didn't really work with our tests). This seems to be a fair compromise, by disabling type-checking when running ts-jest.

I also added some useful shortcuts for `tsc` which I use regularly. I don't think the existing one was used much, because it would either fail or spew out lots of files as a result.

## Verification
- [x] All tests run green

## Documentation


bjosttveit:
Did some performance testing on my own machine, and disabling type-checking does improve performance a bit. However, running the tests in-band makes them significantly slower for me. Maybe the M1 Max just has very good multicore performance. How do the results compare on your machine? How is the performance when running checks on GitHub Actions?

| | Type checking | No type checking |
|--------|---------|--------|
| In-band | image | image |
| Not in-band | image | image |

bjosttveit:
Have you tried reducing the maximum number of workers (to for example 50%) instead of running them in-band?
https://dev.to/vantanev/make-your-jest-tests-up-to-20-faster-by-changing-a-single-setting-i36

olemartinorg:
Fantastic! You approached this much more scientifically correct than I did.. :wink: 

On my machine:

<table>
<tr>
	<td>In-band
	<td>
<tr>
	<td>Not in-band
	<td>
<tr>
	<td>50% workers
	<td>
</table>

Honestly, I'm very impressed by your 19sec one. I have a dual xeon machine at home (fairly old xeon processers, admittedly) with 12 physical cores each (so 24 physical cores, 48 with hyperthreading). The best results I got from that was about 24sec with 50% workers (using all threads there is 5-10sec slower, and makes some test fail - strange!).

In conclusion, I think think we should keep the no-typechecking part, but only use `--runInBand` on github actions. Agree?

bjosttveit (approved):
Yes, no type-checking is good, and running in-band on Actions seems to be the way to go 👍

Issue #442, opened 2022-09-07
Deleting top row of a nested repeating group causes visual bug


Magnusrm:
### Description of the bug

Deleting the top row of a nested group causes only the new top row to be displayed. The data for the remaining groups seems to be intact as the data shows up for the remaining rows when adding new items. 

Deleting the top row inside a nested repeating group should result in displaying all the rows except for the deleted row.

### Steps To Reproduce

1. Clone dynamics-nested-group app
2. Run app with Localtest or in an environment
3. Enter the nested repeating group and add several elements to the group
4. Delete the top row

### Additional Information

![Screen Recording 2022-09-07 at 08 38 05](https://user-images.githubusercontent.com/32294735/188815629-c7a38a8f-6467-45d3-a393-ea8ce0c7457d.gif)

Magnusrm:
The same behaviour happens when openByDefault is activated for a regular un-nested group. The problem may lie with openByDefault logic and not nested groups.

Issue #398, opened 2022-08-19
Visual bug for "add new" button for nested repeating groups on small screens


lorang92:
### Description of the bug

See title.

### Steps To Reproduce

1. Add a nested rep group
2. Use a small screen width ( < 993px ) 
3. See error.

Note that this error is only visible when the nested group has 0 elements.

When you add an element the issue seems to be nonexistent. 

### Additional Information

Wide screens seems ok:
![visual_bug_wide](https://user-images.githubusercontent.com/10416981/185595550-5233d9bf-3acb-4ff6-8c7b-26abc25f4f9d.PNG)

Smaller not so much:

![visual_bug_small](https://user-images.githubusercontent.com/10416981/185595555-f7751972-28ab-47bf-9616-958fb14698f4.PNG)

PR #544, opened 2022-10-14
Bug: Fix repeating group UI issues

bjosttveit:
## Description

Minor css changes to fix the width of the Add button in nested repeating groups. Removed leaveOpen-logic from openByDefault as it seems to do nothing besides causing the bug in #442. Nobody I asked knew what the intended behaviour of leaveOpen was, and openByDefault works as documented without this logic.

## Related Issue(s)
- #398 
- #442 

## Verification
- [x] All tests run green

## Documentation

### DIFF
--- a/test/cypress/e2e/integration/app-frontend/attachments-in-group.js
+++ b/test/cypress/e2e/integration/app-frontend/attachments-in-group.js
@@ -430,7 +430,6 @@ describe('Repeating group attachments', () => {
 
     // Verify that one of the attachments in the next nested row is visible in the table header. This is also a trick
     // to ensure we wait until the deletion is done before we fetch the redux state.
-    cy.get(appFrontend.group.rows[0].nestedGroup.rows[0].editBtn).click();
     cy.get(appFrontend.group.rows[0].nestedGroup.rows[0].uploadTagMulti.tableRowPreview)
       .should('contain.text', filenames[0].nested[1][2]);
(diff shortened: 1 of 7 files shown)


test/cypress/e2e/integration/app-frontend/attachments-in-group.js
```diff
@@ -430,7 +430,6 @@ describe('Repeating group attachments', () => {
 
     // Verify that one of the attachments in the next nested row is visible in the table header. This is also a trick
     // to ensure we wait until the deletion is done before we fetch the redux state.
-    cy.get(appFrontend.group.rows[0].nestedGroup.rows[0].editBtn).click();
```
olemartinorg, line 433: Maybe add tests for the new behaviour where deleting a single row with `openByDefault` now leaves the group closed for editing?

olemartinorg (approved):

test/cypress/e2e/integration/app-frontend/attachments-in-group.js
olemartinorg: I agree, but having assertions for this in the test won't hurt! You changed the behaviour of something, and since this is the only part that changes any of the cypress tests, it looks to me like we don't have enough test coverage to clearly define the behaviour of `openByDefault` in our tests.

bjosttveit: Maybe, the tests do cover all of the documented behaviour, but it was not supposed to do anything when a row is deleted as long as there are more rows left.

PR #540, opened 2022-10-13
Expressions

olemartinorg:
## Description
This PR adds support for dynamic expressions, as used in the following component properties:
- `required`
- `readOnly`
- `hidden` (new!)
- `edit.addButton` (for repeating groups)
- `edit.saveButton` (for repeating groups)
- `edit.deleteButton` (for repeating groups)
- `edit.saveAndNextButton` (for repeating groups) 

Along with the following properties for layouts/pages:
- `hidden` (new!) This makes it an alternative to tracks.

It includes the following expression functions for now:
- `equals` / `notEquals`
- `greaterThan` / `greaterThanEq` / `lessThan` / `lessThanEq`
- `concat`
- `and` / `or`
- `if` (beta!) - else branch is optionally passed in with parameters
- `instanceContext`
- `frontendSettings`
- `component` (resolves upwards inside repeating groups)
- `dataModel` (resolves its position inside repeating groups without the need for position placeholders)

Try this out using the test app: https://dev.altinn.studio/repos/olemartinorg/layout-expressions

Because of the size of this PR, we'll do a review meeting instead of performing code review inside github. Marking this as draft until that meeting is over.

## Related Issue(s)
- #355

## Verification
- [x] **Your** code builds clean without any errors or warnings
- [x] Manual testing done (required)
- [x] Relevant automated test added (if you find this hard, leave it and we'll help out)
- [x] All tests run green

## Documentation
- ~~[ ] User documentation is updated with a separate linked PR in [altinn-studio-docs.](https://github.com/Altinn/altinn-studio-docs) (if applicable)~~
  - Altinn/altinn-studio-docs/issues/625
  - Altinn/altinn-cdn/issues/108

PR #553, opened 2022-10-18
Expressions: Type-changes

olemartinorg:
## Description
Splitting up the massive #540 into smaller parts. This part is the simplest one; mostly type-changes. Until now the props sent from `GenericComponent` and passed to the respective components have been defined for each component, but now I've introduced `PropsFromGenericComponent` to handle this for all of them. This prepares for expressions, where the expressions inside the props will be resolved by `GenericComponent` (thus altering what is sent from `GenericComponent` to all other components.

Removing the `hidden` prop passed into `GenericComponent`, as this is just a result of the read from redux state (which `GenericComponent` does anyway).

## Related Issue(s)
- #540

## Verification
- [x] All tests run green

## Documentation


Magnusrm (approved):
Looks good to me 👑

PR #554, opened 2022-10-18
Expressions: Tracks

olemartinorg:
## Description
Splitting up the massive #540 into smaller parts. This part includes a new redux structure for 'tracks', (i.e. showing/hiding entire layouts/pages). This structure is needed to group together state that revolves around which pages should be visible, and the order of these pages.

The new rules for page visibility and order is:
1. The order is decided by the backend
2. Hidden pages are removed from the ordered list by the backend
3. The frontend may also hide some pages via expressions in the layout definition (but this should not affect the order)

This PR includes the structures needed to store the expressions in the frontend, the result of these expressions (calculated when there are changes to the form data, etc), and the code to calculate the final ordered list of layouts. 

## Related Issue(s)
- #540

## Verification
- [x] All tests run green

## Documentation

### DIFF
--- a/src/altinn-app-frontend/src/types/index.ts
+++ b/src/altinn-app-frontend/src/types/index.ts
@@ -150,6 +150,10 @@ export interface IValidationIssue {
   targetId: string;
 }
 
+export interface IHiddenLayoutsExpressions {
+  [layoutKey: string]: never; // Will be set in a later PR
+}
+
 export interface IUiConfig {
   autoSave: boolean;
   currentView: string;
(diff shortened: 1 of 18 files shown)


src/altinn-app-frontend/src/types/index.ts
```diff
@@ -160,14 +164,41 @@ export interface IUiConfig {
...
+  /**
+   * This state contains the results from calculating `hiddenExpr` (expressions to decide if a certain layout should
+   * be hidden or not). If a layout is in this list, is should also not be displayed.
+   */
+  hidden: string[];
+
+  /**
+   * List of expressions containing logic used to show/hide certain layouts.
+   */
+  hiddenExpr: IHiddenLayoutsExpressions;
```
Magnusrm, line 199: Could `hidden` and `hiddenExpr` be nullable so that one does not have to include a empty list and an empty expression object when hidden functionality is not wanted?

Magnusrm (approved):

src/altinn-app-frontend/src/types/index.ts
olemartinorg: I guess, yes, but i'd rather include these empty lists in tests (which would fail loudly without them) than somehow forget to add them to production code not covered by tests (where it would fail silently/weirdly, or just crash the app). I think nullability is evil as long as we have set up `"strictNullChecks": false` in our `tsconfig.json`.

olemartinorg: Btw, I had that on my TODO-list, but reported it now: #556

Issue #355, opened 2022-07-29
Logic/expressions in layout


olemartinorg:
### Description

Several issues require some form of logic/dynamics in layout files. Along with the solution proposed in #220, we can start implementing this form for layout logic (in issues like #61 and #64), without introducing breaking changes (as in #220).

This issue tracks the implementation of layout expressions in frontend. The project itself is tracked [in more detail here](https://github.com/orgs/Altinn/projects/22).

### In scope

New layout logic, as described much more clearly in #220

### Out of scope

Breaking changes :wink: 

### Tasks

Development is underway in the [feat/layout-expressions](/Altinn/app-frontend-react/tree/feat/layout-expressions) branch.

- [x] Implement tools to work on a layout as a hierarchical structure (respecting repeating group rows, etc), so the expression can be evaluated in the context of a layout node (finding the closest component references when looking up another component id - which could be inside the same repeating group row, the one above, etc).
- [x] Implement support for evaluating recursive layout expressions
- [x] Implement a `useLayoutExpression()` hook that can resolve layout expressions anywhere
- [x] Test the above hook in a unit-test
- [x] Use resolved layout expressions (for the `required` property) when validating a group using `triggers: ['validation']`
- [x] Validate all expressions to make sure they are valid when loading the layout (generate error messages if not)
- [x] Implement support for pre-processing compact lisp-like expressions into full/verbose expressions (do this once in the fetchLayout saga, along with the initial layout expression validation)
- [x] Evaluate expressions in sagas to make sure we still use and populate the `hiddenFields` state even when using layout expressions.
  - This could probably benefit from some index, optimizing the code so that we don't have to iterate the entire layout for every form item change - only for those that affects layout expressions.
- ~~[ ] Optimize the code so layout expressions (and especially the hierarchical layout functions) don't run when we don't absolutely need them to. Memoization should go a long way. ~~ I think I'm deciding against this for now, as it seems to be fast enough (optimizations now might be premature)
- [x] Make sure expression results are used in places where the properties are read. This means expressions needs to be resolved when performing empty field validation (for expressions in the the `required` property), etc.
- [x] Investigate changes/increases in time and memory spent type-checking after these changes. There are quite a few new (and slow/complex) types added in the branch, and during development node has sometimes crashed with out-of-memory errors. Make sure type-checking for happy-path changes does not change significantly for other developers.
- [x] Support hiding entire pages/layouts
- [x] Clean up code to avoid complexity warnings from eslint
[...]

Issue #345, opened 2022-07-18
Restructure redux state to better reflect component state, validation and repeating groups 


olemartinorg:
### Description

Today, `app-frontend-react` will fetch the layout(s) from the server and store them in the `formLayout.layouts` redux state almost unchanged. When, for example, a form input is changed, this will update the state in `formData`, a different redux root state. This is also the case for other state directly tied to a specific component, such as:

- Validation messages - in the `formValidations.validations` state
- Conditional rendering - rules stored in `formDynamics.conditionalRendering` and effects stored in `formLayout.uiConfig.hiddenFields`
- Attachments (for `FileUpload` and `FileUploadWithTag`), stored in `attachments.attachments`
- Tags (for `FileUploadWithTag`), stored in `formLayout.uiConfig.fileUploadersWithTag`
- Repeating group rows, stored in `formLayout.uiConfig.repeatingGroups`

Repeating groups contributes to making this especially difficult to work with, as we have to make sure all state is kept in sync when adding/deleting rows to repeating groups; i.e. that the validation state, attachments, etc. is updated to remove the state belonging to a row that was just deleted, and that subsequent rows gets "moved up" to replace the removed row.

With this architecture it means we have to write a lot of code to move stuff around, which sometimes might be finicky and might lead to bugs and unexpected behaviour ([such as `FileUpload` not being supported inside repeating groups](https://github.com/Altinn/app-frontend-react/issues/259), some [validations not working for DatePicker when in repeating groups](https://github.com/Altinn/app-frontend-react/issues/185) or some [validations not working in combination with conditional rendering when in repeating groups](https://github.com/Altinn/app-frontend-react/issues/296)). It also means that implementing fixes for these require lots of code to move things around, which leads to [large PRs for seemingly minor features](https://github.com/Altinn/app-frontend-react/pull/246).

I propose to invert the way we use the Redux store to keep state about components, rewriting this to work in a way where all the state tied to a component is stored alongside the component itself.

### Additional Information

Redux with nested state and dynamically adding reducers:
- https://github.com/reduxjs/redux-toolkit/issues/259
- https://medium.com/@matheusmm/redux-injecting-reducers-dynamically-e4e193eaa905

### Relevant/related issues
- #205

### Analysis

A proposed solution is to use nested slices to construct a hierarchy from the layout definition, along with mixing in re-usable reducers for common functionality (such as form data, validations, etc).

An example of such a hierarchical state (somewhat abbreviated, with explaining comments):

[...]

PR #561, opened 2022-10-19
Expressions: Engine

olemartinorg:
## Description
Splitting up the massive #540 into smaller parts. After #553 and #554, this PR is the big one with the whole engine for evaluating expressions, and everything needed to at last implement support for expressions in `app-frontend-react`. The last PR will utilize the functionality implemented here into `GenericComponent` etc (and that PR will arrive when this has been merged).

The overview of the functionality in this PR can be seen from the folder structure:
- The expression engine in `src/features/expressions`
	- The runtime expression evaluation engine, with functions and type-casting in `index.ts`
  - `prettyErrors`, which allows for listing multiple errors deep into a structured object and printing them colored to the console
  - `useExpressions` as a react hook, which lets you evaluate expressions (also inside an object) in your component
  - Validation code to make sure expressions are well-formed at the point they are loaded from layout files (gives an error message if expressions are invalid, and replaces the expression with a default value).
  - Shared tests for the expressions engine
- Utilities for traversing and mutating the string in a data model binding in `src/utils/databindings/Databinding.ts`
- Utilities for converting a flat layout to a hierarchical structure in `src/utils/layout/hierarchy*`
- React hooks for reading the entire layout (or all layouts), converting them to a hierarchical structure and (optionally) resolve expressions inside the nodes. 

A core concept in this structure is a new term for `app-frontend-react`; a `node`. Currently we have used the name `component` for the components defined in the layout definitions. And we often refer to a specific `component`, even when that component is possibly rendered multiple times (in a repeating group). I use the term `node` here for that case - and my definition of a node is that it's an instance of a `component` (for which there might be multiple `nodes` when dealing with repeating groups).

The work around converting a flat layout into a hierarchical structure with nodes as component instances can be seen as a precursor to what's needed to implement #345. The tools in `src/utils/layout/hierarchy*` are optional now, and will be used in expressions and validation, but in time I believe these concepts should replace the structure of our current redux store and should be used everywhere.

## Related Issue(s)
- #540
- #355

## Verification
- [ ] All tests run green

## Documentation

### DIFF
--- a/src/altinn-app-frontend/package.json
+++ b/src/altinn-app-frontend/package.json
@@ -90,6 +90,7 @@
     "terser-webpack-plugin": "5.3.6",
     "ts-jest": "29.0.3",
     "ts-loader": "9.4.1",
+    "utility-types": "^3.10.0",
     "webpack": "5.74.0",
     "webpack-cli": "4.10.0",
     "webpack-dev-server": "4.11.1"
--- a/src/altinn-app-frontend/src/features/expressions/ExprContext.ts
+++ b/src/altinn-app-frontend/src/features/expressions/ExprContext.ts
@@ -1,0 +1,171 @@
+import dot from 'dot-object';
+
+import {
+  ExprRuntimeError,
+  NodeNotFound,
+  NodeNotFoundWithoutContext,
+} from 'src/features/expressions/errors';
+import {
+  prettyErrors,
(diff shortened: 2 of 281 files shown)


bjosttveit (approved):
Sett igjennom så mye jeg klarer, så bra veldig ut 🙌 🙌

Issue #569, opened 2022-10-20
openByDefault = "first" does not work with filters


bjosttveit:
### Description of the bug

openByDefault = first sets the editIndex to 0 regardless of filters. If the index 0 is filtered out, no item will be opened. Additionally this leads to the add new element button disappearing. Probably also applies to openByDefault = "last".

![image](https://user-images.githubusercontent.com/47412359/196905141-b0f8862d-b04f-47a1-8476-b34d8bd81275.png)

### Steps To Reproduce

1. Set up a repeating group with a filter that hides the first element, and uses openByDefault = 'first'.

```json
"edit": {
    "openByDefault": "first",
    "filter": [
      {"key": "Innflytter.TidligereBosteder.Postnr","value": "0102"}
    ]
},
```

StianVestli:
Ser ok ut nå etter det man kan se.

Issue #339, opened 2022-07-14
Considerations for v4


haakemon:
# Description
After `v3` comes `v4`. This issues lists changes we need to (or want to) complete in the breaking changes window for `v4`.

```[tasklist]
### Breaking changes
- [ ] https://github.com/Altinn/app-frontend-react/pull/1446
- [ ] https://github.com/Altinn/app-frontend-react/pull/1441
- [ ] https://github.com/Altinn/app-frontend-react/pull/1444
- [ ] https://github.com/Altinn/app-frontend-react/issues/1148
- [ ] https://github.com/Altinn/app-frontend-react/issues/975
- [ ] https://github.com/Altinn/app-frontend-react/issues/820
- [ ] https://github.com/Altinn/app-frontend-react/issues/1468
- [ ] https://github.com/Altinn/app-frontend-react/issues/1485
- [ ] https://github.com/Altinn/app-frontend-react/issues/1486
- [ ] https://github.com/Altinn/app-frontend-react/issues/628
- [ ] https://github.com/Altinn/app-frontend-react/issues/1487
- [ ] https://github.com/Altinn/app-frontend-react/issues/1488
- [ ] https://github.com/Altinn/app-frontend-react/issues/1489
- [ ] https://github.com/Altinn/app-frontend-react/issues/1491
- [ ] https://github.com/Altinn/app-frontend-react/issues/1493
- [ ] https://github.com/Altinn/app-frontend-react/issues/1495
- [ ] https://github.com/Altinn/app-frontend-react/issues/1496
- [ ] https://github.com/Altinn/app-frontend-react/issues/1500
- [ ] https://github.com/Altinn/app-frontend-react/issues/1502
- [ ] https://github.com/Altinn/app-frontend-react/issues/1512
```

```[tasklist]
### Data model related changes
- [ ] https://github.com/Altinn/app-frontend-react/issues/1175
- [ ] https://github.com/Altinn/app-frontend-react/issues/555
- [ ] https://github.com/Altinn/app-frontend-react/issues/1484
- [ ] https://github.com/Altinn/app-frontend-react/issues/273
- [ ] https://github.com/Altinn/app-frontend-react/issues/925
- [ ] https://github.com/Altinn/app-frontend-react/issues/1111
- [ ] https://github.com/Altinn/app-frontend-react/issues/1147
- [ ] https://github.com/Altinn/app-frontend-react/issues/707
- [ ] https://github.com/Altinn/app-frontend-react/issues/743
[...]

PR #571, opened 2022-10-20
Fixed openByDefault to use filter

bjosttveit:
## Description

Open by default "first" and "last" is also broken by filters. In order to fix this the saga doing the open by default also needs to compute the `filteredIndexList` like the GroupContainer does. This function has been moved out of the component so that both the saga and the GroupContainer can use the same implementation.

~~I tried to make a general function, but the existing implementation did not make complete sense to me. It did not seem like it would work as documented, but by trying to reimplement it differently it broke some cypress tests.~~
The issues related to filters are documented here: https://github.com/Altinn/app-frontend-react/issues/339#issuecomment-1286624933

## Related Issue(s)
- #569 

## Verification
- ~~Relevant automated test added (if you find this hard, leave it and we'll help out)~~
- [x] All tests run green

## Documentation


olemartinorg (approved):

bjosttveit:
For future reference, here is the "more correct" implementation of `getRepeatingGroupFilteredIndices`: https://github.com/Altinn/app-frontend-react/blob/2b4ad09f85fd69f9fe83eb260bd500547643d2d4/src/altinn-app-frontend/src/utils/formLayout.ts#L561-L608

PR #576, opened 2022-10-21
Fixing duplicate startInitialDataTaskQueue action

olemartinorg:
## Description
The `startInitialDataTaskQueue` action is dispatched twice (creating a rush of more actions and requests than we need) because the hook dependencies had changed in some properties that wasn't read here. Moving the specific props out to their own variables.

## Related Issue(s)
- #511 (this _might_ be the root cause for that bug, but testing that hypothesis remains)
- #575

## Verification
- [ ] All tests run green

## Documentation


bjosttveit (approved):
Nice catch

Issue #548, opened 2022-10-17
Navigating to an error message in a repeating group should open the row for editing


olemartinorg:
If the user is stopped at navigation to the next page without filling some required fields in a repeating group, the error message in the new error box on the bottom of the page lets you navigate to the correct field - bu this does not work as expected for fields inside repeating group rows.

### Example

![bigtable](https://user-images.githubusercontent.com/700139/196140166-3bb52737-33ef-40aa-8bcf-f2986588f47a.png)

In the screenshot above, clicking 'alder' only navigates you to show the field inside the currently open group, and clicking the other error messages does nothing.

### Expected behaviour
Clicking an error message should expand the relevant repeating group row for editing. The same should also happen for nested repeating groups.

Issue #547, opened 2022-10-17
Remove 'One of the rows is incorrectly filled out' message


olemartinorg:
### Background

After #100, and with the errors already showing up in the repeating group table overview, this error message is unnecessary.

![en-av-radene-fyllt-ut-riktig](https://user-images.githubusercontent.com/700139/196139023-e6f523ad-47fc-45bc-bba6-f973bc7a7a6a.png)

It also follows an older design for error messages.

PR #588, opened 2022-10-28
Fix/repeating groups validation

bjosttveit:
## Description

1. Removed this warning 
image
2. Navigating to error message will now open the repeating group rows for editing and navigate to the correct multiPage
3. Also fixes an issue related to deleting validation when a row (with multiPage) is deleted

Note: One behavior was changed which required updating the cypress tests in `attatchments-in-group`. Previously, clicking save and close in the table row (not edit container) would not reset the multiPageIndex to 0. This would however happen if the save button inside the edit container was clicked. This has been changed so that in both cases, the multiPageindex is set to 0 when clicking save and close. The changes in cypress tests `groups` was just removing the old warning message.

## Related Issue(s)
- #547 
- #548 

## Verification
- [x] All tests run green

## Documentation

### DIFF
--- a/src/altinn-app-frontend/src/components/message/ErrorReport.tsx
+++ b/src/altinn-app-frontend/src/components/message/ErrorReport.tsx
@@ -9,7 +9,7 @@ import { renderLayoutComponent } from 'src/features/form/containers/Form';
 import { FormLayoutActions } from 'src/features/form/layout/formLayoutSlice';
 import { nodesInLayout } from 'src/utils/layout/hierarchy';
 import { getMappedErrors, getUnmappedErrors } from 'src/utils/validation';
-import type { ILayout, ILayoutGroup } from 'src/features/form/layout';
+import type { ILayout } from 'src/features/form/layout';
 import type { AnyChildNode } from 'src/utils/layout/hierarchy.types';
 import type { FlatError } from 'src/utils/validation';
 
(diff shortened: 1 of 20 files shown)


src/altinn-app-frontend/src/components/message/ErrorReport.tsx
```diff
@@ -80,22 +89,74 @@ const ErrorReport = ({ components }: IErrorReportProps) => {
...
+          parent.item.maxCount > 1
+        ) {
+          const child =
+            i == 0
+              ? componentNode
+              : (allParents[i - 1] as AnyChildNode<'unresolved'>);
+
+          // Go to correct multiPage page if necessary
+          if (
+            (parent.item as RepeatingGroupHierarchy<'unresolved'>).edit
```
olemartinorg, line 119: If you had assigned `parent.item` to a variable before the `if` above, I think TypeScript might have inferred this type automatically

src/altinn-app-frontend/src/components/message/ErrorReport.tsx
```diff
@@ -80,22 +89,74 @@ const ErrorReport = ({ components }: IErrorReportProps) => {
...
+              ?.multiPage
+          ) {
+            const childrenWithMultiPageIndex = (
+              layouts[error.layout].find(
+                (c) => c.id === parent.item.id,
```
olemartinorg, line 124: I see it's unfortunate that I intentionally removed the multipage stuff now.. :wink: Btw, `baseComponentId` might be better than `parent.item.id`, as it would fail if the parent was a nested repeating group.

olemartinorg: We could possibly extend all the components to include a `multiPageIndex` property when iterated over in `hierarchy.ts`.. :thinking: That might be better for us in the future than falling back to reading the flat layout here.

olemartinorg (approved):
Great work! :rocket:

src/altinn-app-frontend/src/components/message/ErrorReport.tsx
bjosttveit: Hehe, agree, I tried to look into it but didn't figure it out

bjosttveit: as in: `parent.item.baseComponentId`? That did not work for me

bjosttveit: Like this maybe?
`(c) => c.id === parent.baseComponentId ?? parent.id`
Looks like only group-children get a basecomponentId so if it is a non-nested group it only has id?

olemartinorg: I pushed a commit that adds `multiPageIndex` now - lets see if that works! :crossed_fingers:

bjosttveit: MVP🏅

bjosttveit: Works like a charm!

PR #591, opened 2022-10-31
extract hidden prop for custom components

nkylstad:
## Description
Fix issue with `hidden="false"` sent to custom (web) component. This was causing browser to interpret that `hidden` was set on the html element, thus hiding it.
Solved by setting the `hidden` logic in the `CustomWebComponent` react component.

#592 

## Verification
- [ ] All tests run green


olemartinorg (approved):
:clap:

Issue #556, opened 2022-10-18
Technical debt: Enable null-checks in TypeScript


olemartinorg:
### Description
We currently have `"strictNullChecks": false` in our `tsconfig.json`, which lets us write dangerous code that never checks for types/properties that could be nullable/undefined. This issue alone has lead to multiple hard crashes/`TypeError`s in production code this year alone. Setting this value to true causes 954 errors in the main branch right now, but cleaning these up is mostly a menial chore rather than something that requires acute attention to detail.

Prefer to do this at a point in time when there is less activity in the repo than usual. 

### Related issues
- Altinn/app-template-dotnet/issues/122
- #440
- Altinn/altinn-studio/issues/9032

PR #551, opened 2022-10-17
Null-checking edit property

olemartinorg:
## Description
Adding null-checks for possibly missing `edit` property in repeating group component

## Verification
- [ ] **Your** code builds clean without any errors or warnings
- [ ] Manual testing done (required)
- [ ] Relevant automated test added (if you find this hard, leave it and we'll help out)
- [ ] All tests run green

## Documentation
- [ ] User documentation is updated with a separate linked PR in [altinn-studio-docs.](https://github.com/Altinn/altinn-studio-docs) (if applicable)

PR #597, opened 2022-11-01
Strict null-checks

olemartinorg:
## Description
This PR enables strict null-checks in TypeScript, and makes sure all (currently) nullable values are properly checked. There should probably be some more nullable values, since the entire codebase has been written without strict null checks (so nullability was implicit everywhere until now). This is messy and suboptimal in many ways, but it's a step in the right direction where future changes are now much less likely to throw unexpected errors on values that are null/undefined.

## Related Issue(s)
- #556 
- #551
- #512 
- #438

## Verification
- [x] All tests run green

## Documentation


bjosttveit (approved):
LGTM

Magnusrm (approved):

PR #609, opened 2022-11-02
Make MultipleChoiceSummary match styling of normal summary

ivarne:
### Screenshots
See that 3.1 (`"type": "Checkboxes"`) has the wrong font weight on summary page.
image
a99bfb73-5328-4a45-81fd-931a6806007d


olemartinorg:
I see that a snapshot test failed, @ivarne.. We should get rid of these snapshot tests at some point, but that's not your responsibility. Would you mind running the tests locally, and commit the updated snapshot as well?

ivarne:
Da tror jeg det skal være i orden

olemartinorg (approved):
Thanks! :clap:

PR #610, opened 2022-11-02
Fix language inconsistencies

bjosttveit:
## Description

I noticed that general.log_out was missing in nynorsk:

image

There were also some other inconsistencies and missing texts that I fixed. I also reordered some nynorsk fields so that they have the same order as bokmål and english.


olemartinorg (approved):
Nice! These things should probably be fixed forever after #378 gets implemented. At that point we can infer all the available keys as a TypeScript type, and require that all languages implement the same set of keys (and we get auto-completion when using them - yay!)

bjosttveit:
> Nice! These things should probably be fixed forever after #378 gets implemented. At that point we can infer all the available keys as a TypeScript type, and require that all languages implement the same set of keys (and we get auto-completion when using them - yay!)

That sounds much better!

PR #546, opened 2022-10-14
Simplifying flattenObject()

olemartinorg:
## Description
In the refinement meeting today, we discussed the technical debt that is our always-stringy types. Part of this, I think, is caused by the flattenObject() function, which takes any object and converts it to a dot-object (and converts all types to strings). I remembered I had a shelved patch that cleans up this function and makes the type-casting to string very obvious.

# Related issue(s)
- #205

## Verification
- [x] **Your** code builds clean without any errors or warnings
- [ ] Manual testing done (required)
- [x] Relevant automated test added (if you find this hard, leave it and we'll help out)
- [ ] All tests run green

## Documentation
- ~~[ ] User documentation is updated with a separate linked PR in [altinn-studio-docs.](https://github.com/Altinn/altinn-studio-docs) (if applicable)~~

PR #611, opened 2022-11-02
Re-introducing brokenness in flattenObject

olemartinorg:
## Description
@StianVestli found a bug where old-style dynamics did not evaluate correctly after #546 was merged, as empty string values deep inside objects were deleted in the old `flattenObject` implementation, but that quirkyness was not carried over in #546. That caused the newer code to return an empty string (as it should be) to the conditional rendering rule function, which caused it to hide a Paragraph component that did not hide before. Breaking changes should only be released in new major versions (see #339).

## Related Issue(s)
- #546
- #339

## Verification
- [x] All tests run green

## Documentation


bjosttveit (approved):
I love backwards compatibility 🤒 😅

PR #625, opened 2022-11-04
Schema: Require either numeric or pattern format

bjosttveit:
## Description

Since `react-number-format` v5, the split between NumericFormat (thousand-separator, suffix, etc.) and PatternFormat (format, mask, etc.) is more explicit. For backwards compatibility, our implementation ignores this and figures out which component to use based on the props provided. In order to make it easier to set up number-formatting I have updated the schema so that you can either provide props only for NumericFormat, or only for PatternFormat. Additionally, when using PatternFormat, the `format`-prop is required, this can now be enforced in the schema, since there are effectively two sets of allowed properties with their own restrictions.

At first, all properties are valid:

image

If you add `format`, it will now only match the PatternFormat-props:

image

It will also inform that `format` is required for PatternFormat:

image

If you add props from NumericFormat, it will no longer suggest props from PatternFormat (format, mask, etc.):

image

## Related Issue(s)
- https://github.com/Altinn/altinn-design-system/pull/156

## Verification
- ~~**Your** code builds clean without any errors or warnings~~
- ~~Relevant automated test added (if you find this hard, leave it and we'll help out)~~
- ~~All tests run green~~

## Documentation

### DIFF
--- a/schemas/json/component/number-format.schema.v1.json
+++ b/schemas/json/component/number-format.schema.v1.json
@@ -5,7 +5,7 @@
   "description": "Schema that describes the options that can be configured for number formatting on an `input` component, based on react-number-format package. For complete list of available options, see https://s-yadav.github.io/react-number-format/docs/props",
   "type": "object",
   "additionalProperties": true,
-  "oneOf": [
+  "anyOf": [
     {
       "properties": {
         "allowedDecimalSeparators": {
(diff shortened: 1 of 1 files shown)


olemartinorg (approved):
:exploding_head:

I :heart: JsonSchema

PR #621, opened 2022-11-03
Prettier: 120 character printWidth

olemartinorg:
## Description
As discussed on Slack, setting the printWidth to 120 characters. This PR includes a one-line change in prettier.config.js and a huge diff from `eslint --fix`.

## Related Issue(s)
- https://altinndevops.slack.com/archives/CDU1S3NLW/p1661331887029039

## Verification
- [ ] **Your** code builds clean without any errors or warnings
- [ ] Manual testing done (required)
- [ ] Relevant automated test added (if you find this hard, leave it and we'll help out)
- [ ] All tests run green

## Documentation
- [ ] User documentation is updated with a separate linked PR in [altinn-studio-docs.](https://github.com/Altinn/altinn-studio-docs) (if applicable)

PR #626, opened 2022-11-04
Update vs-code settings (print-width)

bjosttveit:
## Description

Update `editor.rulers` in `settings.json` to match the new prettier print-width.

## Related Issue(s)
- https://github.com/Altinn/app-frontend-react/pull/621


Magnusrm (approved):
👑

PR #627, opened 2022-11-07
Removing validation messages when hiding fields using dynamics

### LINKED ISSUE #617: Expressions: Newly hidden fields keep validation messages
### Description of the bug

Fields hidden using expressions (#540) may retain their validation messages, if any exists.

### Steps To Reproduce

1. Add logic to show/hide another field
2. Induce a validation error in the hidable field
3. Hide the field
4. Observe the error message still shows up in the bottom error box

### Additional Information

_No response_

olemartinorg:
## Description
This change removes validation messages when fields are hidden. This is sub-optimal, as validations should also be hidden along with fields - but implementing such a change is quite a bit riskier (since every location working with validations would have to take into consideration if the validation messages should be hidden or not).

This change is a bit large because:
- It changes the 'single field validation' functionality to pass data/dependencies through other actions instead of relying on global redux state being set in `GenericComponent` which magically gets read by the single field validation action later. That was tied to only being able to work on one field at a time, and it was non-obvious which actions could lead to single-field validation.
- In addition to making sure validation errors disappear, this should also make sure they re-appear when they are removed because the field was hidden for a while.

## Related Issue(s)
- closes #617
- #296
- #628

## Verification
- [x] All tests run green

## Documentation

### DIFF
--- a/src/altinn-app-frontend/__mocks__/initialStateMock.ts
+++ b/src/altinn-app-frontend/__mocks__/initialStateMock.ts
@@ -39,7 +39,6 @@ export function getInitialStateMock(customStates?: Partial<IRuntimeState>): IRun
       validations: {},
       error: null,
       invalidDataTypes: [],
-      currentSingleFieldValidation: null,
     },
     instanceData: getInstanceDataStateMock(),
     instantiation: {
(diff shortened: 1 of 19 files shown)


bjosttveit (approved):
## LGTM

PR #633, opened 2022-11-08
Expressions: `any` type and shared tests sync

olemartinorg:
## Description
Syncing tests with backend. Adding the `in-nested-group2.json` test breaks, as `"14"` is not equal to `14`. The fundamental problem here was that the `component` function was defined as returning `string`, so types would be cast to string no matter what they were. Introducing an `any` type that avoids casting (which also fixes the peculiarity of the `if` function that required some arguments - and the return type - not to be cast).

## Related Issue(s)
- #355

## Verification
- [x] All tests run green

## Documentation

### DIFF
--- a/src/altinn-app-frontend/src/features/expressions/shared.test.ts
+++ b/src/altinn-app-frontend/src/features/expressions/shared.test.ts
@@ -48,7 +48,8 @@ describe('Expressions shared function tests', () => {
             return evalExpr(expr as Expression, component, dataSources);
           }).toThrow(expectsFailure);
         } else {
-          expect(evalExpr(expression, component, dataSources)).toEqual(expects);
+          const expr = asExpression(expression) as Expression;
+          expect(evalExpr(expr, component, dataSources)).toEqual(expects);
         }
       },
     );
--- a/src/altinn-app-frontend/src/features/expressions/validation.ts
+++ b/src/altinn-app-frontend/src/features/expressions/validation.ts
@@ -62,7 +62,7 @@ function validateFunctionArg(
   const expectedType = argTypeAt(func, idx);
   const actualType = actual[idx];
   if (expectedType === undefined) {
-    addError(ctx, [...path, `[${idx}]`], ValidationErrorMessage.ArgUnexpected);
+    addError(ctx, [...path, `[${idx + 1}]`], ValidationErrorMessage.ArgUnexpected);
   } else {
     const targetType = ExprTypes[expectedType];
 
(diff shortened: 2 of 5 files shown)


src/altinn-app-frontend/src/features/expressions/shared.test.ts
```diff
@@ -48,7 +48,8 @@ describe('Expressions shared function tests', () => {
             return evalExpr(expr as Expression, component, dataSources);
           }).toThrow(expectsFailure);
         } else {
-          expect(evalExpr(expression, component, dataSources)).toEqual(expects);
+          const expr = asExpression(expression) as Expression;
+          expect(evalExpr(expr, component, dataSources)).toEqual(expects);
```
olemartinorg, line 52: I'm doing this so that the test fails with a thrown exception in case any validation errors occur. They shouldn't, but it's nice to get that error instead of a silently working test. Should have been there all the time.

src/altinn-app-frontend/src/features/expressions/validation.ts
```diff
@@ -62,17 +62,17 @@ function validateFunctionArg(
...
 
     if (actualType === undefined) {
       if (targetType.nullable) {
         return;
       }
-      addError(ctx, [...path, `[${idx}]`], ValidationErrorMessage.ArgWrongType, expectedType, 'null');
+      addError(ctx, [...path, `[${idx + 1}]`], ValidationErrorMessage.ArgWrongType, expectedType, 'null');
     } else if (!targetType.accepts.includes(actualType)) {
-      addError(ctx, [...path, `[${idx}]`], ValidationErrorMessage.ArgWrongType, expectedType, 'null');
+      addError(ctx, [...path, `[${idx + 1}]`], ValidationErrorMessage.ArgWrongType, expectedType, 'null');
```
olemartinorg, line 75: The first value is always the function name, so this validation code misplaced the error message one value early. Fixing that at the same time.

src/altinn-app-frontend/src/features/expressions/validation.ts
```diff
@@ -84,18 +84,26 @@ function validateFunctionArgs(
...
+  }
+}
+
+function validateFunctionArgLength(
+  func: ExprFunction,
+  actual: (BaseValue | undefined)[],
+  ctx: ValidationContext,
+  path: string[],
+) {
+  const expected = ExprFunctions[func].args;
```
olemartinorg, line 99: I'm just separating out a function for validating the argument length, so that functions supplying their own validator (specifically the `if` function) can skip the length-check, but keep the argument type validator.

bjosttveit (approved):
## LGTM

PR #635, opened 2022-11-08
Expressions: Returning `null` when `component` is hidden

### LINKED ISSUE #630: Expressions: New `not` function, extend logic behind `component`
This issue is fairly easy to implement, and could be implemented before the first release of #540. Slightly discussed on Slack between me and @ivarne, we decided on implementing two additional features of the expressions engine:

1. A new `not` function. Accepts one boolean (or null) as `input`, returns `!input`.
2. Extend the logic in `component` to return `null` if the found component is hidden. This allows for some limited chaining, and makes the `component` functionally more different from the `dataModel` lookup (which won't have the same logic).

olemartinorg:
## Description
In the `component` lookup function, the returned value should be `null` if the component itself is hidden.

The reason this PR is so large for such a minor feature, is mostly because of the way expressions were resolved - they did not have all layouts (in a layout set) available at all times, so some code had to be refactored to make sure of that. Also, this leads to some complexity when factoring in that hiding an entire layout should hide all components within it - and hiding such a component  by hiding the entire layout should make the `component` lookup return null even if the component has a data model value. The added cypress test reflects one of the more complex use-cases this solves.

## Related Issue(s)
- closes #630

## Verification
- [x] All tests run green

## Documentation


bjosttveit (approved):
## LGTM 🥇

Issue #377, opened 2022-08-10
Architecture: Modularize components using classes and inheritance


olemartinorg:
### Description

There are multiple code paths which branch off to do something different depending on the encountered component type. This has historically been needed to implement some custom functionality for "abnormal" components.

### Solution

The idea in this issue is to refactor away some complexities in existing code such that we can have cleaner implementations of the generic frontend app, and at the same time concentrate custom overrides for each component in their respective folders. This should also make it easier to implement a new component, as potential overrides are centralized. One way to implement this is to define an abstract class (i.e. `BaseComponent`), have every component implement their own concrete class from this abstract base class, and then reference instances of these classes in the `components` map in `src/altinn-app-frontend/src/components/index.ts`. `GenericComponent.tsx` can then fetch each react component via `components[props.type].getReactComponent()` or similar.

#### Examples from `GenericComponent.tsx`

This case should instead be handled using a function call, like `!component.shouldRenderValidationsGenerically()` (quite possibly an overly verbose method name).

```typescript
const getValidationsForInternalHandling = () => {
  if (
    props.type === 'AddressComponent' ||
    props.type === 'DatePicker' ||
    props.type === 'FileUpload' ||
    props.type === 'FileUploadWithTag' ||
    (props.type === 'Likert' && props.layout === LayoutStyle.Table)
  ) {
    return componentValidations;
  }
  return null;
};

// some components handle their validations internally (i.e merge with internal validation state)
const internalComponentValidations = getValidationsForInternalHandling();
if (internalComponentValidations !== null) {
  passThroughProps.componentValidations = internalComponentValidations;
}
```

This case should instead be handled using a function call, like `component.shouldRenderLabel()`:

```typescript
const noLabelComponents = [
  'Header',
  'Paragraph',
  'Image',
  'Submit',
  'ThirdParty',
  'AddressComponent',
  'Button',
  'Checkboxes',
  'RadioButtons',
  'AttachmentList',
  'InstantiationButton',
  'NavigationBar',
  'Likert',
  'Panel',
];
```

#### Examples from `validation.ts`:

In `validateEmptyFieldsForLayout`, this code should instead look up the function to see if it has implemented any custom 'empty field' validations. Quite possibly every component type should have its own method to figure out if the formData is "empty" or not.

```typescript
if (
  component.type === 'FileUpload' ||
  component.type === 'FileUploadWithTag'
) {
  // These components have their own validation in validateFormComponents(). With data model bindings enabled for
[...]

Issue #288, opened 2021-06-16
Improve number field formatting


nkylstad:

StianVestli:
org/ssb

olemartinorg:
Very related:
- #717

PR #636, opened 2022-11-08
Date picker validation

### LINKED ISSUE #600: Datepicker: Clean up and modernize validations
### Description

The `Datepicker` component suffers from validation duplication, as it keeps validation messages both locally and in Redux. The function `validateDatepickerFormData` is called both in `validation.ts` and in `DatepickerComponent` - but we should aim to keep the validation state in redux only.

### Additional Information

- #185
- #98
- #377

### Tasks

- [x] Remove duplicate/local validation in `DatepickerComponent.tsx`, move missing validation into `validation.ts`
- [x] Implement delayed saving with `useDelayedSavedState` if useful
- [x] Make sure required field validation does not run immediately if field is empty (should only run when clicking `next`)
- [x] Make sure validation works when Datepicker is in a repeating group

### Acceptance Criterias

_No response_

bjosttveit:
## Description

The code in `DatepickerComponent` is now much simpler and no longer has any embedded validation logic. The behaviour is also changed so that it no longer arbitrarily removes the input when you type an invalid date. This works by storing the text-input instead of a timestamp when the date is invalid. This is fine since an invalid input string will never be parsed to a valid date (especially after specifying ISO_8601 explicitly when parsing). This will cause the validation code to interpret this as an invalid format.

- Removed internal validation logic from `DatepickerComponent`.
  - Had to update `runValidations` in `updateFormDataSagas` to run custom datepicker validations on handleDataChange.
- Added `useDelayedSavedState` to `DatepickerComponent` since typing manually would spam the server.
- Extended `useDelayedSavedState` to support `skipValidation`.
  - Refactored `useDelayedSavedState` to keep it clean.

## Related Issue(s)
- Closes #600 

## Verification
- [x] All tests run green

## Documentation


src/altinn-app-frontend/src/components/base/DatepickerComponent.test.tsx
```diff
@@ -97,133 +97,81 @@ describe('DatepickerComponent', () => {
...
 
-  it('should call handleDataChange with correct value when timeStamp is true when field is changed with a valid date', async () => {
+  it('should call handleDataChange with formatted value (timestamp=true) without skipping validation if date is valid', async () => {
     const handleDataChange = jest.fn();
     render({ handleDataChange, timeStamp: true });
 
     const inputField = screen.getByRole('textbox');
 
-    await userEvent.type(inputField, '12.26.2022');
+    await userEvent.type(inputField, '01012022');
```
olemartinorg, line 119: Maybe use the 31st of December as test data instead? That will catch if something goes wrong at some point switching the date/month. When the date and month is the same, these things might not get caught.

src/altinn-app-frontend/src/features/form/data/update/updateFormDataSagas.ts
```diff
@@ -86,6 +88,26 @@ function* runValidations(field: string, data: any, componentId: string | undefin
...
+      data,
+      flagBasedMinDate,
+      flagBasedMaxDate,
+      component.format,
+      state.language.language,
+    );
+    customComponentValidations = {
+      [fieldKey]: datepickerValidations,
+    };
+  }
```
olemartinorg, line 107: Even though I really hope this thing is temporary until we take a look at #377, maybe move this out to a separate function in `validation.ts`? You could grab the opportunity to name this particular flavor of validations (are they immediate/instant/delayed?). The way I see it, validation hooks could possibly occur at several points in time:
1. When the user types/fills out something (possibly filtering/changing the value, such as with number formatting fields only accepting numbers? This should in turn give the user some kind of feedback - possibly through a short-lived validation message - see #288)
2. When the user has stopped typing (aka `useDelayedSaveState`), before/after the data is saved to the server
3. If the field is set up with `"triggers": ["validation"]` it should also ask the server for any validation messages(?) at this point
4. When page validation (or all-page validation) runs, all the previous validations should run, and in addition any required fields should produce validation messages if not properly filled out.

olemartinorg (approved):
Great job! :godmode:

src/altinn-app-frontend/src/components/base/DatepickerComponent.test.tsx
bjosttveit: Yeah, I think the reason I didn't was that it looks that the format used comes from moment.locale, and this may not be 100% predictable? I guess i can make sure that the format is specified explicitly, and then use a day > 12

bjosttveit: I do assume the format in other tests anyways so should probably fix

src/altinn-app-frontend/src/features/form/data/update/updateFormDataSagas.ts
bjosttveit: Yeah I was also hoping this would be temporary 😅 not super happy about essentially duplicating this bit. I'll refactor a bit.

src/altinn-app-frontend/src/components/base/DatepickerComponent.test.tsx
olemartinorg: Maybe we could use `moment.locale()` in `setupTests.ts` to configure this predictably? :thinking: [See this tip](https://stackoverflow.com/questions/17493309/how-do-i-change-the-language-of-moment-js).

olemartinorg (approved):

src/altinn-app-frontend/src/features/form/data/update/updateFormDataSagas.ts
```diff
@@ -91,19 +90,7 @@ function* runValidations(field: string, data: any, componentId: string | undefin
...
-      flagBasedMinDate,
-      flagBasedMaxDate,
-      component.format,
-      state.language.language,
-    );
-    customComponentValidations = {
-      [fieldKey]: datepickerValidations,
-    };
+    customComponentValidations = validateDatepickerFormData(data, component, state.language.language);
   }
```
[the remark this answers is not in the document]
olemartinorg: That's why I said it could be up to you to name this new(?) type of validation, as it does not include all the same validations as the function that already exists in `validation.ts` (for FileUpload*), so it must be 'something else'.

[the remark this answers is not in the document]
bjosttveit: Ahh I see

olemartinorg (approved):

PR #638, opened 2022-11-09
List component

### LINKED ISSUE #578: New component: List
### Description

After a refinement meeting for #107, we found that some specific of the needs can be isolated into a separate (new) studio/app-frontend layout component: **List** (see my comment in #543)

Visually this list will be presented as a table, but functionally it can either be just a simple static list of data OR a single/multi-select using either a radio or checkbox for selecting one or more rows from the list. Data displayed in this table does not come from the data model, but the List component can populate the data model if rows are selectable (so unlike repeating groups, the data model bindings for this component does not describe where it gets the data from, but rather where the selected data is put).

Under the hood, this should use the new Table component from the design system for the visual representation.

### Functional requirements
_Tasks moved to #638, to make it possible for external developers to check items along the way_

### Additional Information

- #107 
- #543

caroliss:
## Description
First version of listcomponent with sorting and pagination. 

## Related Issue(s)
- closes #578

### Functional requirements (from #578)
- [ ] The component can be configured with a static data source, much like existing options/code lists.
- [x] The component can be configured with a dynamic data source (fetched from backend). The source data must have a logical structure (i.e. an array of objects, where the object keys indicate the column name, and the object values refer to the cell content/text)
- [x] The List component can be configured to use pagination. In this case, if the data source is fetching from the backend, `app-frontend-react` only keeps the current page of data in memory, and re-fetches the data when navigating between pages.
- [x] Columns in the component can be configured as sortable. If data is fetched from backend, the sort column and direction is passed to the backend via query parameters such that it can handle the sorting implementation. 
- [ ] For static data defined in the layout file, sorting can only work on alphabetically sorting strings.
- [x] A row in the List component can be configured as being selectable. This means a radio button will show up as the first column in the table view, and clicking that radio will use the data model bindings to put one or more cells of data from the list data into the data model.
- [ ] A row in the List component can be configured to as behaving as a multi-select. This means a checkbox will show up as the first column in the table view. If this is used instead of a radio button, the data model binding should be interpreted as a `list` data model binding (as used for attachments), i.e. it will store data in an array. It might also be required to interpret the data model binding as a `group` binding (i.e. prefilling/populating a repeating group structure).  
- [ ] Some columns should be possible to hide. This can be useful when you want the backend to send an internal reference number you want to populate into the data model when selecting a row, but you don't want to show that column/data to the user.
- [x] The JsonSchema definition for the column definition in the List component should be as similar to the column definition for repeating groups as possible, to avoid having multiple ways to describe the visual properties of tables.
- [ ] The List component should work nicely on mobile
- [x] Keyboard navigation
- [ ] Screen reader support
- [x] Configuration options for which value(s) in the data model bindings should be used when presenting the chosen row(s) in a Summary component referencing the List
- [ ] Support for the component in repeating groups, using the Summary value in the table header
- [ ] Support for 'required'
- [ ] Proper/intuitive validation messages when no row is selected
- [ ] Support for 'readOnly'?
- [ ] Add query parameter for search connected to a data model parameter (aka 'mapping')
- [ ] Bugfix: Handlechange is called in a loop,  sometimes causing race condition when you have multiple datamodel bindings. Sometimes only one of the datamodelbindings are saved correctly. 

### Other tasks (from #578)
- [x] Define a data structure to be used to describe the data/rows, and implement support for this in [Altinn/app-lib-dotnet](/Altinn/app-lib-dotnet)
- [ ] Create issue(s) for implementing support for this component in Altinn Studio
- [x] Automatic tests (preferably by adding to our Cypress test suite)
- [ ] Documentation


src/altinn-app-frontend/src/features/form/layout/index.ts
```diff
@@ -99,6 +99,17 @@ export interface ILayoutCompDatePicker extends ILayoutCompBase<'DatePicker'> {
...
 export type ILayoutCompDropdown = ILayoutCompBase<'Dropdown'> & ISelectionComponent;
+export interface ILayoutCompList extends ILayoutCompBase<'List'> {
+  tableHeaders?: string[];
+  sortableColumns?: string[];
+  pagination: IPagination;
+  dataList?: IDataList[];
+  dataListId?: string;
+  mapping?: IMapping;
+  secure?: boolean;
+  source?: IOptionSource;
```
olemartinorg, line 110: Could you add the component to `schemas/json/layout/layout.schema.v1.json` as well? :pray: In JsonSchema it's also possible to require either `dataList` OR `dataListId`

src/altinn-app-frontend/src/features/form/layout/index.ts
```diff
@@ -99,6 +99,17 @@ export interface ILayoutCompDatePicker extends ILayoutCompBase<'DatePicker'> {
...
+  fieldToStoreInDataModel: string;
```
olemartinorg, line 111: Not quite sure I agree with this one. I'd prefer to set up the component with `dataModelBindings` which is what we use for these things right now:
```
  "dataModelBindings": {
    "simpleBinding": "My.Model.Path"
  }
```
In this case we'd need a source -> target mapping, so I suggest trying to work with this format:
```
  "dataModelBindings": {
    "myTableColumn": "My.Model.Path",
    "case_number": "My.Model.CaseNumber"
  }
```
Where it would be possible to map multiple table columns in the chosen row to the data model.

src/altinn-app-frontend/src/components/base/ListComponent.tsx
```diff
@@ -0,0 +1,177 @@
...
+  handleDataChange,
+  sortableColumns,
+}: IListProps) => {
+  const dynamicDataList = useGetDataList({ dataListId, mapping });
+  const calculatedDataList = dynamicDataList || defaultDataList;
+
+  const rowsPerPage = useAppSelector(
+    (state) => state.dataListState.dataLists[dataListId || ''].size || pagination.default,
+  );
+  const currentPage = useAppSelector((state) => state.dataListState.dataLists[dataListId || ''].pageNumber || 0);
```
olemartinorg, line 43: Have you tested this with multiple `List` components each using the same `dataListId` on the same page? I suspect they will be linked, but it should be possible to use multiple components mapped to the same `dataListId` individually. My suggestion would be to use the `id` property passed to the component as the key in the redux state instead of `dataListId`.

src/altinn-app-frontend/src/shared/resources/dataLists/dataListsSlice.ts
```diff
@@ -0,0 +1,93 @@
...
+  ISetDataListsPageSize,
+  ISetDataListsWithIndexIndicators,
+  ISetSort,
+} from 'src/shared/resources/dataLists';
+import type { MkActionType } from 'src/shared/resources/utils/sagaSlice';
+
+const initialState: IDataListsState = {
+  dataLists: {},
+  dataListsWithIndexIndicator: [],
+  error: null,
```
olemartinorg, line 19: Are you using the error anywhere? See `src/selectors/getErrors.ts`, you should probably add your error state there as well. It's not really optimal either, but at least we react to an error when fetching the list data instead of rejecting and silently ignoring the problem.

src/altinn-app-frontend/src/shared/resources/dataLists/fetchDataListsSaga.ts
```diff
@@ -0,0 +1,133 @@
...
+      id: dataListId,
+      mapping: dataMapping,
+      secure,
+    };
+    yield put(dataListsActions.fetching({ key, metaData }));
+    const formData: IFormData = yield select(formDataSelector);
+    const language = yield select(appLanguageStateSelector);
+    const dataList = yield select(listStateSelector);
+
+    const pageSize = dataList.dataLists[dataListId].size ? dataList.dataLists[dataListId].size.toString() : '5';
```
olemartinorg, line 98: Does this mean the size is 5 items by default for all data lists, regardless of the component setup? :thinking: Isn't that number configurable using the `pagination` property? And if it's not set, I guess all items should be fetched, not just 5.

src/altinn-app-frontend/src/types/index.ts
```diff
@@ -276,6 +307,15 @@ export interface IFetchSpecificOptionSaga {
...
 }
 
+export interface IFetchSpecificDataListSaga {
+  dataListId: string;
+  formData?: IFormData;
+  language?: string;
+  dataMapping?: IMapping;
+  secure?: boolean;
+  instanceId?: string;
+}
```
olemartinorg, line 317: I think most of the types defined in this file should probably live closer to their implementations, right? Try putting them somewhere like `shared/resources/dataLists/index.d.ts` instead of here.

src/altinn-app-frontend/src/utils/appUrlHelper.ts
```diff
@@ -186,6 +188,70 @@ export const getOptionsUrl = ({
...
+    params.language = language;
+  }
+
+  if (pageSize) {
+    params.size = pageSize;
+  }
+
+  if (pageNumber) {
+    params.page = pageNumber;
+  }
```
olemartinorg, line 246: What if the page number is `0`? Might need to check if `pageNumber !== undefined` on this one

olemartinorg (commented):
Great work so far!! :partying_face: This is looking really good!

Do you have a test-app running where you're testing out this component? I'd love to try it out!

Also, it would be extra cool if you could take a look at creating a Cypress test for the component. That would really cement the functionality you need so that we ensure it continues to work end-to-end when this is merged in, and will remain working even after future changes in `app-frontend-react`.  The basic idea is to create a new test file in `test/cypress/e2e/integration/app-frontend`, extend the [frontend-test](https://dev.altinn.studio/repos/ttd/frontend-test) app by adding your component there, and try running the setup locally (there is a readme in the cypress directory for how to set it up). The more configuration and use-cases you test using Cypress, the better! You also don't have to create many instances of your component in the layout files to test different variations, as cypress can intercept a layout and modify it (see `cy.interceptLayout`) in-flight to test different configurations.

caroliss:
> Great work so far!! 🥳 This is looking really good!
> 
> Do you have a test-app running where you're testing out this component? I'd love to try it out!

@olemartinorg right now we dont have published a test app with this component, because we need to use local app-lib-dotnet. To test our app locally you also need to run the api we use to get cases locally. When the PR in app-lib-dotnet is merged it will be possible to test I guess.

olemartinorg:
@caroliss Allright - makes sense that publishing it on tt02 is out of the question right now, but do you have an app I can test locally if I set up the environment needed for it? I'd love to try it out!

olemartinorg:
[Your changes in the design system](https://github.com/Altinn/altinn-design-system/pull/207) has been released in `v0.24.0`, so when #686 has been merged I suggest you pull from our latest main in order to get the latest changes. :smiley:

src/altinn-app-frontend/src/shared/resources/dataLists/index.d.ts
```diff
@@ -47,3 +47,46 @@ export interface ISetSort {
...
+
+export interface IFetchSpecificDataListSaga {
+  id: string;
+  dataListId: string;
+  formData?: IFormData;
+  language?: string;
+  dataMapping?: IMapping;
+  secure?: boolean;
+  instanceId?: string;
+  paginationDefaultValue?: number;
```
olemartinorg, line 91: `formData`, `language` and `instanceId` seem to never be read (or set) here, and they seem a bit misplaced to me.

src/altinn-app-frontend/src/shared/resources/dataLists/index.d.ts
```diff
@@ -47,3 +47,46 @@ export interface ISetSort {
...
+  listItems: any;
+  _metaData: IDataListPaginationData;
+}
+
+export interface IDataLists {
+  [key: string]: IDataListData;
+}
+
+export interface IDataListActualData {
+  listItems: any;
```
olemartinorg, line 61: I think a bit more specific type could be `Record<string, string>[]`, i.e. a list of objects with string keys and string values.

src/altinn-app-frontend/src/shared/resources/dataLists/index.d.ts
```diff
@@ -47,3 +47,46 @@ export interface ISetSort {
   sortColumn: string;
   sortDirection: SortDirection;
 }
+
+export interface IDataList {
+  listItems: any;
```
olemartinorg, line 52: Possibly `Record<string, string>[]` here as well?

olemartinorg (changes requested):
Great work! :partying_face:

src/altinn-app-frontend/src/components/base/ListComponent.tsx
```diff
@@ -26,110 +27,94 @@ const defaultDataList: any[] = [];
...
 
   const checkSortableColumns = (headers) => {
     const cell: JSX.Element[] = [];
-    let index = 0;
     for (const header of headers) {
       if ((sortableColumns || []).includes(header)) {
         cell.push(
           <TableCell
             onChange={handleSortChange}
             id={header}
```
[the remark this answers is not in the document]
johlie: The id is just passed to TableCell in the designsystem. The id is supposed to be sat on the sorteble columns in case there is more than one sorteble column. Then only one of the columns is sorted a time.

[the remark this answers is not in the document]
olemartinorg: Ahh, right, I see that now. Maybe a bit unfortunate to mix those terms, as `id` is a well-recognized html attribute as well. I'd suggest renaming it to `sortKey` or something - but it's not really a big deal. :blush:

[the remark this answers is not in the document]
johlie: I can change it :)

olemartinorg (changes requested):
Good progress! :rocket: I'll merge the latest changes you made to the design system into main, so that it'll work out of the box without linking in the design system. Also, don't forget the `layout.schema.v1.json` file. :smiley:

olemartinorg:
Would you mind merging from our main now that #711 is done, so that we can run automated tests and prepare for merging these changes? :partying_face:

schemas/json/layout/layout.schema.v1.json
```diff
@@ -117,29 +143,263 @@
...
+            "properties": {
+              "type": {
+                "const": "AddressComponent"
+              }
+            }
+          },
+          "then": {
+            "$ref": "#/definitions/addressComponent"
+          }
+        },
```
olemartinorg, line 162: Looks like you formatted the json file with some kind of automated formatter - making this file suddenly very complicated to read. Could you revert the styling/formatting changes so that we can isolate the actual changes in the diffs? :pray:

src/altinn-app-frontend/src/components/base/ListComponent.tsx
```diff
@@ -127,31 +118,59 @@ export const ListComponent = ({
...
       chosenRowData[key] = datalist[key];
     }
     return chosenRowData;
   };
+  const rowAsValueString = (datalist) => {
+    const chosenRowData: rowValue = {};
+    for (const key in dataModelBindings) {
+      chosenRowData[key] = datalist[key];
+    }
+    return JSON.stringify(chosenRowData);
```
olemartinorg, line 133:
```suggestion
    return JSON.stringify(rowAsValue(datalist));
```

src/altinn-app-frontend/src/components/base/ListComponent.tsx
```diff
@@ -127,31 +118,59 @@ export const ListComponent = ({
...
-              key={datalist}
-              rowData={FindRowDataValue(datalist)}
+              key={JSON.stringify(datalist)}
+              rowData={rowAsValue(datalist)}
             >
+              <TableCell radiobutton={true}>
+                <RadioButton
+                  name={datalist}
+                  // eslint-disable-next-line @typescript-eslint/no-empty-function
+                  onChange={() => {}}
```
olemartinorg, line 167: I think this should work, without having to disable eslint?
```suggestion
                  onChange={() => {
                    // Intentionally empty to prevent double-selection
                  }}
```

olemartinorg (changes requested):

olemartinorg (approved):
:1st_place_medal:

PR #651, opened 2022-11-14
Cypress: Using go-to-task button

olemartinorg:
## Description
Restructuring the cypress code to clean up all the `navigateTo*()` and `completeTaskXForm()` functions. Combining all this functionality into common ones, and adding support for short-circuiting a task and skipping straight to the next one without really filling out any form fields. This should make running some tests a bit faster, and possibly also reduce failures in one early task to spill over into failing other tests that relied on completing an earlier task before getting to their own task.

Also:
- Removing some unused cypress functions
- Making `interceptLayout` intercept only once, not requiring a return value from it, and making it possible to alter an entire layout set. 
- Adding a `reloadAndWait` function for the `group.js` test
- Removing a useless test that measured the width of the 'send in' button

Please note: I think there is a minor race condition in the code introduced in #488. I could sometimes get the tests to fail because the request to save the data model happens _after_ requesting to navigate to the next task. I'm hoping I can spend more time fuzzing that one out and fixing the root cause of it.

## Related Issue(s)
- #488

## Verification
- [x] All tests run green

## Documentation

### DIFF
--- a/test/cypress/e2e/integration/app-frontend/message.js
+++ b/test/cypress/e2e/integration/app-frontend/message.js
@@ -47,12 +47,6 @@ describe('Message', () => {
       cy.get(appFrontend.startAgain).contains(instanceId).and('contain.text', appFrontend.apps.frontendTest);
       cy.get(appFrontend.startAgain).find('a').should('have.attr', 'href', instantiateUrl);
     });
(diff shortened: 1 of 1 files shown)


bjosttveit (approved):
# LGTM 🔥

PR #657, opened 2022-11-15
Fixing inconsistent prev/next navigation buttons

olemartinorg:
## Description
I've observed that the previous/next navigation buttons sometimes mysteriously disappeared, and I was about to create a bug report for it. However, to figure out if it was just my config that was wrong, I looked into it and quickly found the cause behind the problem - these buttons seem to somewhat rely on a `navigation` property set in the layout file (which is an undocumented feature?!). I'm a bit confused about what was supposed to happen here, so I'm opening a draft PR to get some feedback. These changes seems to work in the same way the documentation describes, though.

## Related
- Documentation: https://docs.altinn.studio/nb/app/development/ux/pages/navigation/

## Verification
- [ ] All tests run green

## Documentation


olemartinorg:
Still not sure about this one. I think I have to go sift through the code to fully understand how these things are expected to interact with implicitly added submit button(s), etc.

PR #658, opened 2022-11-15
Bugfix: Missing 'required' property on component was interpreted as 'true'

olemartinorg:
## Description
Test case added by removing a `"required": false` in `frontend-test`. 

## Related Issue(s)
- #{issue number}

## Verification
- [ ] **Your** code builds clean without any errors or warnings
- [ ] Manual testing done (required)
- [ ] Relevant automated test added (if you find this hard, leave it and we'll help out)
- [ ] All tests run green

## Documentation
- [ ] User documentation is updated with a separate linked PR in [altinn-studio-docs.](https://github.com/Altinn/altinn-studio-docs) (if applicable)

PR #660, opened 2022-11-15
Cypress-runner

olemartinorg:
## Description
I'm closing this PR without merging it later, I just need some way to run the cypress tests after #658 and #659 had some strange failing tests. I could run the tests locally, but for some reason I could not get the same failures there. Also, I just deployed a new version of `frontend-test` to tt02 with reproductions of the layout config that caused the aforementioned bugs.

## Related Issue(s)
- #658 
- #659

## Verification
- [ ] All tests run green

## Documentation


olemartinorg:
Closing this one. It seems the recent flaky tests are caused by #651, and specifically the problem I mentioned in that issue.

Most of the time, pressing the `go-to-task` button will save the current data model to the backend _before_ switching to the requested task - but sometimes the request happens _after_ the request to navigate to another task. Also, pressing the button does not seem to run all validations (like any other such button would do), so if the current data model is not valid the server will produce an error that crashes the frontend.

PR #673, opened 2022-11-17
Removing `go-to-task` trick in Cypress tests

### LINKED ISSUE #667: Button with `go-to-task` has a race-condition and can crash the app
### Description of the bug

This describes two problems:

#1: Some cypress tests fail sometimes, because the request to save the data model happens _after_ the request to navigate to the target process (race condition?)

#2: Pressing the button does not seem to validate the current data model when going to the next page, which can cause the whole app to crash. Not sure we can fix this, as going back to the previous task should work fine regardless of the current data model validity (and we can't possibly know if we should validate or not if we don't what's the previous/next task - as the process engine is implemented backend). This could possibly be solved using the regular submit button when going forwards.

### Steps To Reproduce

For problem #1:
See related issues (Cypress reproduced this all the time, but as with all race-conditions, it's not always 

For problem #2:
Try clicking the button in `frontend-test` before making any changes to the form

### Additional Information

- #651
- #660 
- #488

olemartinorg:
## Description
Hopefully fixing #667 by injecting a redux action instead of relying on a request to be intercepted. This _might_ need chagnes in the `frontend-test` app, but I'll let the tests run first without any changes to see if it works without.

## Related Issue(s)
- closes #667

## Verification
- Manual testing
- Automated tests
  - [ ] Has been updated
   
  - [x] No changes made
- Support in Altinn Studio
   
- Sprint board

### DIFF
--- a/test/cypress/e2e/support/app-frontend.js
+++ b/test/cypress/e2e/support/app-frontend.js
@@ -54,6 +54,13 @@ Cypress.Commands.add('getReduxState', (selector) => {
     });
 });
 
+Cypress.Commands.add('reduxDispatch', (action) => {
+  return cy
+    .window()
+    .its('reduxStore')
+    .invoke('dispatch', action);
+});
+
 Cypress.Commands.add('interceptLayout', (layoutName, mutator, wholeLayoutMutator) => {
   cy.intercept({ method: 'GET', url: `**/api/layouts/${layoutName}`, times: 1 }, (req) => {
     req.reply((res) => {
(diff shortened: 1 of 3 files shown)


bjosttveit:
Is this necessary to make tests run after this commit?
https://dev.altinn.studio/repos/ttd/frontend-test/commit/b4161e41440d113fb11fe96347dfcee043a1cd62

bjosttveit (approved):

olemartinorg:
@bjosttveit Yep, it seems like it is. The tests certainly did not work well before I deployed those changes to tt02.

PR #682, opened 2022-11-21
Give server rowIndex for validateRow trigger

bjosttveit:
## Description

- When using `triggers=["validateRow"]` the rowIndex will now be sent to the server for validations. Docs have been updated to reflect this.
- No longer filters out server-validations not related to the current row on validateRow. This was the behaviour only a few days ago, and has been changed back since it is totally legitimate to add validation errors outside of the current group row. The developer is now expected to use the rowIndex in servervalidations to deal with this if necessary. Not really a breaking change, as the validateRow trigger is basically brand new.
- Previously, the group validation trigger would not dispatch an update to validations unless there were new errors present. This was a problem because the server could send back validations that had been fixed, and they would not be updated if there were no new errors. This has been changed so that it always dispatches an update to validations. Also should not be a breaking change, since dispatching no new errors should not make a difference, and no existing tests failed 🤷 

## Related Issue(s)
- [Slack-thread](https://altinn.slack.com/archives/C02ESDSGPN1/p1669016064881859)

## Verification
- Manual testing
- Automated tests
  - [x] Has been updated
    - https://github.com/Altinn/altinn-studio-docs/pull/754
   
  - [x] No changes made
- Support in Altinn Studio
   
- Sprint board

### DIFF
--- a/src/altinn-app-frontend/src/features/form/layout/update/updateFormLayoutSagas.ts
+++ b/src/altinn-app-frontend/src/features/form/layout/update/updateFormLayoutSagas.ts
@@ -29,6 +29,7 @@ import {
   splitDashedKey,
 } from 'src/utils/formLayout';
 import { getLayoutsetForDataElement } from 'src/utils/layout';
+import { nodesInLayouts } from 'src/utils/layout/hierarchy';
 import { getOptionLookupKey, removeGroupOptionsByIndex } from 'src/utils/options';
 import {
   canFormBeSaved,
@@ -510,11 +511,12 @@ export function* updateRepeatingGroupEditIndexSaga({
   payload: { group, index, validate },
 }: PayloadAction<IUpdateRepeatingGroupsEditIndex>): SagaIterator {
   try {
-    if (validate) {
-      const state: IRuntimeState = yield select();
+    const state: IRuntimeState = yield select();
+    const rowIndex = state.formLayout.uiConfig.repeatingGroups?.[group].editIndex;
+
+    if (validate && typeof rowIndex === 'number' && rowIndex > -1) {
       const validations: IValidations = state.formValidations.validations;
       const currentView = state.formLayout.uiConfig.currentView;
-      const rowIndex = state.formLayout.uiConfig.repeatingGroups?.[group].editIndex;
 
       const frontendValidations: IValidations = validateGroup(
         group,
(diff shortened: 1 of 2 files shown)


olemartinorg (approved):

ivarne:
Can’t there be multiple row indexes? Like in a nested group?

bjosttveit:
> Can’t there be multiple row indexes? Like in a nested group?

Right, I suppose you do need the parent's row-index as well if a nested group has the trigger

src/altinn-app-frontend/src/features/form/layout/update/updateFormLayoutSagas.ts
```diff
@@ -510,22 +511,29 @@ export function* updateRepeatingGroupEditIndexSaga({
...
         group,
         state,
         validate === Triggers.ValidateRow ? rowIndex : undefined,
       );
 
+      // Get group's rowIndices to send to server for validations
+      const nodes = nodesInLayouts(state.formLayout.layouts, currentView, state.formLayout.uiConfig.repeatingGroups);
+      const groupNode = nodes.findById(group);
+      const rowIndices = groupNode?.rowIndices();
+      rowIndices?.push(rowIndex);
```
olemartinorg, line 531: The `group` variables already contains the index (embedded inside the string). You can extract it by using the `splitDashedKey` function without constructing the entire layout.

src/altinn-app-frontend/src/utils/layout/hierarchy.ts
```diff
@@ -426,6 +426,21 @@ export class LayoutNode<NT extends NodeType = 'unresolved', Item extends AnyItem
     return parents;
   }
 
+  /**
+   * Returns a list of rowIndices, sorted by most senior parent's rowIndex first
+   * Useful for accessing the datamodel related to a nested component
```
olemartinorg, line 431: Not really sure if we want to encourage that.. :wink: That's why `dataModelBindings` have already had their indexes injected when in a node hierarchy - we should encourage use of that directly. I'm not really sure we need this function at all, since (as mentioned) the row index already exists as part of the `id` property (although that is a risky design choice in itself, I think).

olemartinorg (approved):

src/altinn-app-frontend/src/features/form/layout/update/updateFormLayoutSagas.ts
bjosttveit: Ait nice, did not realize this function existed. It does have the potential for introducing weird bugs though since the `componentId` can be anything at all, but I suppose that case is rare enough. [This issue](https://github.com/Altinn/app-frontend-react/issues/345) keeps coming to mind when working with validations 😉

src/altinn-app-frontend/src/utils/layout/hierarchy.ts
bjosttveit: Agree that that particular design choice is risky, but I guess we will keep it for the time being

src/altinn-app-frontend/src/features/form/layout/update/updateFormLayoutSagas.ts
olemartinorg: Yep - but it's been that way for a long time. We should absolutely get away from it long-term, but short-term I think we should consider running a [sanity check](https://github.com/Altinn/app-frontend-react/issues/648) and warn if component IDs can be mis-interpreted as containing row indexes. (See the task starting with _As long as app-frontend-react puts the repeating group indexes into the component ID_ in the linked issue)

PR #694, opened 2022-11-24
fix validation issues for oneOf(ref, null) structure

### LINKED ISSUE #693: Issues with jsonschema validation on app frontend (new datamodels)
## Description
When XSD has the `xsd:nillable` attribute set, the data model structure becomes 
```json
"oneOf": [
{
  // The actual schema, or reference to it
},
{
  "type": "null"
}
]
```
This is the correct way to set up a schema for a field that is nullable. However it is causing issues in app-frontend validation, since all of the following validation errors are triggered by invalid data against the schema:
- Actual error against the actual schema, f.eks. format, type, etc.
- Error against the second (null) schema
- Error agains the oneOf structure, since it does not "match just one of the schemas exactly".

This results in 3 different error messages in app frontend for the 1 error.

## Screenshots

Image

## Solution
We ignore the oneOf validation messages. This is not something that was actually supported properly in app frontend before, so ignoring those messages should not cause any issues. In addition, we ignore the `null` schema error message. We do not have `null` values in form data in app frontend. We only show the validation error on the actual subschema that is failing, which is what is expected in this case.

nkylstad:
## Description
Updated validation code to handle the `oneOf` structure that is used for nillable data in new data modelling tool.
We are looking into whether we can fix this another way, but for the time being this is the way a lot of data models generated with the new tool are going to look, and we need to support that.

Also updated code to handle the new format-keywords that were recently introduced in https://github.com/Altinn/altinn-studio/issues/9113
- as part of this, tried making the code robust so that it doesn't fail/crash if a new keyword that is not implemented yet is used.

## Related Issue(s)
- closes #693

## Verification
- Manual testing
- Automated tests
  - [x] Unit test(s) have been added
  - [ ] Has been updated
   
  - [x] No changes made
- Support in Altinn Studio
   https://github.com/Altinn/altinn-studio/issues/9314
- Sprint board


src/altinn-app-frontend/src/utils/validation/validation.test.ts
```diff
@@ -1385,6 +1397,36 @@ describe('utils > validation', () => {
       );
       expect(mockResult).toEqual({ invalidDataTypes: false, validations: {} });
     });
+
+    it('should handle oneOf/null structure when data is invalid', () => {
+      const mockValidator = validation.createValidator(mockJsonSchema);
+      const useFormData = {
+        dataModelField_1: 5,
+        dataModelField_2: 'Hello',
+        dataModelField_3: 'Test',
```
olemartinorg, line 1406: Maybe add a test case for when `dataModelField_2` is `null` as well?

olemartinorg (approved):

olemartinorg (approved):

PR #714, opened 2022-12-06
Enable expressions for TextResourceBindings and per row in repeating groups

Magnusrm:
## Description

Enables expressions to be used for TextResourceBindings
Enables expressions to be calculated using row-specific values when useExpressions is given a rowIndex in a group component context.

The use of TextResourceBindings have been updated to run expressions using `useExpressions` before being taken into use by components.

`useExpressions` have been updated to check whether the given component is a group, and if a rowIndex is given the expression will run using a the context of a child-component in the given index.

## Related Issue(s)
- #61 
- #64 

## Verification
- Manual testing
- Automated tests
  - [x] Has been updated
   
  - [ ] No changes made
- Support in Altinn Studio
   
- Sprint board


bjosttveit:
Looking forward to all of the merge conflicts i'm gonna get 😩

src/features/form/containers/RepeatingGroupTableRow.tsx
```diff
@@ -0,0 +1,374 @@
+import React, { useEffect } from 'react';
+
+import { Button, ButtonColor, ButtonVariant, TableRow } from '@altinn/altinn-design-system';
+import { createTheme, makeStyles, TableCell, useMediaQuery } from '@material-ui/core';
```
bjosttveit, line 4: Use `TableCell` from design-system instead of mui.

```suggestion
import { Button, ButtonColor, ButtonVariant, TableCell, TableRow } from '@altinn/altinn-design-system';
import { createTheme, makeStyles, useMediaQuery } from '@material-ui/core';
```

cypress/e2e/integration/app-frontend/group.js
```diff
@@ -245,6 +245,29 @@ describe('Group', () => {
...
+          .find(appFrontend.group.edit)
+          .should('be.visible')
+          .should('have.text', texts.edit);
+        cy.get(table).children().eq(1).find(appFrontend.group.delete).should('not.exist');
+        cy.get(table)
+          .children()
+          .eq(1)
+          .find(appFrontend.group.edit)
+          .should('be.visible')
+          .should('have.text', texts.view);
```
olemartinorg, line 269:
```suggestion
        cy.wrap(table).children().eq(0).find(appFrontend.group.delete).should('be.visible');
        cy.wrap(table)
          .children()
          .eq(0)
          .find(appFrontend.group.edit)
          .should('be.visible')
          .should('have.text', texts.edit);
        cy.wrap(table).children().eq(1).find(appFrontend.group.delete).should('not.exist');
        cy.wrap(table)
          .children()
          .eq(1)
          .find(appFrontend.group.edit)
          .should('be.visible')
          .should('have.text', texts.view);
```

This will make sure it works cleanly with typescript after #735 :partying_face:. It turns out we've been using cypress wrong all along; we were meant to use the `wrap` function inside a `then`, not `get`.

src/features/expressions/index.ts
```diff
@@ -546,6 +546,16 @@ export const ExprDefaultsForComponent: ExprDefaultValues<ILayoutComponent> = {
...
   hidden: false,
+  textResourceBindings: {
+    title: '',
+    tableTitle: '',
+    description: '',
+    body: '',
+    next: '',
+    back: '',
+    altTextImg: '',
+    help: '',
```
olemartinorg, line 557: I guess this works in a pinch, but this would cause a failing expression in `title` to produce a warning and default to an empty string, while a failing expression in `shortName` would cause a full app crash. I think this warrants a possible alternative to `useExpression` that provides a default for every possible text resource binding. I can help with that, if you'd like.

src/features/expressions/useExpressions.ts
```diff
@@ -52,17 +58,21 @@ export function useExpressions<T>(input: T, _options?: UseExpressionsOptions<T>)
...
   const instanceContext: IInstanceContext = useAppSelector(instanceContextSelector);
   const id = (options && options.forComponentId) || component.id;
+  const rowIndex: number | undefined = options && options.rowIndex;
 
   const node = useMemo(() => {
     if (id) {
       const foundNode = nodes.findById(id);
+      if (typeof rowIndex == 'number' && foundNode && foundNode.item.type == 'Group') {
+        return foundNode.children(undefined, rowIndex)[0];
+      }
```
olemartinorg, line 68: Glorious!! :star_struck: Simple, easy and efficient. This is much better than the solution I was thinking of.

olemartinorg (approved):
:partying_face:

src/features/expressions/index.ts
Magnusrm: Yes, i was trying to find all the text resource bindings that are being used, but i guess i missed some 😅 As for how we should handle the values, do you think we need to consider language selection? Or would it be fine to just use English? If we need to use the selected language i guess we will need to find a good place to resolve the texts. Is that what you mean with an alternative to useExpression? Help would be nice 😄

olemartinorg: I'll look into it ASAP! And no, I don't think we should complicate it by actually looking up the language keys in text resources. Instead of describing in detail what I'm envisioning, I'll just commit to this branch. :wink:

olemartinorg: Turned out a little differently (no need for using anything other than `useExpressions()` after all). Pushed now.

Issue #603, opened 2022-11-02
Repeating groups/Grid/List: Column configuration options for table 


olemartinorg:
We'll soon have 3 components displayed as a table, those being [repeating groups](https://github.com/Altinn/app-frontend-react/issues/580), [Grid](https://github.com/Altinn/app-frontend-react/issues/583), and [List](https://github.com/Altinn/app-frontend-react/issues/578). All of these components could benefit from using a common configuration object for how to display columns. This includes support for:

- [ ] Changing the label for a given column (as an alternative to #68)
- [ ] Changing the width of the column to:
  - A fixed width (unless that interferes too much with a responsive design)
  - A fixed percentage width (where a combined 100% still reserves room for the buttons)
  - Auto width (the default)
- [ ] Controlling the text alignment in a column
- [ ] Controlling text overflow, using:
  - Ellipsis without line-wrapping (the default)
  - Line-wrapping and no overflow limitations
  - Line-wrapping with a max height (ellipsis or fade-out if the text overflows the max height)

StianVestli:
org/ssb

andsjo-svv:
org/svv

Magnusrm:
Options:
`"width"`: Sets width of column to a fixed percentage, ex. `"25%"` if the space is available or `"auto"`, which equally distributes remaining width that is not set yet among columns with `"auto"`. Default: `"auto"`.
`"alignText"`: Sets text alignment to `"left"`, `"right"` or `"center"` for text content. Default: `"left"`.
`"textOverflow"`
- `"lineWrap"`: Toggles line wrapping for text content to `true` or `false`. Default `true`.
- `"lineHeight"`: Sets the number of lines displayed for text content before hiding overflow with an elipsis(...). Default: `2`.

### Proposed format for configuration options for Repeating Groups:

`tableHeaders` exists already, which, specifies which fields should be displayed in the table when the group is closed and in which order.
`tableColumns` will not determine any header ordering but will add column formatting options to headers based on the id's of the components in the header.
```
"tableHeaders": ["currentValue", "newValue", "mainUploaderSingle", "mainUploaderMulti"],
"tableColumns": {
  "currentValue": {
    "width": "25%",
    "alignText": "right",
    "textOverflow": {
      "lineWrap": true, 
      "maxHeight": 2
    }
  },
  "newValue": {
    "width": "shrink",
    "alignText": "center",
    "textOverflow": {
      "lineWrap": false
    }
  }
},
```

[...]

PR #731, opened 2022-12-09
Fix wordwrapping in new repeating group table

bjosttveit:
## Description

Quick fix to deal with long words in repeating group.

Before:
image

After:
image

### DIFF
--- a/src/features/form/containers/RepeatingGroupTable.tsx
+++ b/src/features/form/containers/RepeatingGroupTable.tsx
@@ -217,6 +217,9 @@ const useStyles = makeStyles({
   tableButton: {
     width: 'max-content', // Stops column from shrinking too much
   },
+  breakWord: {
+    wordBreak: 'break-word',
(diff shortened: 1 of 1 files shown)


olemartinorg (approved):

Magnusrm:
I think the previous implementation used a feature from MUI table where the line overflow would be hidden. 
Something like:   | Lorem ipsum do... |
Dont know if this way of linebreak is still wanted, but it does at least prevent huge strings to distort the table.

olemartinorg:
Absolutely correct, @Magnusrm.
Possible with some simple CSS as well: https://css-tricks.com/snippets/css/truncate-string-with-ellipsis/
I've mentioned that this should be configurable as well, in #603

bjosttveit:
I'll probably also be doing #603, but I'll wait for your PR @Magnusrm before making more changes to `RepeatingGroupTable` 😅

PR #733, opened 2022-12-12
Fix race-condition in useDelayedSavedState

bjosttveit:
## Description

A race-condition can occur if a value is updated with `setValue` very quickly followed by `saveValue`. This fixes that issue by using a ref for immediateState so that `saveValue` will always save the most up-to-date value.

## Related Issue(s)
- https://altinn.slack.com/archives/C02EVE4RU82/p1670585968531819

## Verification
- Manual testing
- Automated tests
  - [ ] Has been updated
   
  - [x] No changes made
- Support in Altinn Studio
   
- Sprint board

### DIFF
--- a/src/components/hooks/useDelayedSavedState.ts
+++ b/src/components/hooks/useDelayedSavedState.ts
@@ -14,9 +14,18 @@ export function useDelayedSavedState(
   formValue?: string,
   saveAfter?: number | boolean,
 ): DelayedSavedStateRetVal {
-  const [immediateState, setImmediateState] = React.useState(formValue);
+  const [immediateState, _setImmediateState] = React.useState(formValue);
+  const immediateStateRef = React.useRef(formValue);
   const [saveNextChangeImmediately, setSaveNextChangeImmediately] = React.useState(false);
   const [skipNextValidation, setSkipNextValidation] = React.useState(false);
+
+  const setImmediateState = React.useCallback(
+    (value: string | undefined) => {
+      immediateStateRef.current = value;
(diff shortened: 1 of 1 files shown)


Magnusrm (approved):
Well spotted! 🔎

src/components/hooks/useDelayedSavedState.ts
```diff
@@ -14,10 +14,19 @@ export function useDelayedSavedState(
...
-  const [immediateState, setImmediateState] = React.useState(formValue);
+  const [immediateState, _setImmediateState] = React.useState(formValue);
+  const immediateStateRef = React.useRef(formValue);
   const [saveNextChangeImmediately, setSaveNextChangeImmediately] = React.useState(false);
   const [skipNextValidation, setSkipNextValidation] = React.useState(false);
 
+  const setImmediateState = React.useCallback(
+    (value: string | undefined) => {
+      immediateStateRef.current = value;
+      _setImmediateState(value);
```
olemartinorg, line 25: Could we avoid duplicating the state here and just use `immediateStateRef.current` every time? I'm not sure what the purpose of double-storing it is.

olemartinorg (approved):

src/components/hooks/useDelayedSavedState.ts
bjosttveit: Is it not necessary to use State for the hook to trigger an update?

olemartinorg: Oh, right, yep - that's right!

PR #729, opened 2022-12-09
Receipt as layoutpage

andrealouisestanderen:
## Description
- Adds instance information component as layout component
- Add custom receipt as a react component that is rendered instead of Altinn receipt if a custom receipt is available
- Add flag to attachmentsList for including pdf

## Related Issue(s)
- closes #{issue number}

## Verification
- Manual testing
  - [x] I have tested these changes manually
  - [ ] Creator of the original issue (or service owner) has been contacted for manual testing
  - [ ] No testing done/necessary
- Automated tests
  - [ ] Unit test(s) have been added
  - [ ] Cypress E2E test(s) have been added
  - [ ] No automatic tests are needed here
  - [ ] I want someone to help me make some tests
- User documentation @ [altinn-studio-docs](https://github.com/Altinn/altinn-studio-docs)
  - [ ] Has been updated
   
  - [ ] No changes/updates needed
- Changes/additions to component properties
  - [ ] Changes are reflected in both `layout/index.ts` and `layout.schema.v1.json`, and these are all backwards-compatible
  - [ ] No changes made
- Support in Altinn Studio
  - [ ] Issue(s) created for support in Studio
   
  - [ ] This change/feature does not require any changes to Altinn Studio
- Sprint board
  - [ ] The original issue (or this PR itself) has been added to the Team Apps project and to the current sprint board

PR #736, opened 2022-12-12
Print-friendly page for PDF generation

### LINKED ISSUE #710: Automatically generate PDF from layout
### Description

As a user I would like to have a PDF to serve as a receipt for my submitted form data so that I have a copy of what's sent in.

As a app developer I would like to have the PDF layout generated automatically based on my existing page layout(s) so that I don't have to do anything special in order to serve a PDF for the end user, but I would like to be able to override it if I am not happy with the default.

### Additional Information

Epic: https://github.com/Altinn/app-lib-dotnet/issues/63
Backend:

### Tasks

- [x] Create a page that automatically concatenates the existing layouts to serve as the foundation for the PDF the page is triggered by adding pdf=1 to the query parameters of a regular instance url.
- [x] Ensure the excludeFromPdf flag is still respected on both page and component level
- [x] Make sure the generated page removes components like buttons/navigational elements not relevant for print including summary components
- [x] Provide an easy way to specify a specific page to be used instead of the auto generated page by setting mode=pdf on the page object

### Acceptance Criterias

_No response_

bjosttveit:
## Description

I guess this is ready for review as a first version. It seems to behave fine when using chrome (which is what matters). Not sure if it is robust enough to replace the old PDF generator by default yet, so at this point we could really benefit from some other people trying to use it in their apps to see how it goes.

@olemartinorg In order to make sure that all option- and datalists are loaded, there were some small things added to the sagas. It essentially counts the number of options/datalists as they start loading, and when they have finished loading they count up the number that has finished loading and finally sets `loading=false` indicating that all options/datalists are loaded. Let me know if this is a terrible way to achieve this. 

I will create a separate issue regarding the documentation for this. Maybe also an issue in altinn-studio regarding creating a custom PDF-layout there.

- [x] Autogenerated PDF based on all layouts
  - [x] Respects `pages.excludeFromPdf`
  - [x] Respects `components.excludeFromPdf`
  - [x] Respects page order
  - [x] Respects hidden pages
- [x] Ability to provide custom PDF-layout
  - [x] Ability to specify page breaks
- [x] Front page with instance information. (same as #729?)
- [x] Design
  - [x] Make `Panel` work with PDF layout 
    - [x] Update design system to have padding in media print as well?
  - [x] Spacing between paragraph elements
  - [x] Test if natural page breaks work well, espectially for large components like repeating groups.
- [x] Make 100% sure that everything is loaded when `#readyForPrint` is shown.
  - [x] Options
  - [x] Datalists
  - [x] Images
  - [x] Instance information
  -  ...

https://user-images.githubusercontent.com/47412359/210245992-f4c9502a-7e2d-4696-a48d-57a414c77240.mov

## Related Issue(s)
- closes #710
- https://github.com/Altinn/app-frontend-react/pull/729
- https://github.com/Altinn/altinn-design-system/pull/249

## Verification
- Manual testing
- Automated tests
  - [ ] Has been updated
   
  - [ ] No changes made
- Support in Altinn Studio
   
- Sprint board


src/features/pdf/AutomaticPDFLayout.tsx
```diff
@@ -0,0 +1,125 @@
...
+  'List',
+  'Map',
+  'MultipleSelect',
+  'RadioButtons',
+  'TextArea',
+]);
+
+const presentationComponents = new Set(['Header', 'Paragraph', 'Image', 'Panel']);
+
+const renderComponents = new Set([...summaryComponents, ...presentationComponents]);
```
olemartinorg, line 30: Yikes! :laughing: Prime candidates for the `LayoutComponent` class. :partying_face:

src/features/pdf/PDFView.tsx
```diff
@@ -0,0 +1,93 @@
...
+    !instance ||
+    !allOrgs ||
+    !profile
+  ) {
+    return null;
+  }
+
+  const pdfLayout = pdfLayoutName ? layouts[pdfLayoutName] : undefined;
+
+  document.body.style.backgroundColor = 'white';
```
olemartinorg, line 65: Not very react friendly/reusable..? Won't this affect future re-renders when this component is not rendered? This should probably be wrapped in a `React.useEffect` that resets the state upon unmounting. If `PrintButton` renders this component in the background (but sets the global CSS to just print this content and nothing else), it will be very confusing why the page bakground changes to white as soon as you press the print button (and never changes back).

src/features/pdf/AutomaticPDFLayout.tsx
bjosttveit: Haha, I know 😆

src/features/pdf/PDFView.tsx
bjosttveit: oh right, forgot about that one 😅 , for some reason it did not affect future renders when going back and fourth between PDF and regular view. So I totally forgot it like that

schemas/json/layout/layoutSettings.schema.v1.json
```diff
@@ -77,6 +80,11 @@
       "additionalProperties": false,
       "description": "An attribute specifying if the translation dropdown menu should be visible",
       "type": "boolean"
+    },
+    "pdfLayoutName": {
+      "additionalProperties": false,
+      "description": "Name of a custom layout file to use for PDF creation instead of the automatically generated PDF.",
+      "type": "string"
     }
```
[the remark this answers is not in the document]
bjosttveit: How would page-ordering work with different modes? PDF is completely separate from the rest of the pages so it would need its own pageordering to support multiple layout files then?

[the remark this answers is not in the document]
bjosttveit: But to your first point, yes, it should simply be omitted from page-order and the layout will never be accessible through regular navigation

schemas/json/layout/layout.schema.v1.json
```diff
@@ -117,6 +117,9 @@
           ]
         }
       },
+      "pageBreak": {
+        "$ref": "#/definitions/pageBreak"
+      },
```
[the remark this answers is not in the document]
bjosttveit: It is for all components, so that you can e.g. have a new page before a heading which is a GenericComponent.
Specifically it is implemented in SummaryComponent, GenericComponent, and DisplayGroupContainer

src/features/pdf/PDFView.tsx
bjosttveit: So apparently the reason it works is that when not in the PDF-view, it uses the `Presentation` component, which does the same thing 😅 
image

olemartinorg: Gosh, never mind then! :facepalm:

schemas/json/layout/layout.schema.v1.json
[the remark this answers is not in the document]
olemartinorg: Right, because you could co-mingle any component on the print page if you wanted.. I see!

schemas/json/layout/layoutSettings.schema.v1.json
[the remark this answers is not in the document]
olemartinorg: That's a very good point about internal page-ordering when having multiple PDF pages.. There's actually a (undocumented) alternative way of ordering pages, where you can specify (in the layout file itself) what is the next/prev page to go to from that point on. If that was documented, and could be controlled using expressions, it would be much more powerful than the current tracks functionality. (Look for the `INavigationConfig` type if want to unravel that thread.)

olemartinorg (approved):
Fantastic work! :partying_face: :clap: 

Regarding your questioning of the options-fetched-counting, I guess it's a perfectly fine way  to do it. It's more robust than how it was before, at least. It might be hard to maintain and make sure new features keep making sure they load everything before the page is printed, but that is an entirely different issue that shouldn't be tackled here.

PR #746, opened 2022-12-13
Moving files related to layout components

olemartinorg:
## Description
Yet another large change from me.. Sorry! This PR moves around lots of files in order to centralize them in the new `src/layout` folder. This means every folder under there represents a component you can add to your layout file.

A few notes on these:
- `Address` is named `AddressComponent` in the layout files, and is apparently the only one with the `Component` suffix. That's a bit unfortunate, so I just named the folder `Address`.
- `Group` (and by extension, repeating group) along with `Summary` are fundamental constructs, not pure components. However, the Likert component utilized much of the repeating group functionality, so I put that stuff in the `src/layout/Likert` folder.
- I fixed the capitalization of the `Datepicker` component (is was most often named `DatePicker`, but in the layout schema it is named `Datepicker` ([which is what altinn-studio also uses](https://github.com/Altinn/altinn-studio/search?q=DatePicker)). We automatically correct capitalization in component types when we read the layout files, so we still support both variations in the layout files (but on the right one will give you proper schema/intellisense support).
- I moved the layout definitions for each of these components into respective `types.d.ts` files, to give them a place to live (and hint at a structure for when new layout components are developed).

The reason I make this change now is because I wanted a meaningful structure when implementing classes for every component. This will add at least one extra file for every component, so having them all jumbled up in various folders would just make things messier.

## Related Issue(s)
- #377

## Verification
- Manual testing
- Automated tests
  - [ ] Has been updated
   
  - [ ] No changes made
- Support in Altinn Studio
   
- Sprint board


bjosttveit (approved):
Looks goood!

PR #759, opened 2022-12-16
Fix `SummaryGroupComponent` showing non-existing validation errors

### LINKED ISSUE #701: `SummaryGroupComponent` shows non-existing validation errors
### Description of the bug

If there are validation errors on a group, the respective summary component shows a message like this:

image
The problem is that the way it checks for validation errors in the group is wrong, and can potentially say that there are validation errors when there are in fact none. This is because it only checks whether the component-key exists in the validations stored in redux, it does not check if it has an `error`-array with `length > 0`, which is the way the group-component itself checks if it contains errors.

This is the logic currently used in the `SummaryGroupComponent`:

```typescript
if (pageRef && componentIdWithIndex && validations[pageRef] && validations[pageRef][componentIdWithIndex]) {
  groupErrors = true;
}
```

`RepeatingGroupTable` uses the more correct function `componentHasValidations` from `validation.ts` to check for validations in it's children, this does a more comprehensive check:

```typescript
return Object.keys(validations[layoutKey]?.[componentId] || {})?.some((bindingKey: string) => {
    const length = validations[layoutKey][componentId][bindingKey]?.errors?.length;
    return length && length > 0;
  });
```

### Steps To Reproduce

1. Start a new instance in the `frontend-test` app.
2. Navigate to the `group` layout set
3. Do not add prefill elements
4. Add a new row in the repeating group, and fill in some data in the nested group on the second multi-page
5. Save the row, and observe that there are no validation errors
6. Go to summary and see an error message for the nested group

In the redux store, there are several empty validation objects:

image

Since the group summary component only checks if there are any keys present, it will say there are validation errors even though no errors actually exists inside the objects:

image
image

### Additional Information

_No response_

bjosttveit:
## Description

Switched to using `nodesInLayout` to get all children of the group component to check against `formValidations`  properly. I'm not sure if tests are relevant here, as i do not know exactly what caused empty validation objects being left in `formValidations` (causing the old code to show errors), only that it happened quite a lot.

## Related Issue(s)
- closes #701

## Verification
- Manual testing
- Automated tests
  - [ ] Has been updated
   
  - [x] No changes made
- Support in Altinn Studio
   
- Sprint board

### DIFF
--- a/src/components/summary/SummaryGroupComponent.tsx
+++ b/src/components/summary/SummaryGroupComponent.tsx
@@ -14,6 +14,7 @@ import { renderLayoutComponent } from 'src/features/form/containers/Form';
 import appTheme from 'src/theme/altinnAppTheme';
 import { getDisplayFormDataForComponent, getFormDataForComponentInRepeatingGroup } from 'src/utils/formComponentUtils';
 import { getRepeatingGroupStartStopIndex, setMappingForRepeatingGroupComponent } from 'src/utils/formLayout';
+import { nodesInLayout } from 'src/utils/layout/hierarchy';
 import { getLanguageFromKey } from 'src/utils/sharedUtils';
 import { getTextFromAppOrDefault } from 'src/utils/textResource';
 import type { ComponentFromSummary } from 'src/features/form/containers/DisplayGroupContainer';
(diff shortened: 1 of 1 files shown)


Magnusrm (approved):
Looks solid 👍

olemartinorg (approved):

PR #763, opened 2022-12-16
Fixing multiple minor bugs in repeating groups + expressions

### LINKED ISSUE #620: Expressions: Implement Cypress test(s)
We need cypress tests for the new expressions, along with regression tests for previous known bugs (in order to cover advanced functionality). Inspiration can be taken from the existing layout-expressions test app, but this should be applied to the frontend-test app.

### Relevant bugs that should get regression tested:
- #599
- #617

olemartinorg:
## Description
1. The `saveButton` and `saveAndNextButton` properties can have expressions on them, and should be evaluated per-row.
2. When overriding text resource bindings using expressions for a group, they should also be evaluated per-row if possible.
3. Updated the group cypress test (especially the prefill one) to properly handle all prefilled rows, not just the special 'marked' one.
4. The DEFAULT_FOR_ALL_VALUES_IN_OBJ functionality in expressions did not work properly when evaluating deeper inside an object
5. When evaluating an expression, it's perfectly fine to do a "hidden": ["dataModel", "something"] if your data model contains a boolean, but that boolean was not converted/cast at the end, because the `dataModel` function returns the value cast to a string, and the boolean is a string in the data model anyway. However, we can fix this by casting to the same type as the default type (which is always the same as the expected return type, thanks to the deep resolving typescript magic).

Also: Converting the weird 'rows' with fixed lengths to be functions instead.

**NOTE:** There are accompanying changes to `frontend-test` for this as well, finally fixing the brokenness when having selected prefills for repeating groups. All other PR tests will fail until this is merged, and their branches merge/rebase from main.

## Related Issue(s)

- closes #620

## Verification

- Manual testing
- Automated tests
  - [ ] Has been updated
  
  - [x] No changes made
- Support in Altinn Studio
  
- Sprint board

### DIFF
--- a/src/features/expressions/index.ts
+++ b/src/features/expressions/index.ts
@@ -59,7 +59,7 @@ export function evalExprInObj<T>(args: EvalExprInObjArgs<T>): ExprResolved<T> {
 
 function getDefaultValueFor(path: string[], defaults: any) {
   const pathString = path.join('.');
-  const pathStringAnyDefault = [...path.slice(0, path.length - 2), DEFAULT_FOR_ALL_VALUES_IN_OBJ].join('.');
+  const pathStringAnyDefault = [...path.slice(0, path.length - 1), DEFAULT_FOR_ALL_VALUES_IN_OBJ].join('.');
   const defaultValueSpecific = dot.pick(pathString, defaults);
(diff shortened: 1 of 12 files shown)


bjosttveit (approved):
## LGTM 🚀

PR #767, opened 2022-12-21
WCAG updates for List Component

caroliss:
## Related Issue(s)

- closes #{issue number}

## Verification

- Manual testing
- Automated tests
  - [ ] Has been updated
  
  - [ ] No changes made
- Support in Altinn Studio
  
- Sprint board

### DIFF
--- a/src/layout/List/ListComponent.tsx
+++ b/src/layout/List/ListComponent.tsx
@@ -143,7 +143,7 @@ export const ListComponent = ({
     >
       <TableHeader>
         <TableRow>
-          <TableCell radiobutton={true}></TableCell>
+          <td />
           {renderHeaders(tableHeaders)}
         </TableRow>
       </TableHeader>
--- a/src/layout/List/ListComponent.tsx
+++ b/src/layout/List/ListComponent.tsx
@@ -11,10 +11,13 @@ import {
   TableHeader,
   TableRow,
 } from '@altinn/altinn-design-system';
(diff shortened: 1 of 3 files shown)


src/layout/List/ListComponent.tsx
```diff
@@ -127,67 +131,85 @@ export const ListComponent = ({
...
-      {pagination && (
-        <TableFooter>
+    <FormControl component='fieldset'>
+      <FormLabel
+        component='legend'
+        classes={{ root: cn(classes.legend) }}
+        id={`${id}-label`}
+      >
+        <RenderLegend />
+      </FormLabel>
```
olemartinorg, line 155: Any specific reason you chose the mui one, and not [the one from our design system](https://designsystem.altinn.studio/?path=/story/components-fieldset--normal)? :thinking: If the one in the design system is not working right, we might need to look into that.

olemartinorg (approved):

src/layout/List/ListComponent.tsx
caroliss: Just because that was what was used for RadioGroup and CheckboxContainer. Tested with FieldSet from the design system now, but it doesnt seem to support screen reader? The screen reader works better with FormControl from mui, reading the label and "table", but with the FieldSet it doesnt read the label (but all the content of the table at the same time)

olemartinorg: Oh, huh. I'll create an issue for that in the design system then. Thanks for the research!

PR #768, opened 2022-12-21
Fix *FIXED* bug

### LINKED ISSUE #764: When adding more than one *FIXED* ModelError to a field that has an Error, the Error isn't removed even if it's one of the *FIXED* ones
### Description of the bug

When adding more than one `*FIXED*` ModelError to a field with an Error, the Error isn't removed (if the field we add the `*FIXED*` ModelError to isn't the one that triggered Validation). It works as expected if you only add one `*FIXED*` ModelError to that field. 

The application we noticed this on is rrh-innrapportering (https://altinn.studio/repos/brg/rrh-innrapportering/issues). We have deployed a temporary fix that avoids the problem (but that isn't ideal for permanent use), so you will need to check out this commit or an earlier one (https://altinn.studio/repos/brg/rrh-innrapportering/commit/578a2f893d000ad1fd934b4727af4d21e79c8808). 

The code is in `/App/logic/Validation/ValidationHandler.cs`, starting on line 190 in the commit linked above. (Direct link: https://altinn.studio/repos/brg/rrh-innrapportering/src/commit/578a2f893d000ad1fd934b4727af4d21e79c8808/App/logic/Validation/ValidationHandler.cs). 

### Steps To Reproduce

1. For any company, choose "Har reelle rettighetshavere"
2. Select "Ja" on "Har personen enten fødselsnummer eller d-nummer? *"
3. Fill in 11 digits on "Fødselsnummer".
4. Fill in the last name of a person you can test with into the "Etternavn" field and that you have the fødselsnummer of. This will make the Validation fail when you click outside the textfield.
5. Fill in the correct fødselsnummer. 
6. The correct information pops up, but the error under the Etternavn-field is still there. 

### Additional Information

image
image

bjosttveit:
## Description

The problem seems to be that when the validations are mapped to the redux store format using `mapDataElementValidationToRedux`, it will sometimes call `mergeComponentValidations` as part of this process. This strips away the *FIXED* validations by trying to remove the errors that are supposed to be fixed. However, in this step of the process we are not yet looking at existing validation errors, so it will simply strip away the *FIXED* validations before we actually merge with the existing validations. I therefore added a `fixValidations` option (`default=true`) to the merge-functions so that it is possible to merge without applying fixed validations and simply keeping them around. This option is then set to `false` in `mapDataElementValidationToRedux`.

This seems to fix the issue in `brg/rrh-innrapportering`, but because the actual app uses some secrets to look up info on personnummer, I only tested by hacking the validation code.

## Related Issue(s)

- closes #764 

## Verification

- Manual testing
- Automated tests
  - [x] Unit test(s) have been added
  - [ ] Has been updated
  
  - [x] No changes made
- Support in Altinn Studio
  
- Sprint board

### DIFF
--- a/src/utils/validation/validation.test.ts
+++ b/src/utils/validation/validation.test.ts
@@ -1240,6 +1240,14 @@ describe('utils > validation', () => {
           code: '',
         },
         {
+          field: 'dataModelField_1',
+          severity: Severity.Fixed,
+          scope: null,
+          targetId: '',
+          description: 'Another error message',
+          code: '',
+        },
+        {
           field: 'dataModelField_2',
           severity: Severity.Success,
           scope: null,
@@ -1263,6 +1271,7 @@ describe('utils > validation', () => {
             simpleBinding: {
               errors: [getTextResourceByKey('Error message', [])],
               info: [getTextResourceByKey('Info message', [])],
+              fixed: [getTextResourceByKey('Another error message', [])],
             },
           },
           componentId_2: {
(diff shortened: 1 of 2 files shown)


olemartinorg (approved):
Looks good! :+1:  This stuff is becoming really complex, would you mind adding some tests for what `*FIXED*` is supposed to do, and what it's supposed to fix in this broader context?

PR #725, opened 2022-12-08
Improve the display of headers when referenced in summary

ivarne:
Headers look ugly when referenced in a `"type": "Summary"` component and the `Edit` button shows up by default.
image

This PR makes a custom view for a header component.

image

## Verification
- Manual testing
  - [x] I have tested these changes manually
  - [ ] No testing done/necessary
- Automated tests
  - [ ] Unit test(s) have been added
  - [ ] Cypress E2E test(s) have been added
  - [ ] No automatic tests are needed here
  - [x] I want someone to help me make some tests (if required)
- User documentation @ [altinn-studio-docs](https://github.com/Altinn/altinn-studio-docs)
  - [x] No changes/updates needed
- Sprint board
  - [ ] The original issue (or this PR itself) has been added to the Team Apps project and to the current sprint board

PR #793, opened 2022-12-28
When a summary component sets display.hideBottomBorder we don't need space for a border

ivarne:
After giving `Header` components a custom look in https://github.com/Altinn/app-frontend-react/pull/725 I got new complaints about the blue line under the header.

image

There is an existing config `display.hideBottomBorder` (introduced in https://github.com/Altinn/app-frontend-react/pull/489 without docs) that can be set on `Summary` components, to remove the line. Personally I think removing the line should be default for headers, but it's in a different component, so a `if(component.type == "Summary")` would be needed in another file, and I think we should find better solutions (eg. a `const hideBottomBorder = component.hideBottomBorder ?? component.componentTypeDefaults.summary.display.hideBottomBorder`).

This PR removes the extra spacing that is required for the blue dashed line when the blue dashed line is removed. I'm not sure if this extra spacing might be required in other instances, but it looks strange with header components.

image

## Related Issue(s)
- https://github.com/Altinn/app-frontend-react/pull/725
- https://github.com/Altinn/app-frontend-react/pull/489

## Verification

- Manual testing
- Automated tests
  - [ ] Has been updated
  
  - [x] No changes made
- Support in Altinn Studio
  
- Sprint board

### DIFF
--- a/src/components/summary/SummaryComponent.tsx
+++ b/src/components/summary/SummaryComponent.tsx
@@ -27,11 +27,9 @@ export interface ISummaryComponent extends Omit<ILayoutCompSummary, 'type'> {
 }
 
 const useStyles = makeStyles({
-  row: {
+  border: {
     marginBottom: 10,
     paddingBottom: 10,
-  },
-  border: {
     borderBottom: '1px dashed #008FD6',
   },
   link: {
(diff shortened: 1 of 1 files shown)


olemartinorg (approved):
Great, thanks! I also agree these things are something we should make code to 'just figure out' in the future, but I guess some minor CSS changes are fine for now.

PR #798, opened 2022-12-29
Modularized components

### LINKED ISSUE #377: Architecture: Modularize components using classes and inheritance

olemartinorg:
## Description
Started modularizing layout components. In other words, a new `LayoutComponent` class has been added (which every layout component now has to implement). It is the common 'tool box' used by `GenericComponent.tsx` and serves as a place where every peculiarity of each layout component implementation should be implemented.

This is the start, and builds a framework we can extend further. As discussed in the comments under, `getDisplayFormData` is probably the next thing to start working on moving into this class.

As this change is only about moving code around, I'm relying on our automatic tests to catch any bugs introduced, and so I haven't really tested this change myself. It turned out that not all components expected all the same properties as they get from `GenericComponent.tsx`, so I had to clean some of them up for this to work. For example, `CheckboxContainerComponent` could render validation messages if given `validationMessages`, but that doesn't ever seem to happen (so I removed that part).

## Related Issue(s)
- #377

## Verification

- Manual testing
- Automated tests
  - [ ] Has been updated
  
  - [x] No changes made
- Support in Altinn Studio
  
- Sprint board


olemartinorg:
What do you think, @bjosttveit? I know you had thoughts about how to solve this. Did you imagine it differently, or were we on the same track?

bjosttveit:
> What do you think, @bjosttveit? I know you had thoughts about how to solve this. Did you imagine it differently, or were we on the same track?

I guess my thoughts were mostly idealistic and more relevant to a major release by having component classes do much more than having a render function (like validation logic, having some sort of `renderSummary()`, and also replacing `getDisplayFormData` by using a function defined in the class for getting a string-representation of its data, etc.). I also thought about having different types of groups be different classes, and kind of inplicitly included https://github.com/Altinn/app-frontend-react/issues/345 in my thoughts. That would be a very large refactor though 😅. But this is definitely on the same track and a good step in the right direction!

olemartinorg:
Ah, yup, sounds like we think much in the same way, then.. :sweat_smile: I think smaller steps at each point in time will be easier to implement (and test), so I also think this is just the beginning. I thought about starting the refactor to include `getDisplayFormData` last week, but I think that's next (and then moving on to Summary).

I'm also having (admittedly half-baked) thoughts about running `nodesInLayout` first, then passing each `LayoutNode` to `GenericComponent` instead - and in that process making (repeating) groups something more similar to a normal component (just rendering a table, but also making sure it can render child components). There's a lot of code in `src/utils/formLayout.ts` that does much of the same as the node hierarchy, so it would be nice to get rid of the duplication. That may be a practical route to complete #345, but that issue is spinning out in a direction where replacing Redux/Sagas seems to be on the table (at least in my head, and with my current understanding of the issue).

I'll open up this one for review, so that we can tackle each puzzle piece at a time (so not closing #377 quite yet).

bjosttveit:
Sounds good! So I guess we are on the same page, except you actually have a plan to get there, I had no idea where to begin 😅

bjosttveit (approved):
# 🎖️

PR #800, opened 2022-12-30
Bugfix/124 validation message still active for hidden field in repeating group

### LINKED ISSUE #124: Validation messages not removed inside repeated groups
## Description
Error message after validation on field inside repeating group is not removed.
## To Reproduce
(Specific for [our app](https://altinn.studio/repos/dsb/melding-om-sikkerhetsraadgiver/src/commit/7b65b6909d2e6ec390db5bd5df661fc4c7562a2a))
1. Add new or edit an entry in a repeating group.
2. Trigger a validation error on a field that is validated depending on another field.
3. Address the validation error by changing the other field the validation is dependent upon. 

## Expected behavior
The validation error message is removed.

## Screenshots
Kapture 2022-05-24 at 13 40 07

## Additional info
* The error-message is removed if i focus the field to remove the error as demonstrated in the GIF.
* The dynamics causing the relevant field to hide and show seems to be irrelevant for the problem.
* In this example its server-side validation that checks if the SSN is required or not. I have also tried setting the field to required in the layout and relying on the dynamic to remove and add the field, but this produces the same result.
* [Relevant slack discussion](https://altinn.slack.com/archives/C02ESDSC929/p1653052884981219)

Magnusrm:
## Description

Handle race condition where validation message is added to a hidden field after it is dynamically set to hidden in a repeating group.

The field is now checked for being hidden before the single-field-validation is being run, and right before the validation-messages are being updated in redux to make sure the field has not been set to hidden in the time since we started the single-field validation process. 

Specific states are also now retrieved before usage instead of retrieving the entire state at the start of the process.

## Related Issue(s)

- closes #124 

- PR to frontend-test for cypress test: https://dev.altinn.studio/repos/ttd/frontend-test/pulls/13

## Verification

- Manual testing
- Automated tests
  - [ ] Has been updated
  
  - [x] No changes made
- Support in Altinn Studio
  
- Sprint board


src/features/form/validation/singleField/singleFieldValidationSagas.ts
```diff
@@ -35,14 +39,19 @@ export function* runSingleFieldValidationSaga({
...
 
-      const validations = mergeValidationObjects(state.formValidations.validations, mappedValidations);
+      const mappedValidations = mapDataElementValidationToRedux(serverValidation, layouts || {}, textResources);
+      const validationsFromState = yield select(selectValidationsState);
+      const validations = mergeValidationObjects(validationsFromState, mappedValidations);
+
+      // Reject validation if field is hidden
+      const hiddenFields = yield select(selectHiddenFields);
+      if (componentId in hiddenFields) {
+        throw Error('Single field validation rejected as field is hidden for user');
```
olemartinorg, line 53: 1. Not sure if the `in` keyword works like that.. `in` works on keys, not values - and an array-type in JavaScript has numeric keys. See more here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/in#basic_usage
2. Also, throwing an error probably makes the whole saga reject with the thrown error, storing the error in the Redux state and causing an `unknown error` to show on-screen. A bit aggressive for an error like this, which I think should be ignored silently (no need to update validations for components that have been hidden).
3. Additionally, if I were to read this code now (without knowing about this specific bug), I'd probably think "hey, there's no need to run this GET request first, then check if the field is hidden after getting a response - we can optimize away the whole request by just checking if the field is hidden before the request". I think we should check if the field is hidden _before_ sending the request as well, and update the comment here to describe that we meant to reject validation if the field _has been hidden in the time after we sent the request_ as well.

All in all, I think this warrants a cypress test. We should document this behavior via an e2e test, so as to make sure it doesn't regress in the future.

olemartinorg (approved):
Looks good! The norm is to put out a `runSingleFieldValidationRejected` action when returning (without an error object in it), so you can tell when the action completed, but I guess it's not that important (nobody really listens to it anyway).

Magnusrm:
> Looks good! The norm is to put out a `runSingleFieldValidationRejected` action when returning (without an error object in it), so you can tell when the action completed, but I guess it's not that important (nobody really listens to it anyway).

Oh, ok! I'll just quickly add a `runSingleFieldValidationRejected` action then for good measure and for future reference 👍

test/e2e/fixtures/texts.json
```diff
@@ -15,6 +15,7 @@
   "prev": "Forrige",
   "next": "Neste",
   "requiredField": "Feltet er påkrevd",
+  "requiredFieldFromBackend": "Feltet er påkrevd (fra backend)",
```
olemartinorg, line 18: Oooh, sorry about that one! :see_no_evil: I might have changed that label without thinking all that much, but thanks for making the change here. I guess a new deployment of `frontend-test` is needed as well.

I've heard people ask why the older error message stayed just for this field here, that's why I added the note about this coming from backend.

Magnusrm: No worries! 😄 I already deployed a new version of frontend-test which is how i noticed this 😅 I think frontend-test should be good if nothing else pops up.

PR #806, opened 2023-01-03
Responsive listcomponent

johlie:
## Description
Added responsive design to listcomponent. 

## Related Issue(s)

- closes #{issue number}

## Verification

- Manual testing
- Automated tests
  - [ ] Has been updated
  
  - [ ] No changes made
- Support in Altinn Studio
  
- Sprint board


src/layout/List/types.d.ts
```diff
@@ -12,6 +12,7 @@ export interface ILayoutCompList extends ILayoutCompBase<'List'> {
   dataListId: string;
   secure?: boolean;
   bindingToShowInSummary?: string;
+  chosenPrioritizedColumsMobile?: string[];
```
olemartinorg, line 15: Remember to add extra properties to the layout json schema as well. :smiley:

olemartinorg: Also, I find the words "chosen" and "prioritized" to be a bit vague when referring to a specific feature. Are these the columns you _choose_ to display on mobile, **or** are they _prioritized_ in a way? From glancing at the code, it seems this property needs to be set, and have to include all column keys you want to display on mobile...

Could we instead make it so that all columns are shown on mobile if this property is not set? And then possibly call it something like `tableHeadersMobile`? That way it's obvious it's the same functionality as the existing `tableHeaders` functionality, just for mobile. Alternatively `mobileColumns`, as that indicates it's the same as `sortableColumns`, also just for mobile.

olemartinorg (changes requested):
Is there a way to re-use some of the code here? Now we suddenly have two very different paths (`isMobile` and `isLaptop`) which seem to include lots of the same code. Both use the Table component, at least.. Why not extract what needs to be different from mobile/laptop and just implement different variations of that instead?

Also, I expect a function called `isMobile` to return a boolean (from the `is` prefix), but these return React components. I don't think this top-level-switch between mobile/desktop sits quite right with me, but if we end up keeping it, I think these should be called something like `RenderMobile` and `RenderDesktop` instead.

PR #811, opened 2023-01-04
Null check empty optionLabel from repeating group

bjosttveit:
## Description

When using a repeating group as a source for options, there was a bug where adding a new element to the source repeating group would cause a crash because of a missing null check. This is because it would iterate over the length of the repeating group, and when a new element is added, the length (repeatingGroupIndex) is incremented before any new data is added. This caused a `cannot read properties of undefined` exception. Fixed by a simple null check on the option to be added.

Confirmed that the fix works in `ra0244-01`.

## Related Issue(s)

- https://altinn.slack.com/archives/C02EVE4RU82/p1672816833823229

## Verification

- Manual testing
- Automated tests
  - [ ] Has been updated
  
  - [x] No changes made
- Support in Altinn Studio
  
- Sprint board

### DIFF
--- a/src/utils/options.ts
+++ b/src/utils/options.ts
@@ -122,11 +122,13 @@ export function setupSourceOptions({
 
   const options: IOption[] = [];
   for (let i = 0; i <= repGroup.index; i++) {
-    const option: IOption = {
-      label: replacedOptionLabels[i + 1].value,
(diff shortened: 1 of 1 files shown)


Magnusrm (approved):
Nice! 😄

PR #827, opened 2023-01-11
Easier preview for PDF development

bjosttveit:
## Description

Made some changes to `ProcessWrapper` to make it easier to preview PDF while developing. Now there are three possible states of PDF rendering depending on the query parameter `?pdf`.

- `?pdf` is not defined: Only form is rendered as normal.
- `?pdf=1`: used for generating PDF. Only `PDFView` is rendered to keep PDF-generation fast.
- `?pdf=preview`: both form and `PDFView` are rendered at the same time, but form is hidden and `PDFView` is visible when in print-mode, and opposite if not in print mode (using css). This can potentially be very slow, and is only meant for development use.

Does this seem like a good solution? It makes it much easier to modify form data and quickly check how this changes the PDF without constantly changing the query-parameter back and fourth as I have been doing before.

### DIFF
--- a/src/features/pdf/PDFView.tsx
+++ b/src/features/pdf/PDFView.tsx
@@ -61,8 +61,6 @@ const PDFView = ({ appName, appOwner }: PDFViewProps) => {
   }
 
   const pdfLayout = pdfLayoutName ? layouts[pdfLayoutName] : undefined;
-
-  document.body.style.backgroundColor = 'white';
   return (
     <div className={css['pdf-wrapper']}>
       <h1 className={cn({ [css['title-margin']]: !appOwner })}>{appName}</h1>
--- a/src/index.css
+++ b/src/index.css
@@ -13,6 +13,12 @@ body {
   background-color: #efefef;
 }
 
+@media print {
+  body {
+    background-color: white !important;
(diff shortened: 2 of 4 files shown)


olemartinorg (approved):
Looks good! If this is meant to be used when clicking `PrintButton` as well, I think we should consider having the pdf-preview flag in Redux store instead of in the URL. Or, it will look a bit funky if we have to change the URL and keep that state in sync when pressing a button.

bjosttveit:
For now at least, it is not meant to be used with `PrintButton`, as I am not yet sure exactly what this button is supposed to do. But as you say, the button does not need to modify the hash, it could be done through redux with some more modifications. This is only meant to offer a preview without having to add anything visible to the form (like a print button), and I imagine this would still be used even if the print button was changed to achieve the same thing.

olemartinorg:
> and I imagine this would still be used even if the print button was changed to achieve the same thing

:clap: Yep, I get it! So in the future:

```typescript
const previewPDF = searchParams.get('pdf') === 'preview' || state.something.previewPDF;
```

bjosttveit:
Exactly 🤌

PR #835, opened 2023-01-12
List component: Using responsive/mobile table

johlie:
## Description
Changes after new table in design system. The tests are not working. 

## Related Issue(s)

- closes #{issue number}

## Verification

- Manual testing
- Automated tests
  - [ ] Has been updated
  
  - [ ] No changes made
- Support in Altinn Studio
  
- Sprint board


src/layout/List/ListComponent.tsx
```diff
@@ -55,50 +46,34 @@ export const ListComponent = ({
...
     }
-    return cell;
-  };
+    return {};
+  }, [formData, calculatedDataList]);
 
   const dispatch = useAppDispatch();
 
-  const handleSortChange = ({ sortedColumn, previousSortDirection }: SortProps) => {
+  const handleSortChange = (props: SortProps & { column: keyof Record<string, string> }) => {
```
olemartinorg, line 71:
```suggestion
  const handleSortChange = (props: SortProps & { column: string }) => {
```
A "key in a record where the keys are strings", can be reduced to just use the "string" type. :smile:

src/layout/List/ListComponent.tsx
```diff
@@ -55,50 +46,34 @@ export const ListComponent = ({
...
-      } else {
-        cell.push(<TableCell key={header}>{getTextResourceAsString(header)}</TableCell>);
+  const selectedRow: Record<string, string> = React.useMemo(() => {
+    let matchRow: boolean[] = [];
+    for (const row of calculatedDataList) {
+      for (const key in formData) {
+        matchRow.push(formData[key] === row[key]);
+      }
+      if (!matchRow.includes(false)) {
+        return row;
```
olemartinorg, line 62: I might be misunderstanding, but if `formData` is not set here (or is an empty object), the first row will be selected, right?

src/layout/List/ListComponent.tsx
```diff
@@ -120,96 +95,51 @@ export const ListComponent = ({
...
+      <Pagination
+        numberOfRows={totalItemsCount}
+        rowsPerPageOptions={pagination?.alternatives ? pagination?.alternatives : []}
+        rowsPerPage={rowsPerPage}
+        onRowsPerPageChange={handleChangeRowsPerPage}
+        currentPage={currentPage}
+        setCurrentPage={handleChangeCurrentPage}
+        descriptionTexts={getLanguageFromKey('list_component', language)}
+      />
+    ),
```
olemartinorg, line 125: Shouldn't pagination be optional? Seems like it is always rendered now.

src/layout/List/types.d.ts
```diff
@@ -7,13 +7,14 @@ export interface IPagination {
 }
 
 export interface ILayoutCompList extends ILayoutCompBase<'List'> {
-  tableHeaders?: string[];
+  tableHeaders: { [Col in keyof T]: string };
```
olemartinorg, line 10: I don't think `T` is defined here. It should probably be:

```suggestion
  tableHeaders: Record<string, string>;
```

olemartinorg: Also, make sure to update the JsonSchema when changing these types! The two definitions must be kept in sync.

olemartinorg (commented):
Would you mind running `yarn upgrade-interactive`, selecting the new design system `v0.26.0` and committing those changes to `package.json` and `yarn.lock` as well? That way the tests should be running smoothly. :pray:

src/layout/List/ListComponent.tsx
johlie: No, it is only returned a row if one is clicked. If formData is not set or is an empty object the function returnes {}. We made a function to figure out which row that was selected by comparing the formdata and the rows. We made a list with only booleans and if that list doesn´t include any false when the loop is finished it returns the selectedRow. The loop checks all the fields in formdata against the row and if one of the fields does not match false is added to the string. It is not the best solution, but we could not figure out any better way. It is just to make suggestions if you have any :)

olemartinorg: Yup, I just figured `matchRow` won't contain any `false`-values when there is nothing in `formData` to iterate. In that case, `row` is returned. I guess you tested this with a fresh instance?

My suggestion would just be to return early with `{}` if `formData` is not set/is empty, making that behaviour explicit. So, when reading the function, it would say:
1. If `formData` is not set/empty, return an empty object
2. If `formData` matches one of the rows, return that row
3. Otherwise, return an empty object

As it stands, 1 is missing from the code.

schemas/json/layout/layout.schema.v1.json
```diff
@@ -873,12 +873,14 @@
...
-          },
+          "type": "object",
           "title": "Table Headers",
-          "description": "An array of strings that is going to be headers of the table. Can be added to the resource files to change between languages"
+          "examples": [
+            {
+              "AFieldInRowDataModel": "header1",
+              "OtherFieldInRowDataModell": "header2"
+            }],
+          "description": "An object where the fields in the datamodel is mapped to headers. Must correspond to datamodel representing a row. Can be added to the resource files to change between languages"
```
olemartinorg, line 883: "Datamodel" means something very different, so we should use a different word. This about keys in the DataList, right? If I saw the example for the first time, it would be difficult for me to guess if the key or value here is the key in DataList, and which of them is the label the user will read. How about this example instead?

```json
{
  "productId": "product.id",
  "description": "Beskrivelse av produkt"
}
```

johlie:
la inn en PR i frontend-test siden jeg måtte oppdatere den for at cypress-testene skulle funke: https://dev.altinn.studio/repos/ttd/frontend-test/pulls/15
Får enda en feil på noen av cypress testene, men vet ikke hva det er på grunn av:
image

johlie:
Nei, var ikke på group, var på flere andre, men virker som ikke de feilet nå.

johlie:
eksempel på feil som er i group nå: 
image

olemartinorg (approved):
Thank you! Merging this and fixing the remaining cypress test in-house.

Issue #232, opened 2022-06-17
WCAG: Contrast on our focus states


Febakke:
### Description of the bug

When using a keyboard to navigate a page it is important to have a good focus indicator so the user knows where the focus is at any point. We have got feedback that the exit button and our primary button should have better focus states. 

[Read more on WCAG 1.4.11 ](https://www.w3.org/WAI/WCAG21/Understanding/non-text-contrast.html)

Primary:
![Skjermbilde 2022-06-17 kl  12 49 47](https://user-images.githubusercontent.com/33222679/174285759-5e425b9f-0dc9-4e83-8001-b7a7b49b3d21.png)

Exit button:
![Skjermbilde 2022-06-17 kl  12 49 20](https://user-images.githubusercontent.com/33222679/174285776-db86a6a4-e61e-4dfc-aa1c-92979f3e76fa.png)

## tasks
- [ ] Are there more focus states we should have look on while we are at it?
- [ ] Design new focus states
- [ ] implement

### Steps To Reproduce

Go to any app and navigate using a keyboard.

bjosttveit:
Navigation buttons can easily be fixed by using design system buttons instead of the `AltinnButton` which is currently used. The design system buttons have much clearer focus states.

![Image](https://user-images.githubusercontent.com/47412359/214056733-a547f954-b251-4185-b139-34f0d1dc3957.png)

The exit button could maybe use the quiet variant of the design system button.

In that case, there would be no need to design new focus states.

Thoughts @Febakke ?

bjosttveit:
I can do the low hanging fruit first, like using the design system buttons where applicable (navigation, submit, possibly exit-button?)

Could you make a clear list of which components are problematic @Febakke ?

PR #855, opened 2023-01-24
Remove old design system dependency and import css directly

bjosttveit:
## Description

When trying to fix some styling I could not figure out where the css files were coming from. Turns out the (really) old [design system](https://github.com/Altinn/DesignSystem) was still imported as a dev-dependency and injected into `index.css` (only a single stylesheet). To get rid of this dependency I copied the file into this repo since this package has now been deprecated.

## Related Issue(s)

- https://github.com/Altinn/app-frontend-react/issues/232

## Verification

- Manual testing
- Automated tests
  - [ ] Has been updated
  
  - [x] No changes made
- Support in Altinn Studio
  
- Sprint board

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,6 @@
     "@types/uuid": "8.3.4",
     "@typescript-eslint/eslint-plugin": "5.48.2",
     "@typescript-eslint/parser": "5.48.2",
-    "altinn-designsystem": "3.1.0",
     "axe-core": "4.6.2",
     "babel-jest": "29.3.1",
     "babel-loader": "9.1.2",
--- a/src/index.css
+++ b/src/index.css
@@ -1,4 +1,4 @@
-@import '~altinn-designsystem/dist/css/style.dist.tjenester3.min';
+@import 'src/styles/tjenester3.css';
 
 /* Custom CSS for all app */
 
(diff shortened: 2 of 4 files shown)


olemartinorg:
Uhm.. I agree we should remove this older dependency, and I've been in the room when this has been part of larger discussions, but I don't quite see what this will accomplish.

1. Did something break in a way in which it was impossible to continue depending on this package? (i.e., has it been unpublished?)
2. Are we doing this to override/make changes to the file? In that case, there is only one commit, to it's hard for me to review this and track changes.
3. By adopting this file in its entirety, we also take on the responsibility of maintaining it. +23K lines of CSS is not the direction we're aiming for, right?

Again, I agree that we should remove this dependency, but moving is not removing. I'd much rather prefer we start isolating the parts where this is still needed, and possibly also start working on removing this dependency from those places. If we could get to a place where only certain parts of the big CSS file is imported (or having a bigger plan to gut it after moving it here), I'd be more convinced to take on changes like this.

bjosttveit:
> Uhm.. I agree we should remove this older dependency, and I've been in the room when this has been part of larger discussions, but I don't quite see what this will accomplish.
> 
>     1. Did something break in a way in which it was impossible to continue depending on this package? (i.e., has it been unpublished?)
> 
>     2. Are we doing this to override/make changes to the file? In that case, there is only one commit, to it's hard for me to review this and track changes.
> 
>     3. By adopting this file in its entirety, we also take on the responsibility of maintaining it. +23K lines of CSS is not the direction we're aiming for, right?
> 
> 
> Again, I agree that we should remove this dependency, but moving is not removing. I'd much rather prefer we start isolating the parts where this is still needed, and possibly also start working on removing this dependency from those places. If we could get to a place where only certain parts of the big CSS file is imported (or having a bigger plan to gut it after moving it here), I'd be more convinced to take on changes like this.

For example, the back and close buttons gets some styling from this related to focus that I need to remove to solve #232 . In this PR I have only copied the file exactly as is. But I agree that the components that use classes from this file should be updated to no longer depend on it, and then eventually remove this file completely.

olemartinorg (approved):
In that case, you'll get a go from me for an 'embrace and extinguish' strategy!

PR #857, opened 2023-01-25
WCAG: Likert component fixes

### LINKED ISSUE #316: WCAG issues with the likert component

mijohansen:
## Description

This is a pr to help the apps-team. Extensive PR-comments could be added by receiving team if needed. 

## Related Issue(s)

- closes #316

## Verification

- Manual testing
- Automated tests
  - [ ] Has been updated
  
  - [ ] No changes made
- Support in Altinn Studio
  
- Sprint board

### DIFF
--- a/src/features/form/containers/GroupContainer.tsx
+++ b/src/features/form/containers/GroupContainer.tsx
@@ -85,6 +85,7 @@ export function GroupContainer({ id, container, components }: IGroupProps): JSX.
   const options = useAppSelector((state) => state.optionState.options);
   const textResources = useAppSelector((state) => state.textResources.resources);
   const repeatingGroupIndex = repeatingGroups && repeatingGroups[id] ? repeatingGroups[id].index : -1;
+
   const repeatingGroupDeepCopyComponents = useMemo(
     () => createRepeatingGroupComponents(container, renderComponents, repeatingGroupIndex, textResources, hiddenFields),
     [container, renderComponents, repeatingGroupIndex, textResources, hiddenFields],
--- a/src/features/form/containers/GroupContainer.tsx
+++ b/src/features/form/containers/GroupContainer.tsx
@@ -5,8 +5,10 @@ import { Grid } from '@material-ui/core';
 import { Add as AddIcon } from '@navikt/ds-icons';
 
 import { useAppDispatch, useAppSelector } from 'src/common/hooks';
+import { ConditionalWrapper } from 'src/components/ConditionalWrapper';
 import { ExprDefaultsForGroup } from 'src/features/expressions';
 import { useExpressions } from 'src/features/expressions/useExpressions';
+import { FullWidthWrapper } from 'src/features/form/components/FullWidthWrapper';
 import { RepeatingGroupsEditContainer } from 'src/features/form/containers/RepeatingGroupsEditContainer';
 import { RepeatingGroupTable } from 'src/features/form/containers/RepeatingGroupTable';
 import { FormLayoutActions } from 'src/features/form/layout/formLayoutSlice';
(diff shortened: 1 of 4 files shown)


src/features/form/containers/GroupContainer.tsx
```diff
@@ -189,6 +191,8 @@ export function GroupContainer({ id, container, components }: IGroupProps): JSX.
     return null;
   }
 
+  const isNested = typeof container.baseComponentId === 'string';
```
Magnusrm, line 194: Adding type for strict typechecking.
```suggestion
  const isNested: boolean = typeof container.baseComponentId === 'string';
```

src/features/form/containers/GroupContainer.tsx
```diff
@@ -233,58 +237,65 @@ export function GroupContainer({ id, container, components }: IGroupProps): JSX.
...
-                setEditIndex={setEditIndex}
-                onClickRemove={onClickRemove}
-                repeatingGroupDeepCopyComponents={repeatingGroupDeepCopyComponents}
-                forceHideSaveButton={true}
-              />
-            );
-          })}
+                return (
+                  <div
+                    key={index}
```
Magnusrm, line 277: I see you did not change this, but i think the id is safer to use as key here than the index.
```suggestion
                    key={`key-${id}`}
```

Magnusrm (approved):
Nice work! Thanks for the contribution 🙏  Some minor nitpicking below.

src/features/form/containers/GroupContainer.tsx
olemartinorg: Personally, I'm not a fan of "strict" typechecking. TypeScript is best when it's not in your way, and TypeScript is smart enough to figure out that the return value from a comparison is a boolean. There's no need to write anything more than we absolutely _have_ to. In the future I want to enable `noImplicitAny` in `tsconfig.json` so that we are nagged to specify types wherever TypeScript _can't_ automatically infer the type.

Magnusrm: I must have misunderstood some previous talk about typechecking then 😅 Thanks for clarifying! I have no problem personally with keeping the types inferred where they can be 👍

mijohansen: This is not code that this PR touches, so this is out of scope.

src/features/form/containers/GroupContainer.tsx
mijohansen: This is not code that is touched by this PR, so it's out of scope. If this need to be changed it could be made in another issue.

PR #861, opened 2023-01-25
Expressions optimization and streamlining

### LINKED ISSUE #822: Instance with many repeating group wont open
### Description of the bug

We have app: KRT-1012a-1 which is abo-schema (We instanciate to users mailbox). 
This app contains a reapeating group: 
image

If there are many repeating group (maybe 120 stk), it takes12 sek to open the group after clicking Rediger knapp. In general the entire app is very slow.

Also, it cant handle that many components in summary-page, and the entire summary-page appears white after it should iterate the repeating group here. 
image

The instance is: 
/instance/50164456/78c06a1a-9684-4af5-bc39-30b243733bb3

 Note. The repeating groups are empty and only contain one field of data (component 5.1). User has to fill in the rest. 

### Steps To Reproduce

1. Find the instance and try to run it ? 

### Additional Information

_No response_

olemartinorg:
## Description
Marking this as a product feature, because it opens up a few more things:
- I've tested out setting `title` and `tableTitle` in `textResourceBindings` via expressions now, which did not work correctly before (at least it crashed the app for me when I tested on earlier versions).
- Some things may be more consistent and as-expected after this change. Such as text resources for `title` when displaying a component in a `Summary` (which did not always seem to look up variables in text resources before, but does now).
- It should now be a lot more performant to use expressions overall, and less error-prone. Some needs (such as #754) are not as pressing anymore, as it's fairly easy to work around them using expressions (that, as mentioned, sometimes caused the app to crash before).

With that said, some bullet points detailing some of the more technical changes here:
- Instead of `useExpressions()` and `useExpressionsForComponent()`, most expressions are now evaluated once in the `ExprContext` (which just loads `resolvedNodesInLayouts()` and provides the result to all child components via two new hooks - `useExprContext()` and `useResolvedNode()`). This is what gives us the much-needed performance boost.
- Overall, more usage of the node hierarchy instead of constructing and manipulating component definitions/objects directly. This is especially true in `Summary`, where the node hierarchy now replaces most component config wrangling.  
- The node hierarchy could resolve expressions before, but it had no functionality to ensure that some expressions (such as `edit.deleteButton`) were evaluated per-row (as this is a property on the `Group` component, but could have different results per-row). Now, instead of just a default value (if the expression fails to evaluate), you have to pass in a configuration for every possible expression. This configuration tells the expression engine if an expression should be evaluated per-row or not, so expressions like `edit.deleteButton` will stay unresolved in the `Group` component and will be evaluated and available in the `node.item.row[123].groupExpressions.edit.deleteButton` property instead (for each row index). This also fixes an issue where these expressions were being resolved in unrelated code, and could end up failing (because, for example, a `component` lookup was not available when not in a row context), resulting in an error message in the developer console.

## Related Issue(s)
- closes #822
- #754 
- #377 
- #345

## Verification
- Manual testing
    - [x] Manually test the new PDF generator (we should have cypress tests for the automatic PDF layout)
    - [x] Manually test some prominent/production apps side-by-side with an older version
  - [x] Creator of the original issue (or service owner) has been contacted for manual testing
- Automated tests
  - [x] Cypress E2E test(s) have been ~added~ updated
  - [x] No automatic tests are needed here (this carries no functional _changes_, only minor new features that did not exist anymore. Tests were only changed to bootstrap correctly for them to work properly, but they were a much-needed aid when making these changes in order to verify existing functionality did not change. I have plans to add tests for some of the new functionality I noticed was working after this change.)
  - [ ] Has been updated
  
  - [x] No changes made
- Support in Altinn Studio
  
- Sprint board


src/components/summary/SummaryGroupComponent.tsx
```diff
@@ -94,27 +84,15 @@ function SummaryGroupComponent({
...
-        getComponentForSummaryGroup(state.formLayout.layouts[pageRef], componentRef)) ||
-      undefined,
-    shallowEqual,
-  );
+  const node = useResolvedNode(componentRef);
+  const textResourceBindings = node?.item.textResourceBindings;
+
+  const removeExcludedChildren = (n: AnyNode<'resolved'>) =>
+    !excludedChildren ||
+    !excludedChildren.includes(n.item.id) ||
```
bjosttveit, line 92: Does `n.item.id` include multiPageIndex?

bjosttveit (approved):
Looks good as far as I can tell 😅

olemartinorg:
@bjosttveit Yup, sorry, it's a huge change that affects everything... The fact that you spotted that multiPageIndex thing means you probably looked more closely at this than I would! :partying_face:

src/components/summary/SummaryGroupComponent.tsx
bjosttveit: Is it not correct that it should not contain it?

olemartinorg: Whoops, I misunderstood completely.. It's true that there was no backwards compatibility issue here. The `componentDefinition.id` property has never included the `multiPageIndex`, but items in `children` _might_. I.e., if you pass `excludedChildren={someGroup.children}`, you'd expect every child to be excluded, and they would - except if `someGroup.edit.multiPage === true`.

So I was thinking I might add a check for `excludedChildren.includes('${n.item.multiPageIndex}:${n.item.id}')` as well - to make multiPage groups `children` property work the same as non-multiPage groups. The more I think about it, I don't know if it's a valid/needed use case, so we should probably just skip it.

bjosttveit: Its probably not needed

PR #866, opened 2023-01-26
WCAG: Input purpose/autocomplete property

### LINKED ISSUE #27: WCAG: Identify Input Purpose
## Description
The app developer must to be able to insert an autocomplete tag in the input elements where this is appropriate.
This is a list of appropriate inputs that must have this tag: https://www.w3.org/TR/WCAG21/#input-purposes

**Intent**

> The intent of this Success Criterion is to ensure that the purpose of a form input collecting information about the user can be programmatically determined, so that user agents can extract and present this purpose to users using different modalities. The ability to programmatically declare the specific kind of data expected in a particular field makes filling out forms easier, especially for people with cognitive disabilities.

[Source](https://www.w3.org/WAI/WCAG21/Understanding/identify-input-purpose.html)

[A full reference of options](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete)

[A reference of norwegian standard on the input](https://www.uutilsynet.no/veiledning/135-identifiser-formal-med-inndata/1238)

## Acceptance criteria
- [ ] The purpose of a form input can be programmatically determined

mijohansen:
## Description

This PR is a part of the effort to close WCAG-issues, any extended comments and checks could be added by the receiving team if needed.

## Related Issue(s)

- closes #27

## Verification

- Manual testing
- Automated tests
  - [ ] Has been updated
  
  - [ ] No changes made
- Support in Altinn Studio
  
- Sprint board


schemas/json/layout/layout.schema.v1.json
```diff
@@ -122,29 +149,95 @@
...
+        },
+        {
+          "if": { "properties": { "type": { "const": "AttachmentList" } } },
+          "then": { "$ref": "#/definitions/attachmentListComponent" }
+        },
+        {
+          "if": { "properties": { "type": { "const": "Checkboxes" } } },
+          "then": { "$ref": "#/definitions/radioAndCheckboxComponents" }
+        },
+        {
```
olemartinorg, line 164: Please refrain from (auto-) formatting this file for us, it makes it harder to see the actual changes you made. If you'd like to format it, please do so in a separate PR to make it a separate discussion.

schemas/json/layout/layout.schema.v1.json
```diff
@@ -964,6 +1059,66 @@
           "$ref": "expression.schema.v1.json#/definitions/boolean"
         }
       }
+    },
+    "autocomplete": {
+      "type": "string",
+      "default": "off",
+      "enum": [
```
olemartinorg, line 1066: Would be nice to have some descriptive text here, preferably with a link to somewhere you can read more about the different choices! :blush: When working in vscode with no other context, `"autocomplete": "impp"` makes little sense.

Also, is the default really `off`? :thinking: In other words, do we block browser autocompletion by default?

olemartinorg (changes requested):

schemas/json/layout/layout.schema.v1.json
olemartinorg: I looked into the URL pointed to in #27, but it had no listing of on/off, at least.. Would be nice to have a source clarifying what those mean.

schemas/json/layout/layout.schema.v1.json
mijohansen: Hmm, guess its a good idea to turn on the linter for theses files then? Anyway will isolate the changes and manually format the file.

schemas/json/layout/layout.schema.v1.json
mijohansen: Have updated the issue with a link that also includes the `on` and `off` option.

Default could be `on` if that is what we want. @Febakke might help us answer what the default for this value should be.

olemartinorg: Correct me if I'm wrong here, but my cursory reading suggests that:
- `off` as the default instructs the browser to never suggest to autocomplete fields. Might be a bit harsh to turn that off for everyone overnight.
- `on` as the default encourages the browser to remember this field forever and _please, please_ suggest previously filled values for the user.

I haven't extensively tested it, so I may very well be wrong here, but I'm leaning against `on` as the default, or possibly just not define it by default (if that's an option).

schemas/json/layout/layout.schema.v1.json
mijohansen: These files doesnt change often so its a good idea. Files is fixed now so it should be easier to read the changes.

schemas/json/layout/layout.schema.v1.json
mijohansen: * the `impp` option is removed,
* the default is set top `on`

Febakke: What was it before we added this? Was the default `off` from before? 

The WCAG issue is only to add the attribute if field is one of the values from the list. So on `on` and `off` is not part of this issue. So best to keep it as is if possible.

olemartinorg: Allright, the WCAG issue decides it then - no attribute by default (seems like the safest option for backwards compatibility as well).

mijohansen: Ok, removed "required" and default values.

olemartinorg (approved):

Issue #178, opened 2022-05-27
Look into upgrading how app specific information is presented in our header


Febakke:
## Description
Our apps are getting more and more functionality, way beyond what was intended in the original design of the modal view we have today. In some future example we see that the space is getting cramped, is there anything we can do with the general design to hinder this? In a way we have two headers today. Our top "menu/header" shows the Altinn logo and who you are logged in as. (In the future also language will be chosen from here?) At this time we look at these elements as global, they do not change depending on what application the user is accessing. 

Inside the modal we have the app specific information. The name of the app and the app owner name. This is were we are putting in more features making it more cramped. This is the main content of the app and should be given space. Is there something we can do give the app specific elements more space? 

**Element to assess**
* App owner logo as a part of the global header?
* A lot of space between modal and header, are there other ways we structure it to give more focus to the main content?
* Expand design of modal or look into other ways of presenting the app specific information. 

## In scope
How will our current design hold up with future features being introduced?
Do our global / app specific information make sense how it set up today? 

## Tasks
- [ ] Research similar "powered by" solutions. How do they show the app owner / Platform owner
- [ ] How does the global header function today? Is it useful or just there because thats how the original design was?
- [ ] Map out the functionality and information that will most often be present in our apps. What is most important and should have focus?

PR #871, opened 2023-01-27
Footer

### LINKED ISSUE #110: As a user I would like to have correct contact info available in the app
## Description
Give the app owner the possibility to include contact info. This should be presented in a universal way, so that it is presented the same on all apps using the Altinn design. As a first version we propose that we create a footer that the app developer can populate with information.

## Tilgjengelighetserklæring
Kan denne footeren også inneholde lenke til tilgjengelighetserklæring?
## Screenshots
Skjermbilde 2021-10-25 kl  09 11 32

## Considerations
What kind of contact info do we need to support?
- Mail
- Phone
- Chatbot?

## Specification tasks
- [ ] Should we support tilgjengelighetserklæring in the footer?
- [ ] Create designs for all channels that can be shown in footer
- [ ] Mobile view

### LINKED ISSUE #538: As an app-developer I would like to add a link to our accessibility statement (Tilgjengelighetserklæring)
NB: This is not A & AA. I just want it to visible as it is important for the accessibility statement.
### Description

Add the link to our accessibility statement in a footer. 
Image
[Figma](https://www.figma.com/file/Pvti3aRcOwH0k5Z7DrLznk/Arbeidsomr%C3%A5de%3A-Altinn-Studio-APPS?node-id=4213%3A78024&t=WPn0BkUZcEDb1lcA-1)

We could solve this as a part of #110 or if it will save us some time, only handle the accessibility statement. 

### Additional Information

_No response_

### Tasks

- [x] Design how it should look / Take #110 into concideration here
- [x] How do the app-developer configure this? handled by Studio team in this issue: https://github.com/Altinn/altinn-studio/issues/9585

### Acceptance Criterias
- [ ] A link is available from the footer 
- [ ] Default link is to our accessibility page
- [ ] The app developer should be able to overwrite this link and send their user to their own statement

bjosttveit:
## Description

Removed the proposed grid-solution as I think it could be too complicated for a user friendly configuration in studio. Currently just a simple flexbox. Cypress tests depend on https://github.com/Altinn/app-lib-dotnet/pull/183 being released and frontend-test updating nugets.

## Related Issue(s)

- closes #538 
- closes #110 

## Verification

- Manual testing
- Automated tests
  - [x] Has been updated
    - https://github.com/Altinn/altinn-studio-docs/pull/836
  
  - [x] No changes made
- Support in Altinn Studio
  
    - https://github.com/Altinn/altinn-studio/issues/9810
- Sprint board


ivarne:
Sorry for not noticing this earlier, but I somehow feel this is very related to the issues about customizing the header.
* https://github.com/Altinn/app-frontend-react/issues/138
* https://github.com/Altinn/app-frontend-react/issues/178

Why did you go with a custom `footer.json` file instead of making this just a `footer` property of `applicationsettings.json` or `ul/../Settings.json`?

bjosttveit:
Those issues are indeed related, and are something we also want to allow customization of.

I suppose the reason we did not want to define it in `applicationMetadata.json` is that it would not be very flexible in case we wanted to allow different footers depending on the process step in the future. It cannot really be put into `Settings.json` since the footer needs to be available outside of the process, like instance selection. Yes, it would cause 404 in older apps as of now. 😕

bjosttveit:
We have discussed this again today, and we have decided to keep it in a separate file with the current setup for now. And then at a later point we can combine multiple files into a single request were appropriate to reduce the number of requests when loading the form.

ivarne:
OK: Looking forward to Application Insights being full of 404 errors looking for a footer.
image

schemas/json/layout/footer.schema.v1.json
```diff
@@ -0,0 +1,82 @@
+{
+  "$id": "https://altinncdn.no/schemas/json/layout/footer.schema.v1.json",
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "Altinn footer",
+  "description": "Schema that describes the footer configuration for Altinn applications.",
+  "type": "object",
+  "additionalProperties": true,
```
olemartinorg, line 7: Personally, I'd like for us to start disallowing additional properties everywhere. Our form layout definition allows additional properties, and it has made it difficult to notice that we forgot to add schema definitions for some new property. IMO, if you add some value to the footer schema that we never expected (and will not handle in any way), that should also present you with an error in vscode.

schemas/json/layout/footer.schema.v1.json
```diff
@@ -0,0 +1,82 @@
...
+    "emailComponent": {
+      "properties": {
+        "title": { "$ref": "#/definitions/title" },
+        "target": {
+          "type": "string",
+          "title": "Target",
+          "description": "The email to link to."
+        }
+      }
+    },
```
olemartinorg, line 54: I assume both `title` and `target` should be required? Same goes for all the other footer components (none of their properties are required). A `{"type": "Link"}` doesn't really mean much.

schemas/json/layout/footer.schema.v1.json
```diff
@@ -0,0 +1,82 @@
...
+        "icon": { "$ref": "#/definitions/icon" }
+      }
+    },
+    "phoneComponent": {
+      "properties": {
+        "title": { "$ref": "#/definitions/title" },
+        "target": {
+          "type": "string",
+          "title": "Target",
+          "description": "The phone number to link to."
```
olemartinorg, line 72: Can you use text resource keys in here? Same goes for the `title` property. We should probably include that information in the description.

src/features/form/layout/formLayoutSlice.ts
```diff
@@ -33,6 +35,7 @@ export interface ILayoutState {
   error: Error | null;
   uiConfig: IUiConfig;
   layoutsets: ILayoutSets | null;
+  footerLayout: IFooterLayout | null;
```
olemartinorg, line 38: I think we could create an entirely new slice for the footer. In my head, the formLayout slice is tied to the process, while the footer is something we could (and should) fetch before we start thinking about the process/instance.

src/utils/urls/appUrlHelper.ts
```diff
@@ -116,6 +116,7 @@ export const getLayoutSettingsUrl = (layoutset: string | null | undefined) => {
 };
 
 export const getLayoutSetsUrl = () => `${appPath}/api/layoutsets`;
+export const getFooterLayoutUrl = () => `${appPath}/api/footer`;
```
olemartinorg, line 119: A bit late to comment this, but some other apis use `/api/v1/<name>` which I think is better. We've already made `footer.schema.v1.json` so there might be a footer v2 in the future where we'd like to have another API endpoint as well.

olemartinorg (changes requested):

olemartinorg:
> OK: Looking forward to Application Insights being full of 404 errors looking for a footer.

@ivarne Same thing when looking for `api/v1/featureset` in #899 soon. We'll discuss this on our end, to try to find a solution. IMO, 404 is barely an "error" (reminder that 4xx means "you did something wrong" while 5xx is "I did something wrong", and if lots of 4xx responses on an app trigger all the alarm bells it would be _very_ easy for any malicious actor to make those alarm bells go off).

schemas/json/layout/footer.schema.v1.json
bjosttveit: Right, should include that in the description (for title at least, have not added text-resource support for target, do you think it should have?)

schemas/json/layout/footer.schema.v1.json
bjosttveit: Right, totally forgot about required for the components, will fix

src/utils/urls/appUrlHelper.ts
bjosttveit: I can change that, app-lib PR is not merged yet

src/features/form/layout/formLayoutSlice.ts
olemartinorg: Yup, I guess that could work! Thinking more about it, it might just be confusing to have the footer layout being loaded into the form layout state, when those two layout definitions have nothing to do with each other.

schemas/json/layout/footer.schema.v1.json
olemartinorg: Not sure if it is needed, but if it's easy to implement it might add value. One immediate use-case I can think of is having different phone numbers depending on language.

bjosttveit: Would be very easy

ivarne:
@olemartinorg Yes, but `api/v1/featureset` is one endpoint that will reduce noise in the future. `api/v1/footer` is yet another minor feature that shouldn't need to probe for the existence of an api.

src/features/footer/data/footerLayoutSlice.ts
```diff
@@ -0,0 +1,38 @@
...
+import type {
+  IFetchFooterLayoutFulfilled,
+  IFooterLayoutActionRejected,
+  IFooterLayoutState,
+} from 'src/features/footer/data/types';
+import type { MkActionType } from 'src/shared/resources/utils/sagaSlice';
+
+export const initialState: IFooterLayoutState = {
+  footerLayout: null,
+  error: null,
```
olemartinorg, line 12: Is this ever read? What happens if the footer API responds with a catastrophic failure? Hint: Look into `getHasErrorsSelector`

olemartinorg (approved):
Looks good! :rocket:

I think we should have automatic tests for this thing (as I believe all features should have some form of automatic tests :wink:). You can even use Cypress to mock the API response to test out different footers. I used that approach in #899, which made it a joy to develop the functionality (as I could have Cypress open in a different window where it tests the entire feature I'm in the process of making, while I'm making it - it caused me to find and fix several edge-cases). Without tests, anyone can break this functionality without noticing.

src/features/footer/data/footerLayoutSlice.ts
bjosttveit: Aha, maybe did it a little to quickly 😅

bjosttveit:
Nice, I assumed I needed to update frontend-test with a footer and updated nugets to test with cypress. If it is possible to mock I can do it before merging 👍

src/selectors/getErrors.ts
```diff
@@ -24,6 +24,7 @@ const getHasErrorsSelector = (state: IRuntimeState) => {
     state.profile.error ||
     state.language.error ||
     state.formLayout.error ||
+    state.formLayout.error ||
```
olemartinorg, line 27: It should probably be the footer layout state error, not the form layout error twice. :grin:

bjosttveit: Wow 🤦 nice catch

bjosttveit: looks like my typescript extension bugged out and could not find the file, so the autocomplete messed me up 😅

PR #881, opened 2023-02-01
Add steps to release pipeline to publish to azure CDN

tjololo:
## Description

Using the existing release script to upload the files controlled by app-frontend-react to the new cdn solution.
For now both the old and the new CDN should be updated, once we are truly confident in the setup we can remove the old.

The new steps require som secrets, will mark the PR as ready when they are available.

## Related Issue(s)

- relates to: [altinn/altinn-studio#7566](https://github.com/Altinn/altinn-studio/issues/7566)

## Verification

- Manual testing
- Automated tests
  - [ ] Has been updated
  
  - [ ] No changes made
- Support in Altinn Studio
  
- Sprint board

### DIFF
--- a/.github/scripts/release.sh
+++ b/.github/scripts/release.sh
@@ -195,7 +195,9 @@ else
     schemas_rsync_opts=( -am --include='*/' --include="component/*" --include="layout/*"  --exclude='*' )
     set -x
     rsync "${toolkits_rsync_opts[@]}" $TARGET $AZURE_STORAGE_ACCOUNT_NAME
-    rsync "${schemas_rsync_opts[@]}" $TARGET_SCHEMAS/json $AZURE_STORAGE_ACCOUNT_NAME -v
+    if [[ "$PRE_RELEASE" == "no" ]]; then
+      rsync "${schemas_rsync_opts[@]}" $TARGET_SCHEMAS/json $AZURE_STORAGE_ACCOUNT_NAME -v
+    fi
     set +x
     echo "-------------------------------------"
   else
--- a/.github/scripts/release.sh
+++ b/.github/scripts/release.sh
@@ -176,11 +176,11 @@ fi
 
 echo "-------------------------------------"
 if [[ -z "$AZURE_STORAGE_ACCOUNT_NAME" ]]; then
-  echo "Skipping publish to azure cdn. As --azure-cdn-uri flag not defined"
+  echo "Skipping publish to azure cdn. As --azure-sa-name flag not defined"
 else 
   if [[ -d "$AZURE_STORAGE_ACCOUNT_NAME" ]]; then
     echo
-    echo "azure-cdn-uri seems to be a local directory. Simulating azcopy sync with rsync to folder"
+    echo "azure-sa-name seems to be a local directory. Simulating azcopy sync with rsync to folder"
     echo
     toolkits_rsync_opts=( -am --include='*/' --include="${APP_FULL}/*" )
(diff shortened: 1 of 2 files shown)


olemartinorg (approved):
Hmm, a few questions..

Won't this rely on the existing CDN repo being there? Right now it copies the entire `toolkits` folder, which contains all the previous versions as well as the new one. If we pull out the rug by removing the git repo backing from the old CDN, would that mean we would overwrite the `toolkits` folder on azure with one that _only_ contains the newest version? If new files are merged in, that won't become a problem.

tjololo:
Good questions as always, since the azure step leverages copy no files will be deleted, but if a file is updated it will be overridden. The downside is that this will do a full copy of all the files every time we release so I think I will need to change that.
Might need to extend the release.sh with an --azure (or something) argument where the logic to copy only the new files to azure.
This will keep all the release logic inside the release.sh script. If you agree I can make these changes to the script and request a new review after I'm done

olemartinorg:
@tjololo Yup, I agree! Part of the reason I made the `release.sh` script is that it's actually possible to run it locally in order to test out what would happen. That makes changes much easier to verify, so I'm all for adding to that script instead of putting things on the outside.

As an aside, will we at some point remove the git-backing for CDN? Right now the repo is fairly large for a git repo, and we would like to make it a bit larger (by adding source maps), but so far we've held that back in anticipation of the new CDN. Maybe we could get usage reports from azure in the future to make it possible to remove older (unused) versions? It's most likely not very critical we keep multi-year old patch releases.

tjololo:
Updated the release.sh with logic for syncing with azure.
The new functionality still piggybacks on the copy logic against altinn-cdn, when we are confident in the new azure cdn we can replace the logic with scripts that copy all the files to be published to a folder and sync it. Then this release logic would not need to fetch all the old versions just sync the new version.
The sync operation filters out and only uploads folders matching the regexes and compares the md5 hashes to determin if a sync is needed on those files.
The script has three options for running:
1. `--azure-sa-name <name>` set to an existing folder would trigger an rsync to this folder with the files to be synced (this is an best effort simulation
2. `--azure-sa-name <name>` set to a storageaccount name and `--azure-sa-token <sas-token>` with a token for that storage account. As long as `--azure-sync-cdn` is not present it will perform an dry-run and output what files would be updated in the storageaccount. Needs acces to the storage account but will display changes that would actually be made
3. `--azure-sa-name <name> --azure-sa-token <token> --azure-sync-cdn` performes the actual sync against azure

.github/scripts/release.sh
```diff
@@ -26,6 +27,25 @@ while [[ $# -gt 0 ]]; do
       COMMIT=yes
       shift # pop option
       ;;
+    --azure-cdn-uri)
+      AZURE_CDN_URI=$(realpath "$2")
+      shift # pop option
+      shift # pop option
+      ;;
```
olemartinorg, line 34: This is references below, but the variable seems to be unused? Was this left over from a round of cleanup? :thinking:

tjololo: Correct, removed

olemartinorg: What about the references below? It now says things like `As --azure-cdn-uri flag not defined`, but there is no such flag anymore. Probably best to do another pass over the comments. :pray:

tjololo: nice catch, not on my game today it seems 🙈

olemartinorg (approved):

Issue #286, opened 2021-11-08
Implement support for copying an instance from old in app frontend


acn-sbuad:
## Description

The workflow should be that triggering copy in SBL should send a request to app frontend. That again builds the request to app backend and renders the newly created instance. 

## Considerations
> Describe input (beyond tasks) on how the user story should be solved can be put here.

### Ops requirements
> Are there any requirements for monitoring? What is being built and what could go wrong? 
> Are there any requirements related to backup? 

## Acceptance criteria
- [ ] Frontend respons to an endpoint that triggers creation of a new instance based on an old

## Specification tasks

- [ ] Development tasks are defined

## Development tasks
> Add tasks here

## Test
> Add test cases here as checkboxes that are being tested as part of the changes.

## Definition of done
Verify that this issue meets [DoD](https://digdir-digitalisering.atlassian.net/wiki/spaces/T3KP/pages/5049246/Definition+of+Done) (Only for project members) before closing.

- [ ] Documentation is updated (if relevant)
  - [ ] Technical documentation (docs.altinn.studio)
  - [ ] User documentation (altinn.github.io/docs)
- [ ] QA
- [ ] Manual test is complete (if relevant)
- [ ] Automated test is implemented (if relevant)
- [ ] All tasks in this userstory are closed (i.e. remaining tasks are moved to other user stories or marked obsolete)

FinnurO:
Digital gravferdsmelding

olemartinorg:
Relevant:
- https://www.altinn.no/hjelp/innboks/arkiv/
- https://github.com/Altinn/altinn-roadmap/issues/67
- https://github.com/Altinn/altinn-studio/issues/1566

annerisbakk:
Tatt ut av gr.leggende funksj. Følges opp i App

RonnyB71:
Closing, solved by https://github.com/Altinn/altinn-platform/issues/355

PR #886, opened 2023-02-01
Show presentation texts and dueBefore on the select-instance page

ivarne:
This builds on 
- https://github.com/Altinn/app-lib-dotnet/pull/179
- https://github.com/Altinn/app-lib-dotnet/pull/177

and exposes that information in the frontend. It is fully backwards compatible with older versions of nuget packages (but will only show the new columns on newer package versions).

TODO:
* Find a better header for `PresentationTexts` (I went with `Texts`)
* Fix mobile layout to only show the new columns when they have values.
* Fix mobile layout. It looks ugly.

Images:
image
image

## Verification

- Manual testing
- Automated tests
  - [ ] Has been updated
  
  - [ ] No changes made
- Support in Altinn Studio
  
- Sprint board

### DIFF
--- a/src/features/instantiate/containers/InstanceSelection.tsx
+++ b/src/features/instantiate/containers/InstanceSelection.tsx
@@ -23,7 +23,8 @@ export interface IInstanceSelectionProps {
   onNewInstance: () => void;
 }
 
-function getDateDisplayString(timeStamp: string) {
+function getDateDisplayString(timeStamp?: string) {
+  if (!timeStamp) return '';
   let date = new Date(timeStamp);
   const offset = date.getTimezoneOffset();
   date = new Date(date.getTime() - offset * 60 * 1000);
@@ -75,11 +76,21 @@ export default function InstanceSelection({ instances, onNewInstance }: IInstanc
                 items={[
                   {
                     key: 1,
+                    label: getLanguageFromKey('instance_selection.presentation_texts', language),
+                    value: Object.values(instance.presentationTexts ?? {}).join(' '),
+                  },
+                  {
+                    key: 2,
+                    label: getLanguageFromKey('instance_selection.due_before', language),
+                    value: getDateDisplayString(instance.dueBefore),
+                  },
+                  {
+                    key: 3,
                     label: getLanguageFromKey('instance_selection.last_changed', language),
                     value: getDateDisplayString(instance.lastChanged),
                   },
                   {
-                    key: 2,
+                    key: 4,
                     label: getLanguageFromKey('instance_selection.changed_by', language),
                     value: instance.lastChangedBy,
                   },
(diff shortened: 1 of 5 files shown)


Febakke:
I do agree with the action point that the mobile view could be better. If there is room for doing changes to it I can look into a more readable design?

ivarne:
I'm not sure. This view is very analogous to the "message box", but limited to just one app, so they should probably be looked at together. My tentative wish list (for functionality, I'm not a designer) would be something like.

* Display Last Changed and Last Edited By (always)
* If availible display PrestentationValues and DueDate (but don't make a huge visual clue that they are missing)
* Button to continue working on the app (if not ended)
* Optional (by app configuration) button to create a new instance with a copy of the data
   * This would kind of make more sense if we had tabs for active and archived instances. Obviously archived instances would link to the Receipt instead of the edit view.
* `New Instance` button is hidden (or disabled) if the current user isn't allowed to create a new instance.

There are reasons to think some of this belongs in the message box, but configurability and dynamic functionality is much easier when it's an app specific thing.

Febakke:
> Button to continue working on the app (if not ended)

Could you elaborate on this?

> Optional (by app configuration) button to create a new instance with a copy of the data
> This would kind of make more sense if we had tabs for active and archived instances. Obviously archived instances would link to the Receipt instead of the edit view.

I see that we have an issue related to this here: #286 
To hide not relevant information for the user another way to solve this would be to give the user a choice from the start: 
- Start new instance
- Continuo 
- Or make a copy of earlier instance

That way we could focus on more simple views. But this is a more time consuming task.

As of design, we have done some work on the mobile view for the table for the design system. It is added to some parts of Altinn 3. Maybe it should be added here as well. Or am I way of @olemartinorg ?

Tables for small screens are not easy. If we would use the design from the design system it would look something like this. It is a little heavy on the left side when there is no radio button so we might need to do some spacing changes. And it takes more vertical space, but it is more readable then the one shown here. 
Skjermbilde 2023-02-06 kl  12 46 48

olemartinorg:
> As of design, we have done some work on the mobile view for the table for the design system. It is added to some parts of Altinn 3. Maybe it should be added here as well. Or am I way of @olemartinorg ?

Nope, you're spot on! :wink: I'll create a separate issue to describe that change. See here: #897

ivarne:
I'd probably use "Tabs" (eller "Fane") instead of a radio button to decide whether to show active or archived instances.

I'm aware of https://github.com/Altinn/app-frontend-react/issues/286, but as I understand it the plan there is to somehow integrate it with the inbox that is shared with altinn 2. I'm just raising it as an issue here, because it might be relevant as part of a bigger picture if we decide to do something more than "Just extending the current functionality".

>> Button to continue working on the app (if not ended)
>
>Could you elaborate on this?

Just the same "Fortsett her" button that we have today. If extend the view to also include instances where the process has finished, the button to continue obviously should not be shown for those instances.

Febakke:
I wasent thinking of radiobuttons, I was more thinking of different flows based on the task of the user, keeping the GUI simple. Someone that is going to start an instance from scratch dont need to see already started and archived instances. But this is something we can look into 🙌

ivarne:
I don't think there is any context to derive the current task of the user from when they access the root url of the app without any parameters indicating a choice made in the inbox or wherever the user found the link. So far I know 4 possibilities (not all relevant for every app).

1. Create a new instance
2. Continue on an existing instance with more steps left in the process
3. Create a new instance based on a copy of a previous instance
4. Show a stateless view of the app (some apps)

Febakke:
This is maybe outside the scope of this PR, but I guess we lack design for handling more functionality on this "start" page? I can create an issue to start looking on some different possibilities based on your comments?

ivarne:
I’m beginning to think I should just remove the mobile changes in this PR, remove header on presentationTexts, and say this PR is a strict improvement. Further work should be moved to an issue.

olemartinorg:
Removing this PR from our sprint - but just let me know when it's ready for review again, and we'll include it in our planning. 😊

Issue #555, opened 2022-10-18
Validation trigger on group should re-validate on delete + new row


olemartinorg:
### Description
Adding `"triggers": ["validation"]` on a repeating group component does not have very well-defined behaviour. It should be possible to add this trigger, and have server-side validations re-run on the group data. In the related Slack discussion, an app was set up to validate that a group had at least 1 row (using server-side validations, but this should be supported by the JsonSchema validation as well).

Triggers on a repeating group should run when:
- Deleting a row (or rather, after a row has been deleted - as it might fail to delete)
- Adding a new row (in order to run server-side validations here, we'd have to fill out some empty data into the row components and pass that to the server)

The validations on the group should *not* run when saving data inside a component, as that should only trigger validation if that trigger is set on the component itself. 

### Related issue(s)
- #468 
- #269
- Discussion on Slack: https://altinn.slack.com/archives/C03DX42K6AY/p1666075681994599
- Other related dicussion: https://altinn.slack.com/archives/C02EVE4RU82/p1665565549630379 (consider if the need to trigger something on adding a new row is a different need, and should be implemented differently)
- Yet more discussion: https://altinn.slack.com/archives/C03DX42K6AY/p1675249074411239 (a use-case where it would be useful to know when a row is opened for editing, a new row is added, and potentially a trigger for deleting a row).
- **NB!** Will potentially be solved by #1175

RonnyB71:
Potential breaking change and should maybe be pushed to a new major version.

olemartinorg:
- #339

FinnurO:
There are other issues related to validation which could be useful to look at in order to get a complete view of the validation area.

olemartinorg:
I have an update on this, and a suggestion for a less breaking solution. It was originally thought we needed to introduce a breaking change for this to work, because adding a new row to a repeating group should (naturally) set all the fields in the group to empty values (and pass this as a single save operation to the backend). However, we can continue on with the current functionality and add this behaviour as an optional feature (in the form of a trigger) without breaking backwards compatibility.

My new suggestion for a solution here is to add multiple triggers for operations the user is performing. I've read through the relevant discussions again, making sure this solution should cover the use-cases.

So, right now this is our possibilities:

```json
{
  "type": "Group",
  "triggers": ["validation", "validateRow"]
}
```

[...]

PR #899, opened 2023-02-07
Saving datamodel along with list of changed fields and previous values

### LINKED ISSUE #810: Repeating groups: ProcessDataWrite mangler info om hva som har trigget eventet
### Description of the bug

I ProcessDataWrite eventet så kan jeg vanligvis se hvilket datafelt eller componentid som har trigget ProcessDataWrite via Request Headers. Men hvis jeg sletter en rad i en repgruppe så ser jeg ikke dette.
Hvis vi skal kunne benytte dette til å manipulere data så trenger vi og å vite hva som har trigget dette, enten i form at datamodel felt eller componentid.

### Steps To Reproduce

Opprett en side med repgruppe , stopp koden i ProcessDataWrite og sjekk etter info ved sletting av rad i repgruppen

Man kan hente ut commit b82277c573258348049252ab3b083460db7452b9 fra https://altinn.studio/repos/stian.vestli/ra0244-01
og teste dette her.

### Additional Information

_No response_

olemartinorg:
## Description

**Note:** This requires backend support, which has not arrived yet. See the related issues for more details.

When saving, we currently pass along a best-effort indication of which component ID performed the operation, but this is not sufficient info for the backend at times, and we can't guarantee that we send in the correct component ID here (because saving to the backend isn't, and shouldn't be tied 1-1 to saving in a component).

The way this works is that is queries a 'feaureset' endpoint in the start of the application runtime (which currently returns a 404 on all backend versions). If that endpoint responds, with an object like `{"multiPartSave": true}`, subsequent save operations will be a multipart PUT request with two parts (`dataModel` and `previousValues`). The `dataModel` part is the same as what we saved before, while `previousValues` will be a key-value object of dot-notation keys, along with the previous values for all changed data model paths (or `null` if the value did not exist before). The former `X-DataField` and `X-ComponentId` headers are not passed along in a multipart request, and should be considered deprecated.

## Related Issue(s)

- closes #810
- #506
- #555

Backend implementation (currently missing):
- Altinn/app-lib-dotnet/issues/187
- Altinn/app-lib-dotnet/issues/188

## Verification

- Manual testing
- Automated tests
  - [ ] Has been updated
  
  - [x] No changes/updates needed (no user-facing or developer-facing features here, but the backend implementation should be documented)
  - [x] No changes made
- Support in Altinn Studio
  
- Sprint board


ivarne:
> The former X-DataField and X-ComponentId headers are not passed along in a multipart request, and should be considered deprecated.

These headers is not used by the library, but directly by apps. Apps will likely just break silently in a way that is unlikely to be covered by tests and often makes schemas impossible to complete or cause issues with invalid data being submitted. Can you try to explain why these can't be kept for backwards compatibility?

olemartinorg:
Yup. I don't know all the details, so I'm leaning on people who do. I also suggested in the linked issue for a backend implementation that this should be optional/opt-in: Altinn/app-lib-dotnet/issues/187

With that said, I can elaborate on the reason I believe those headers should be deprecated; I really want to go in a direction where saving data is more asynchronous than it is right now. Let's say you have a spotty internet connection - I would want saving to occur whenever it is possible (but requiring a stable connection for things like triggers, and actual final submits). Right now, we run save requests all the time, and if one of them fails, the entire app crashes. The assumption that "changing something in one component == one save request" is broken, IMO - and getting rid of it now can let us decouple this code in app-frontend in the future. I've written more about this in #506 and #555. Also here: https://github.com/Altinn/app-frontend-react/issues/339#issuecomment-1321920974

ivarne:
I totally agree with your plans 👍😍. The current semantics is just as annoying to me too.

I just think we need to be very careful about breaking backwards compatibility, and preferably have some kind of safety. Maybe we can just search gitea for the names of the headers and open issues in the probably very small number of apps that uses the functionality?

ivarne:
Haha, and it seems like I read the issues in the wrong order. You explained the issue pretty well in in the app-lib-dotnet issue😳

olemartinorg:
I did a quick search in my folder of about ~341 apps deployed on prod and tt02, and only 5 of them seem to use the headers at the moment. Or, really just 3 (some were deployed in multiple environments and were counted multiple times because of that). So probably not a big problem. But yeah, I agree that should be wary of breaking backwards compatibility. Though, this won't come into effect unless you upgrade the nuget packages - and we have fairly well-established documentation for what you need to be aware of when upgrading those.

bjosttveit (approved):
## Nice work 👨‍🔧

PR #900, opened 2023-02-07
WCAG: Added `aria-expanded` to repeating group edit button

### LINKED ISSUE #898: WCAG: Screen reader does not tell the user that our groups open and closes
### Description of the bug

Because our repeating group is a specially made component we need to be careful to use aria the right way so that assistive tools know how to handle it. Right now there are no indication what happens when you toggle the group open and closed. 

I think we need to use the method described here: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-expanded#buttons

### Steps To Reproduce

1. Go to an app with a repeating group
2. Use a screen reader
3. try to open and close a group. You should not get any information of the state of the widget. 

### Additional Information

_No response_

bjosttveit:
## Description

Added an `aria-expanded` attribute to the edit button in repeating groups that is set to true when the group row is open for editing. It also sets `aria-control` to the `RepeatingGroupsEditContainer`, but only when the element is open, since it is not present in the DOM when closed.

## Related Issue(s)

- closes #898 

## Verification

- Manual testing
- Automated tests
  - [ ] Has been updated
  
  - [x] No changes made
- Support in Altinn Studio
  
- Sprint board

### DIFF
--- a/src/features/form/containers/RepeatingGroupTableRow.tsx
+++ b/src/features/form/containers/RepeatingGroupTableRow.tsx
@@ -237,12 +237,14 @@ export function RepeatingGroupTableRow({
           >
             <div className={classes.buttonInCellWrapper}>
               <Button
+                aria-expanded={isEditingRow}
+                aria-controls={isEditingRow ? `group-edit-container-${id}-${index}` : undefined}
                 variant={ButtonVariant.Quiet}
                 color={ButtonColor.Secondary}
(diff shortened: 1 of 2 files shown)


olemartinorg (approved):
:rocket:

olemartinorg:
Is this related to #901 as well, @bjosttveit? They seemed similar from the descriptions, but I'm assuming they're subtly different?

Issue #896, opened 2023-02-06
New PDF enhancements


bjosttveit:
### Description

During a discussion with Eirik, some points about enhancing the new PDF generator came up:

- [ ] Support grid layout (if chromium can handle the flex-layout)
- [x] More page-break options like `avoid` (maybe support the whole css spec instead of boolean like today)
  - [x] Also add support for page break in automatic layout
  - https://github.com/Altinn/app-frontend-react/pull/916
- [ ] Show components differently than in summary? (more compact?)

RonnyB71:
Closing this as a duplicate of https://github.com/Altinn/app-frontend-react/issues/1744

PR #916, opened 2023-02-09
PDF: Added more page-break options

bjosttveit:
## Description

This changes the pageBreak options from a boolean to 'auto' (default, equivalent to false previously), 'always' (equivalent to true previously), and 'avoid'. This matches the css spec. Another change is that `pageBreak` can be added to components in the form so that it also works in the Automatic generation method.

This came up in a discussion with Eirik, and given that these are "breaking changes" compared to what was there previously, I would like to get this through before posting the updated documentation which I suppose amounts to "officially" releasing the new PDF feature.

## Related Issue(s)

- https://github.com/Altinn/app-frontend-react/issues/896

## Verification

- Manual testing
- Automated tests (**Separate issue**)
  - ~~[ ] Unit test(s) have been added~~
  - ~~[ ] Cypress E2E test(s) have been added~~
  - ~~[ ] No automatic tests are needed here~~
  - ~~[ ] I want someone to help me make some tests~~
  - [x] Has been updated (will be updated before release)
  
  - [x] Changes are reflected in both `src/layout/layout.d.ts` and `layout.schema.v1.json`, ~~and these are all backwards-compatible~~
  - [ ] No changes made
- Support in Altinn Studio
  
- Sprint board

### DIFF
--- a/schemas/json/layout/layout.schema.v1.json
+++ b/schemas/json/layout/layout.schema.v1.json
@@ -962,15 +962,17 @@
       "properties": {
         "breakBefore": {
           "title": "Page break before",
-          "description": "PDF only: Boolean value or expression indicating if a page break should be added before the component.",
-          "default": false,
-          "$ref": "expression.schema.v1.json#/definitions/boolean"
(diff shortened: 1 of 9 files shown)


Magnusrm (approved):
Very nice! 👍

PR #942, opened 2023-02-21
Error message - remove long ID, duplication fix

### LINKED ISSUE #444: FileUploadWithTag does not show tagTitle value in error messages
### Description of the bug

When not choosing a tag in the fileUploadWithTag component, the error message does not contain the tagTitle value from resources. It either displays "tagTitle" or [object object], depending on how you trigger validation.

### Steps To Reproduce

1. Set up a valid FileUploadWithTag component
2. Set a value for textResourceBinding.tagtitle that is defined as an id in resources 
3. Run and open schema
4. Upload a file to the FileUploadWithTag component
    - Do **NOT** choose a tag
5. Trigger validation by pressing "Lagre" or "Neste"
6. See faulty error message without tagTitle value

### Expected result:
An error message appears that contains the actual value of the resource. 
(In our case "Du må velge datatype" since it is defined in resources as '{"id": "tagTitle", "value": "Datatype"}')

### Actual result:
After clicking "Lagre": error message is "Du må velge [object object]".
After clicking "Neste": error message is "Du må velge tagtitle", where "tagtitle" is the id of the value.
The bottom error message also contains a long ID, where it for other components only contains the error message.

### Tried:
- Using different tagTitle values that are defined -> Same result.
- Entering a string that is not defined as an ID in resources -> The string is displayed in both situations (also instead of "[object object]"). 
    The long ID is still present in the bottom div.
- Updated to newest releases -> Same result.
- Tested locally with newest pushes for studio and frontend -> Same result.

### Notes
The tagTitle value is available to the component as it is displayed as the title of the tag dropdown menu.
I cannot find any obvious errors in the code, but I believe it is related to validation.

### Images
After pressing only "Lagre"
fileUploadWithTag_lagre

After pressing only "Neste"
fileUploadWithTag_neste

After pressing both "Lagre" and "Neste"
fileUploadWithTag_both

Example of expected result, (from input field)
expected_error_input
expected_error_input_end

lassopicasso:
## Description

This PR addresses two bugs related to file upload with tag.

- ID being displayed in error message.
- While working on the ID-bug above I encountered a second bug where the error message being duplicated if user try to save without a tag and then try to navigate to next page. 

The ID being displayed in the error message/report also includes an ascii unit. Added errorMessage function to check if this is included in the string. If that is the case a substring is returned and otherwise return string as is.
Added cypress tests to secure that this functionality doesn't break on a later stage.

For the duplicated error, I use the Map object to only display error messages that contains unique ID's for file upload elements. 
`let uniqueObjArray = [...new Map(objArray.map((item) => [item["id"], item])).values()];`
For more info about Map: [MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map)

## Related Issue(s)

- closes #444 

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
  - [x] No changes made
- Support in Altinn Studio
  
- Sprint board
- Labels

### DIFF
--- a/src/components/message/ErrorReport.tsx
+++ b/src/components/message/ErrorReport.tsx
@@ -68,6 +68,14 @@ export const ErrorReport = ({ components }: IErrorReportProps) => {
     return null;
   }
 
+  if (errorsMapped.length > 0) {
+    errorsMapped.forEach((error, index) => {
+      if (error.componentId === 'fileUploadWithTags-changename') {
+        errorsMapped[index].message = error.message.slice(37);
+      }
+    });
+  }
+
   const handleErrorClick = (error: FlatError) => (ev: React.KeyboardEvent | React.MouseEvent) => {
     if (ev.type === 'keydown' && (ev as React.KeyboardEvent).key !== 'Enter') {
       return;
--- a/src/components/message/ErrorReport.tsx
+++ b/src/components/message/ErrorReport.tsx
@@ -9,6 +9,7 @@ import { FullWidthWrapper } from 'src/features/form/components/FullWidthWrapper'
 import { renderLayoutComponent } from 'src/features/form/containers/Form';
 import { FormLayoutActions } from 'src/features/form/layout/formLayoutSlice';
 import { getLanguageFromKey, getParsedLanguageFromText } from 'src/language/sharedLanguage';
+import { AsciiUnitSeparator } from 'src/utils/attachment';
 import { nodesInLayouts } from 'src/utils/layout/hierarchy';
 import { getMappedErrors, getUnmappedErrors } from 'src/utils/validation/validation';
 import type { ILayout } from 'src/layout/layout';
@@ -66,14 +67,6 @@ export const ErrorReport = ({ components }: IErrorReportProps) => {
 
   if (!hasErrors) {
     return null;
-  }
-
-  if (errorsMapped.length > 0) {
-    errorsMapped.forEach((error, index) => {
-      if (error.componentId === 'fileUploadWithTags-changename') {
-        errorsMapped[index].message = error.message.slice(37);
-      }
-    });
   }
 
   const handleErrorClick = (error: FlatError) => (ev: React.KeyboardEvent | React.MouseEvent) => {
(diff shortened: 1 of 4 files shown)


lassopicasso:
I would much appreciate it if you guys could give me some feedbacks on my code. I'm a little unsure about these solutions I came up with. ;)

src/components/message/ErrorReport.tsx
```diff
@@ -166,9 +159,16 @@ export const ErrorReport = ({ components }: IErrorReportProps) => {
...
+                      {error.message.includes(AsciiUnitSeparator)
+                        ? getParsedLanguageFromText(
+                            error.message.substring(error.message.indexOf(AsciiUnitSeparator) + 1),
+                            {
+                              disallowedTags: ['a'],
+                            },
+                          )
+                        : getParsedLanguageFromText(error.message, {
+                            disallowedTags: ['a'],
+                          })}
```
bjosttveit, line 171: Maybe you could separate out this logic into a function that takes in error.message and returns the substring if the separator-symbol is present and otherwise just returns the string as is? I think the code would look cleaner.

olemartinorg (commented):
Looks good! 🚀 

I had to look into this `AsciiUnitSeparator` thing, and to me it seems like our data structures for storing validation messages does not allow for metadata such as 'which one of multiple attachments is this error message for' (which is what the attachment uploader needs). In my opinion that's what we should fix (the root of the problem), not just work around it by supporting the separator in `ErrorReport.tsx`. Would you mind creating an issue for that (mark it with a label to indicate that it's tech debt), so we can clean it up properly?

In the mean time, since we now know more about this separator character, maybe we should also add some tests to make sure the character don't appear wherever we display error messages? That is, when the ErrorReport is visible after uploading an attachment with tags, this character shouldn't be visible inside it (make sure the test fails without this fix as well), and also that it doesn't show up in the file upload component itself. You never know, some day someone else will come along and break this functionality so that the separator shows up again. 🤷‍♂️ It would be nice to catch that in testing.

lassopicasso:
I agree with you on the issue @olemartinorg, will create an issue for this and also add some tests to secure that the functionality won't break in the mean time. :) Thanks for the feedback!

PR #943, opened 2023-02-21
Shifting complexity in expressions/hierarchy

olemartinorg:
## Description
This PR is rather large, and contains somewhat complex type-changes. I apologize for the lack of readability! All in all, the main points of interest in this PR:

- I'm removing the `NodeType` type that was used everywhere in `hierarchy.types.d.ts`. The point of this type was to make it possible to switch between resolved/unresolved properties (basically supporting having the full hierarchy tree represented with dynamic expressions intact, or with them evaluated). After #861, there's almost no usage of unresolved layout hierarchies, as accessing the full resolved tree got to be cheap and easy.
- With no real need for all the extra functions in `hierarchy.ts`, I'm moving some of them to a `_private` export, meant to symbolize that they're not really meant to be used outside of where they're in use now (unit tests). Please use the end result (the fully resolved, deep tree/hierarchy of components), not any of the intermediate steps.
- The `ExprResolved` type we used to have was cool and all, but it performed somewhat poorly. When combining it with other types, especially `MaybeSpecificItem` in `ExprContext` (again, coming from #861), you quickly reached the limits of TypeScript (or some other limit, who knows), and adding another layer depth of complexity to these types would break them. I don't like that the new, shining, and more future-proof framework we're building on is hard to use and extend, and after bumping my head against this thing, I decided to simplify the types for the layout hierarchy as much as I could, to hopefully invite further developments in that direction. To do that, I had to move/shift the complexity to the old-school methods of manipulating simple components/layouts directly. This is most clearly visible in this PR by the next bullet point:
- The types used for layout components (`ILayoutComp*`, `ILayoutComponent`) used to be defined such that they could contain either a concrete value (i.e. `required: true`) or an expression (i.e. `required: ["equals", true, false]`). This has now changed, and layout components are defined _to have places that can contain expressions_, but in order to actually use such an object, you at some point have to indicate to the type-system weather or not you expect these expressions to be resolved or not. Before, resolved expressions could be assignable directly to an `ILayoutComp*` type, and you could sometimes get away with just using the layout redux state directly (in case you didn't have any expressions in the layout). This could cause bugs and requires code that is expression-aware to be the most complex. I've moved this complexity so that almost every code using the layout types need to wrap them in either `ExprResolved<...>` (usually via `HComponent` or `AnyItem`/`LayoutNode`) or `ExprUnresolved<...>`. for resolved and unresolved (as-from-the-api) expressions, respectively. 
- With `NodeType` gone, and with the complexity moved to legacy-style code via a forced `ExprUnresolved<...>` type, I found and deprecated a whole lot of code that I think should go away when we slowly clean up usages. It should also discourage further use of the older functionality, and encourage usage of the layout hierarchy.

## Related Issue(s)
- #377
- #345
- #861

## Verification/QA

    - (That means, I tested `evalExpression()` in the console, which had minor changes, and I tested and fixed the failing cypress tests)
- Automated tests
  - [ ] Has been added/updated
  
  - [x] No changes made
- Support in Altinn Studio
  
- Sprint board
- Labels


src/components/summary/SummaryComponent.tsx
```diff
@@ -192,9 +192,13 @@ export function SummaryComponent(_props: ISummaryComponent) {
         )}
       />
     );
-  } else if (layoutComponent?.getComponentType() === ComponentType.Presentation) {
+  } else if (
+    layoutComponent?.getComponentType() === ComponentType.Presentation &&
+    formComponentLegacy.type !== 'Summary' &&
+    formComponentLegacy.type !== 'Group'
+  ) {
```
bjosttveit, line 199: I don't think checking that the type is not `Summary` or `Group` is necessary here as they do not have a ComponentType, so they would fail because `null !== ComponentType.Presentation`.

bjosttveit (approved):
Looks good to me as long as the tests are passing, can't really tell just by looking 😅

src/components/summary/SummaryComponent.tsx
olemartinorg: Checked now, and it's still needed. TypeScript doesn't know that `layoutComponent` and `formComponentLegacy` is supposed to be the same thing (in many ways), so checking the type of `layoutComponent` does not narrow the type of `formComponentLegacy` (which is used in the implementation below). So these checks will make sure to narrow the type of `formComponentLegacy` so that `GenericComponent` can render it.

PR #959, opened 2023-02-27
Rename data type to "organization" to match data in store

mariepauline:
## Description

Different spelling of type organization causes the data to be inaccessible in code. In the store "organization" is used and this PR renames every instance of the word to this.

Redux store:
image

This is just a suggestion, but I thought it was better to provide a fix than to just create an issue.

## Verification/QA

  - [X] I have tested these changes manually
- Automated tests
  - [X] Unit test(s) have been added/updated
  - [X] No testing done/necessary (no DOM/visual changes)
  - [ ] Has been added/updated
  
  - [X] No functionality has been changed/added, so no documentation is needed
  
  - [X] No changes made
- Support in Altinn Studio
  
  - [X] This change/feature does not require any changes to Altinn Studio
- Sprint board
  - [X] Not relevant
- Labels
  - [X] I don't have permissions to add labels, please help me out

### DIFF
--- a/src/shared/components/altinnAppHeader.test.tsx
+++ b/src/shared/components/altinnAppHeader.test.tsx
@@ -13,7 +13,7 @@ describe('AltinnAppHeader', () => {
   it('should show organisation name when profile has party, and party has organisation with name, and "type" is not set', () => {
     const profile = getProfileStateMock();
     if (profile.profile.party) {
-      profile.profile.party.organisation = organisationMock;
+      profile.profile.party.organization = organisationMock;
     }
 
     render({
@@ -27,7 +27,7 @@ describe('AltinnAppHeader', () => {
   it('should not show organisation name when profile has party, and party has organisation with name, and "type" is set', () => {
     const profile = getProfileStateMock();
     if (profile.profile.party) {
-      profile.profile.party.organisation = organisationMock;
+      profile.profile.party.organization = organisationMock;
     }
 
     render({
--- a/src/shared/components/altinnAppHeader.tsx
+++ b/src/shared/components/altinnAppHeader.tsx
@@ -170,14 +170,14 @@ export const AltinnAppHeader = ({ type, profile, language }: IHeaderProps) => {
                     <span className={`d-block ${blueClass}`}>{renderParty(profile)}</span>
                     <span className={blueClass}>
                       {party &&
-                        party.organisation &&
-                        `${getLanguageFromKey('general.for', language)} ${party.organisation.name.toUpperCase()}`}
+                        party.organization &&
+                        `${getLanguageFromKey('general.for', language)} ${party.organization.name.toUpperCase()}`}
                     </span>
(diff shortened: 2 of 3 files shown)


olemartinorg:
@mariepauline 
Hi, and thanks for the contribution! 🙏 I checked one of the renamed locations, and in the `applicationmetadata.json` file, we set `partyTypedAllowed.organisation` (not `organization`) so this PR would seemingly break that functionality when we're loading the file from existing apps. Maybe you could limit the PR to just rename the places affected, and/or describe the problem you're having? Thanks again!

mariepauline:
@olemartinorg Yes, I can look into limiting the changes :) 

The problem is when trying to access data about organization from state.
image
`organisation` on `IParty` becomes undefined currently, since it is called `organization` in the store. I haven't dived too much into where the mismatch happens but a rename seems to fix it.  
image

The naming of organisation was set 3 years ago, so maybe there is another way to change it to make it work? I guess the only thing that's important is that it is the same all over

mariepauline:
I see that in order for just making this work, it's possible to just change in `IParty` (and where it is used), but then there is a mismatch in the repo so it is not a very good long term solution.
image

olemartinorg:
> The problem is when trying to access data about organization from state

This is the first time I'm hearing of someone relying on our internal state! 🤔 We have no stability guarantees on that (on the contrary, it's pretty much guaranteed to change). I'd love to hear more about what you are working on, and in what way you're building on our baseline app frontend. 😊 If the redux state is something we should take care to keep somewhat backwards stable as a contract to your project(s), that could affect our roadmap/backlog, as we have concrete suggestions to change it (see #345).

> it's possible to just change in IParty (and where it is used), but then there is a mismatch in the repo so it is not a very good long term solution

If by mismatch, you mean that it's spelled differently different places, I think we'll just have to live with that. We should be careful when rolling out backwards incompatible changes, and I don't think uniform spelling is a good enough justification for that. I agree that changing the property name in `IParty` should be good enough for this specific issue. 🙏

mariepauline:
@olemartinorg Well, we're not really 😅 I'm just playing around with the code to see how we can start on our app for Brønnøysundregistrene. So a meeting where we can present our case would be great! 😄 
I just found the problem and we thought we could flag it, but we are not blocked by it so it is no rush for us to have it fixed.

olemartinorg:
I see! Looking forward to hear more about your project. 🤩 

As for this issue, if you reduce the change to renaming the property in `IParty`, I'll gladly approve it and merge it into main. 😁 Thanks for finding this problem and figuring it out!

olemartinorg:
@mariepauline I'm marking this as draft for now. Go ahead and mark it ready for review if you get around to it again, or let me know if you want to close it.

mariepauline:
@olemartinorg I have now changed only IParty and tested it as well as I could

olemartinorg (approved):

PR #962, opened 2023-02-27
Update helptext to use digdir designsystem and fix wcag issues

### LINKED ISSUE #728: help text for radio-buttons
### Description of the bug

Help-text for radio-buttons component appears in bold, but for other components appears normal . 
image

Help-text for input-component: 
image

### Steps To Reproduce

add helptext for radio-component and input-component and run the program.

### Additional Information

_No response_

### LINKED ISSUE #817: Helptext not reachable with screen reader
### Description of the bug

You cant read content in a helptext (tooltip) after it has been expanded. This should maybe not read automaticly, but you should be able to reach it via keyboard navigation. 

This is tested with both NVDA and VoiceOver

### Steps To Reproduce

1. Go to an app with a help text
2. Use a screen reader and navigate the app using your keyboard
3. Try to get the screen reader to read the text within the tooltip.
When it is open it should be within reach of the screen reader

### Additional Information

- [ ] Design task: Research what went wrong and how other systems handle this problem

### Sources
[Solution suggestion](https://www.accessibility-developer-guide.com/examples/widgets/tooltips/#manually-displayed-tooltip)
[Navs helptext component](https://aksel.nav.no/komponenter/core/helptext)

Magnusrm:
## Description

Removes old implementation of HelpText and implements HelpText component from Digdir's designsystem.

## Related Issue(s)

- closes #817 
- closes #728 
## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
  - [x] No changes made
- Support in Altinn Studio
  
- Sprint board
- Labels


src/features/form/components/HelpTextContainer.test.tsx
```diff
@@ -0,0 +1,43 @@
...
+        language={language}
+        helpText={'Help text content'}
+        {...props}
+      />,
+    );
+  };
+
+  it('receives correct accessible title', () => {
+    render();
+    screen.debug();
```
olemartinorg, line 28: Might not need this in the final codebase?

src/utils/stringHelper.test.ts
```diff
@@ -12,4 +13,19 @@ describe('stringHelper', () => {
...
+    expect(getLabelFromChildren(getParsedLanguageFromText('<span>This is my message</span>'))).toEqual(
+      'This is my message',
+    );
+    expect(getLabelFromChildren(getParsedLanguageFromText('<div><span>This is my message</span></div>'))).toEqual(
+      'This is my message',
+    );
+
+    const myReactNode: React.ReactNode = getParsedLanguageFromText('<div><span>This is my message</span></div>');
+    expect(getLabelFromChildren(myReactNode)).toEqual('This is my message');
+  });
```
olemartinorg, line 30: Cool stuff! Does it work for `<div><span>text1</span><span>text2</span></div>` as well? Might add a test-case for that to be sure, and document what is supposed to happen in that case.

olemartinorg (commented):
Nice! If we decided to remove rich text/HTML support from help texts, that makes this a breaking change for existing apps that possibly rely on it. What's the reasoning behind that decision, and have you looked into existing apps to make sure the fallout is acceptable?

Magnusrm:
> 

Rich text in help-texts should still be supported in this implementation. Looked into a case where links and line-breaks were used in help-texts and tested that specific configuration with this result:
image

src/utils/stringHelper.test.ts
Magnusrm: It does unfortunately not work for that example. The output is an array with `[<span>text1</span>, <span>text2</span>]`. Mabye the function could be updated to run again for each item in the array? 🤔

olemartinorg: An array? Huh, that was unexpected - the TypeScript return type says only strings will be returned. I would expect an empty object (or null) as input would also return a string, then (empty string, maybe). I would suggest adding such cases to the test and adjust the typescript return type (or implement it such that the example input returns `text1text2` as expected).

olemartinorg (approved):
🚀 Looks great!

PR #976, opened 2023-03-07
Massive cleanup in component rendering/generation

olemartinorg:
## Description
This PR includes an overhaul to the way we render layout components. Until now, we've had the legacy code that generates component properties, resolves data model bindings/mapping/text resource bindings for repeating groups, and in the end injects these properties into `GenericComponent` to render it. After this PR, some things will change under the hood:

- When rendering `GenericComponent`, a `node` object is passed into it. You can no longer render imaginary components that do not belong anywhere in the node hierarchy (and by extension, exist inside the redux store). This makes sure we develop new functionality with the node hierarchy and dynamic expressions in mind.
- A few properties can still be overridden when rendering a component (and Summary), but we should avoid polluting the code with custom properties, as expressions are not aware of these.
- The `LayoutComponent` class has now been split into 3: `PresentationComponent`, `FormComponent` and `ActionComponent` (for buttons, navigation). These 3 component types may get different functionality in the future, and make sure we implement the important parts for every component type (i.e. we have to implement a function to get the component data to display as a string in the repeating group table, and every form component also needs to be able to render itself as a Summary item).
- `Summary` has been majorly refactored in this PR, and now relies on the node hierarchy throughout. As mentioned, every component now decide for themselves how it should be rendered in a Summary.
- Lots of deprecated functionality and dead code has been removed. 
- Some unit tests that were hard to fix has been removed - we should prioritize Cypress tests instead, as those tend to survive these kinds of refactoring jobs that do not alter functionality.

## Related Issue(s)
- #377

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
  - [x] No changes made
- Support in Altinn Studio
  
- Sprint board
- Labels


bjosttveit:
# 🤯🤯🤯🤯

bjosttveit (approved):
Christmas came incredibly early this year 🤩

src/layout/FileUpload/shared/summary.ts
```diff
@@ -0,0 +1,52 @@
...
+      if (attachmentsForComponent) {
+        const foundAttachment = attachmentsForComponent.find((a) => a.id === uuid);
+        if (foundAttachment && foundAttachment.name) {
+          return foundAttachment;
+        }
+      }
+
+      return null;
+    })
+    .filter((a) => a !== null) as IAttachment[];
```
framitdavid, line 29: What do you think about refactoring a little to avoid filtering null and casting the type? Would it be a little more readable to follow the suggestion? The suggestion has moved `attachmentsForComponent `outside the loop to avoid creating and assigning on every iteration. :) 

```suggestion
  const attachmentsForComponent = attachments[componentId];

  if (!attachmentsForComponent) {
    return [];
  }

  let componentAttachments: IAttachment[] = [];

  uuids.forEach((uuid) => {
    const foundAttachment = attachmentsForComponent.find((a) => a.id === uuid);
    if (foundAttachment?.name) {
      componentAttachments = [...componentAttachments, foundAttachment];
    }
  });

  return componentAttachments;
```

src/layout/GenericComponent.tsx
```diff
@@ -287,7 +252,14 @@ export function GenericComponent<Type extends ComponentExceptGroupAndSummary>(
     label: RenderLabel,
     legend: RenderLegend,
     componentValidations,
-    ...props,
+  };
+
+  const componentProps = {
+    ...fixedComponentProps,
+    // TODO: Pass on the node object instead of all the properties in it. This could work fairly simply for most
```
framitdavid, line 259: Should we create a technical debt issue and put it on the backlog to ensure this will be fixed and not continues as a for every living todo? 🤔 :D

src/layout/MultipleSelect/index.tsx
```diff
@@ -1,11 +1,35 @@
...
+    if (!node.item.dataModelBindings?.simpleBinding) {
+      return {};
+    }
+
+    const value = formData[node.item.dataModelBindings.simpleBinding] || '';
+    return useCommaSeparatedOptionsToText(node.item, value);
+  }
+
+  useDisplayData(node: LayoutNodeFromType<'MultipleSelect'>): string {
+    return Object.values(this.useSummaryData(node)).join(', ');
```
framitdavid, line 28: I see we are using `Object.values(some-input).join(', ');` in many places within our codebase. Have we considered to create StringUtils.ts to make our codebase a little more readable by using descriptive utils functions instead? :D

What about creating a util-method and using it like this? 🤔
`return toCommaSeperatedString(this.useSummaryData(node)); `

src/layout/Panel/PanelGroupContainer.tsx
```diff
@@ -162,31 +147,25 @@ export function PanelGroupContainer({ container, components }: IPanelGroupContai
...
-
-                  {referencedGroup && (
-                    <Grid item>
-                      <SuccessIconButton
-                        id={`save-reference-button-${container.id}`}
-                        label={getLanguageFromKey('general.save', language)}
-                        onClick={handleSave}
-                      />
-                    </Grid>
+                  {/*  // TODO: Add test case for filling out a new row in panel, not saving it, and midway through*/}
```
framitdavid, line 150: Should we create an issue on the backlog to ensure this TODO get prioritized? 🤔 :D

framitdavid (commented):
First of all, you have done a lot of quality and great work! 🚀  This will help the team a lot to maintain and understand the code base. I'm happy to see your changes, thank you!! 👏 

I have only a few considerations and thoughts which I commented on this PR.

.eslintrc.json
```diff
@@ -169,6 +169,12 @@
       "rules": {
         "@typescript-eslint/no-var-requires": ["off"]
       }
+    },
+    {
+      "files": ["src/layout/*/index.tsx"],
+      "rules": {
+        "react-hooks/rules-of-hooks": "off"
```
framitdavid, line 176: It feels a little strange/painful to have to turn this rule off, but as long we are using classes I don't think this is an issue. It 
 might be a sign that we should implement our code differently, but I'm not sure. The rule is mainly used to ensure that hooks is initialized in top scope and for classes this rule does not make sense.

src/layout/Button/index.tsx
```diff
@@ -1,20 +1,15 @@
...
+import { ActionComponent } from 'src/layout/LayoutComponent';
 import type { PropsFromGenericComponent } from 'src/layout';
 
-export class Button extends LayoutComponent<'Button'> {
+export class Button extends ActionComponent<'Button'> {
   render(props: PropsFromGenericComponent<'Button'>): JSX.Element | null {
     return <ButtonComponent {...props} />;
   }
 
   renderWithLabel(): boolean {
```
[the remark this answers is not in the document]
olemartinorg: Not sure what you mean by re-implement when needeed. This functionality is in use, as some components expect `GenericComponent` to render them with a label first, while others do not. The default is then to render with label (returns true), but the Button component should be rendered without, thus is overrides the base implementation and returns false instead.

The philosophy here, I think, is that new components have to implement some functions where there are no sane defaults (making them `abstract` in the base class, so that not implementing them is a compile-time error). For others, such as this one, the sane default is to always render with a label, but if that's not what you wanted you can override the method and provide your own result.

So for Summary, every `FormComponent` now needs to implement a `renderSummary` (and so every developer needs to active think about "how should this component be displayed in a summary?" when implementing a new component type). Previously, most components have not even been tested in a Summary, because there was a default (but not always very functional) implementation - which lived far away from your component implementation.

Not sure if that all made sense, but the TL;DR is that no, we can't remove this thing - it's in use.. 😊

src/layout/Checkboxes/index.tsx
```diff
@@ -1,15 +1,37 @@
...
   renderWithLabel(): boolean {
     return false;
   }
+
+  private useSummaryData(node: LayoutNodeFromType<'Checkboxes'>): { [key: string]: string } {
+    const formData = useAppSelector((state) => state.formData.formData);
+    const value = node.item.dataModelBindings?.simpleBinding
+      ? formData[node.item.dataModelBindings.simpleBinding] || ''
+      : '';
+    return useCommaSeparatedOptionsToText(node.item, value);
```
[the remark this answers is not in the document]
olemartinorg: I agree completely! Will move these hooks into the `hooks` folder.

src/layout/GenericComponent.tsx
olemartinorg: Yup, I agree! Will do that.

src/layout/Panel/PanelGroupContainer.tsx
olemartinorg: Yeah, I'll do that. My plan was to do that ASAP, as the Panel thing turned out to be horribly broken (but was just as broken before this PR, so I didn't break it). Better to create an issue in case I don't get around to it, so I'll do that.

.eslintrc.json
olemartinorg: Yup, I know.. 😆 Little on the fence on this one as well. I guess we could separate out the methods that are meant to be used as hooks, but I'm not sure I like that approach either.

src/layout/Button/index.tsx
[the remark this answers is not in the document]
framitdavid: Of course, it all makes sense now. Thanks for the clarification.  🚀

.eslintrc.json
framitdavid: Sounds like a pair-programming issue to me, to discuss the different approaches that could be implemented. Anyway, this is not a task for this PR. 😃

src/layout/GenericComponent.tsx
olemartinorg: Ah, sorry, was about to create that issue, but it seems I did not properly read which TODO this was. I already have a working fix for this, in relation to the rest of the work I'm doing in #377. Fixing this would add ~1000 extra lines of changes to this PR, so I kept it out of this PR to keep it "small". No need for an issue, in other words.. 😆

src/layout/Panel/PanelGroupContainer.tsx
olemartinorg: - #985
It got rather large in the end, when I wrote down everything I was thinking about. Good thing it's documented in an issue now, so thanks for that nudge!

src/layout/FileUpload/shared/summary.ts
olemartinorg: Finally looked at closely at this, and yup, that's much better! I took the liberty to rewrite it to use `for (const uuid of uuids)` and `componentAttachments.push()` as well. 🥳 Thank you!

framitdavid: No worries! Sounds great to use the push method, then componentAttachments can be const instead of let. Great work! 🥳

PR #984, opened 2023-03-09
Expand form width button

### LINKED ISSUE #95: Button to expand app modal to full width
## Description
For some use cases the current app modal is too narrow to present the data to the user in a reasonable way. We should support expanding the modal to fit the whole width.

See figma scetches: https://www.figma.com/file/wnBveAG2ikUspFsQwM3GNE/Altinn-Studio-Apps?node-id=15394%3A23195

## Screenshots
image

## Acceptance criteria

- Expand/minimize button enabled 
## Specification tasks

- [x] Development tasks are defined
- [ ] Should this be default or opt-in (setting in settings.json?)

## Definition of done

- [ ] Documentation (docs.altinn.studio) is updated (if relevant)
  - [ ] Technical documentation
  - [ ] User documentation
- [ ] QA
- [ ] Manual test is complete (if relevant)
- [ ] Automated test is implemented (if relevant)
- [ ] All tasks in this userstory are closed (i.e. remaining tasks are moved to other user stories or marked obsolete)

bjosttveit:
## Description

This adds a new option in `Settings.json` to enable a full width toggle button. This works by defining css variables for the page's max-width and padding. These css-variables are overridden by conditionally adding a class to the `Presentation` component that overrides these variables for all of its children. Previously, the page max width and the modal padding was hardcoded many different places, but now this is all refactored to use the css variables instead.

I also took this oppurtunity to refactor and get rid of material-ui `makeStyles` (and replace with css modules) in most of the places that I updated, and also refactored/updated many of the old components used in the Presentation such as the NavBar (back, close, language-selector). I also removed a lot of mui `Grid`, and using css modules to style the flex layout instead.

Finally I changed some layout surrounding the summary component, to make the `Edit` button positioning work better, as it looked pretty bad in the expanded view.

https://user-images.githubusercontent.com/47412359/223988190-3a602210-7821-4711-bf13-ce50516f9199.mov

## Related Issue(s)

- closes #95 

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
  - [ ] No changes made
- Support in Altinn Studio
  - [?] Issue(s) created for support in Studio
  
  - [?] This change/feature does not require any changes to Altinn Studio
- Sprint board
- Labels


src/components/presentation/NavBar.test.tsx
```diff
@@ -101,7 +101,7 @@ describe('NavBar', () => {
       showLanguageSelector: false,
     });
     expect(screen.queryAllByRole('button')).toHaveLength(0);
-    expect(screen.queryByTestId('altinn-back-button')).toBeNull();
+    expect(screen.queryByTestId('form-back-button')).toBeNull();
```
framitdavid, line 104: Personally, I think we should follow the best practices from React Testing Library. What do you think? I see that test-id was used before your changes. 🙈 

```suggestion
    expect(screen.queryByRole('button', { name: 'button-text' })).toBeNull();
```

RTL Docs has great documentation on their recommendations https://testing-library.com/docs/queries/about/#priority

src/components/presentation/NavBar.test.tsx
```diff
@@ -123,7 +123,9 @@ describe('NavBar', () => {
     });
     await waitForElementToBeRemoved(screen.queryByRole('progressbar'));
     const dropdown = screen.getByRole('combobox', { name: /Språk/i });
-    await act(() => userEvent.selectOptions(dropdown, 'en'));
+    await act(() => userEvent.click(dropdown));
+    const en = screen.getByText(/Engelsk/i, { selector: '[role=option]' });
+    await act(() => userEvent.click(en));
```
framitdavid, line 128: The act should not be necessary, because RTL has already wrapped those events within an act internally.
```suggestion
    await userEvent.click(en);
```

src/components/presentation/NavBar.tsx
```diff
@@ -13,51 +16,77 @@ export interface INavBarProps {
...
 export const NavBar = (props: INavBarProps) => {
-  const hideCloseButton = useAppSelector((state) => state.formLayout.uiConfig.hideCloseButton);
+  const dispatch = useAppDispatch();
   const language = useAppSelector((state) => state.language.language || {});
-  const theme = useTheme();
+  const { hideCloseButton, showLanguageSelector, showExpandWidthButton, expandedWidth } = useAppSelector(
+    (state) => state.formLayout.uiConfig,
+  );
+
+  const handleExpand = React.useCallback(() => {
```
framitdavid, line 28: This method should not be necessary to cache in memory, the method is invoked when clicking a button. I can't see any performance improvements by wrapping this within useCallback. 😄

src/features/form/components/FullWidthGroupWrapper.tsx
```diff
@@ -1,31 +1,16 @@
...
-      data-testid='fullWidthGroupWrapper'
-    >
-      {children}
-    </div>
-  );
-}
+export const FullWidthGroupWrapper = ({ children }: IFulLWidthGroupWrapperProps) => (
+  <div
+    className={css.fullWidth}
+    data-testid='fullWidthGroupWrapper'
```
framitdavid, line 12: Is this testid used somewhere? 🤔 I could not find where it is used. 😃

src/features/form/containers/RepeatingGroupsEditContainer.tsx
```diff
@@ -1,18 +1,17 @@
...
 
 import { Button, ButtonColor, ButtonVariant } from '@digdir/design-system-react';
-import { Grid, makeStyles } from '@material-ui/core';
+import { Grid } from '@material-ui/core';
 import { Back, Delete as DeleteIcon, Next } from '@navikt/ds-icons';
 import cn from 'classnames';
 
 import { useAppSelector } from 'src/common/hooks/useAppSelector';
 import { renderLayoutNode } from 'src/features/form/containers/Form';
+import css from 'src/features/form/containers/RepeatingGroup.module.css';
```
framitdavid, line 10: What do you think about renaming CSS to classes? Because the "objects" contains a set of class names.
```suggestion
import classes from 'src/features/form/containers/RepeatingGroup.module.css';
```

src/layout/InstantiationButton/InstantiationButtonComponent.tsx
```diff
@@ -16,18 +16,16 @@ const rowStyle = {
...
-    <div className='container pl-0'>
+    <div
+      className='a-btn-group'
+      style={btnGroupStyle}
+    >
       <div
-        className='a-btn-group'
-        style={btnGroupStyle}
+        className='row'
+        style={rowStyle}
```
framitdavid, line 25: What do you think about creating a CSS module for this component instead of using global CSS? Then we do not need the inline styling. I needed to think twice about why it should be necessary to set `marginLeft: 0`, because that should be the default on a `<div/>` tag. Then I recognized that the global row class have other values than preferred for this component. What do you think? 😄 
```suggestion
        className={classes.row}
```

I think it would be nice if we do not have any global CSS for such things, but instead have it in CSS modules. Then we achieve separation of concerns. I think it would be easier to maintain the CSS within the project if it's not global. If we need some spacing, and theming that is the same throughout our components, we could use CSS variables for that.

src/layout/Button/ButtonComponent.tsx
```diff
@@ -59,28 +57,26 @@ export const ButtonComponent = ({ mode, ...props }: IButtonProvidedProps) => {
...
-              {getLanguageFromKey('general.save', props.language)}
-            </SaveButton>
-          )}
-          <SubmitButton
-            onClick={() => submitTask({ componentId: props.id })}
-            id={props.id}
-            language={props.language}
+    <div className={css['button-group']}>
+      <div className={css['button-row']}>
+        {autoSave === false && ( // can this be removed from the component?
```
framitdavid, line 62: Good question, I don't know if it can be removed or not 😢 Maybe @olemartinorg can help us out here? 🤔

framitdavid (commented):
@bjosttveit you have a very nice job here! 🚀 I love to see that you have converted makeStyles to css-modules.

I have a few thoughts, but feel free to not change them all if you disagree with me. Some of the comments/thoughts might be more personally preferences... 🙈

src/components/presentation/NavBar.module.css
```diff
@@ -0,0 +1,19 @@
...
+}
+
+.wrapper {
+  display: flex;
+  flex-direction: row;
+  align-items: flex-end;
+  gap: 1rem;
+}
+
+button.buttonMargin {
```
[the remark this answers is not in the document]
bjosttveit: I totally agree, but there is something weird with the css specificity with the design system buttons, so it would be overridden by something internal (i think) if i left it out. The alternative was to use !important, or of course find the root of the problem in the design system 😅

olemartinorg:
I'll let David, king of code reviews, take the lead on this review! 👑 

While I assume it hasn't been in the requirements, I would assume there comes a new requirement after this for the layout to be expanded by default. Would that be a reasonable expansion to the properties?

Also, I haven't tested this yet, but does the button disappear when the viewport is mobile? At that point it can't expand, so it would make sense for the button to be hidden. And, of course test that in the `mobile.ts` cypress test.. 😉

bjosttveit:
> I'll let David, king of code reviews, take the lead on this review! 👑 
> 
> 
> 
> While I assume it hasn't been in the requirements, I would assume there comes a new requirement after this for the layout to be expanded by default. Would that be a reasonable expansion to the properties?
> 
> 
> 
> Also, I haven't tested this yet, but does the button disappear when the viewport is mobile? At that point it can't expand, so it would make sense for the button to be hidden. And, of course test that in the `mobile.ts` cypress test.. 😉 

Making the button disappear for small widths is a good idea especially if it is not already expanded. 

I have talked about default expansion with Lasse and Marianne, and that could indeed become a requirement. But as most of this PR is just refactoring to make it possible, we thought it was ok to leave that for a future issue. It should be relatively easy to add more controls for the expanded-state in uiconfig.

src/components/presentation/NavBar.module.css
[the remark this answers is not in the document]
framitdavid: ah, then keep it as it is! Better than !important. 🙈

PR #987, opened 2023-03-09
Making Group and Summary regular components, only node passed to components

olemartinorg:
## Description
Yet another installment after #976 made `GenericComponent` possible render with just the `node` object as input. This time:

- All component classes now only receive `node` from `GenericComponent`, meaning this PR takes this concept one step further. Components no longer should require every component property as input when rendering themselves. A lot of tests had to be rewritten to make this work.
- `Group` and `Summary` has been promoted (or demoted?) to regular components, which render via `GenericComponent`. No more need for a super-function that decides if a component can be rendered using `GenericComponent` or directly rendered (any component can now ask to be directly rendered).
- No need for `SummaryComponentSwitch` anymore, as every component type can now implement how it is to be rendered in a Summary instead of the Summary component having to know about different component types.

## Related Issue(s)
- #976 
- #377

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
  - [x] No changes made
- Support in Altinn Studio
  
- Sprint board
- Labels


src/features/form/containers/Form.tsx
```diff
@@ -4,67 +4,12 @@ import Grid from '@material-ui/core/Grid';
...
-import { PanelGroupContainer } from 'src/layout/Panel/PanelGroupContainer';
 import { ReadyForPrint } from 'src/shared/components/ReadyForPrint';
 import { extractBottomButtons, hasRequiredFields } from 'src/utils/formLayout';
 import { useExprContext } from 'src/utils/layout/ExprContext';
 import { getFormHasErrors, missingFieldsInLayoutValidations } from 'src/utils/validation/validation';
-import type { ComponentExceptGroupAndSummary } from 'src/layout/layout';
-import type { LayoutNode } from 'src/utils/layout/hierarchy';
-import type { LayoutNodeFromType } from 'src/utils/layout/hierarchy.types';
-
-export function renderLayoutNode(node: LayoutNode) {
```
framitdavid, line 21: Nice clean up! 🚀

src/layout/AttachmentList/AttachmentListComponent.tsx
```diff
@@ -41,7 +41,7 @@ export function AttachmentListComponent(props: IAttachmentListProps) {
         className='attachmentList-title'
         component='span'
       >
-        {props.text || ''}
+        {text || ''}
```
framitdavid, line 44: I think we don't need to fallback to an empty string here, because null or undefined should not be rendered. 🤔 
```suggestion
        {text}
```

src/layout/Checkboxes/CheckboxesContainerComponent.tsx
```diff
@@ -80,21 +80,15 @@ const defaultOptions: IOption[] = [];
 const defaultSelectedOptions: string[] = [];
 
 export const CheckboxContainerComponent = ({
-  id,
```
framitdavid, line 83: Well done with the props, looks much better to group them within the node prop. 👏

src/layout/FileUploadWithTag/EditWindowComponent.tsx
```diff
@@ -152,7 +153,7 @@ export function EditWindowComponent(props: EditWindowProps): JSX.Element {
               disabled={saveIsDisabled}
               className={classNames(classes.select, 'custom-select a-custom-select', {
                 'validation-error': props.attachmentValidations.filter((i) => i.id === props.attachment.id).length > 0,
-                'disabled !important': props.attachment.updating ? true : props.readOnly,
+                'disabled !important': props.attachment.updating ? true : readOnly,
```
framitdavid, line 156: I think this expression could be simplified to simething like the suggestion?
```suggestion
                'disabled !important': props.attachment.updating || readOnly,
```

_Looks like I just cleaned my glasses. Since I look at small things like this..._

src/layout/GenericComponent.tsx
```diff
@@ -256,10 +257,8 @@ export function GenericComponent<Type extends ComponentExceptGroupAndSummary = C
 
   const componentProps = {
     ...fixedComponentProps,
-    // TODO: Pass on the node object instead of all the properties in it. This could work fairly simply for most
```
framitdavid, line 259: 👏

src/layout/Group/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+        overrides={overrides}
+      />
+    );
+  }
+
+  renderSummaryBoilerplate(): boolean {
+    return false;
+  }
+
+  useDisplayData(): string {
```
framitdavid, line 40: Not related to this PR, but should this method have been renamed to getDisplayData? Within the React ecosystem, I associate the prefix "use" with a react-hook that uses other hooks. By reading the code within this class it looks like a normal function.

But again, maybe some other classes that are also inherent in the FormComponent actually use hooks within the useDisplayData method. 

No change is required - these were just some thoughts while reading the code.

framitdavid (approved):
As always a lot of great work! 🚀

src/components/summary/SummaryComponent.tsx
```diff
@@ -132,13 +137,27 @@ export function SummaryComponent({ summaryNode, overrides }: ISummaryComponent)
...
         })}
       >
-        <SummaryComponentSwitch
-          summaryNode={summaryNode}
-          targetNode={targetNode}
-          change={change}
-          label={label}
-          overrides={overrides}
-        />
+        {RenderSummary && component instanceof FormComponent ? (
```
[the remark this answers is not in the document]
olemartinorg: Yup, but typescript doesn't agree with that.. 🤷 It complains because it didn't infer that component must be a FormComponent if RenderSummary is set, so the call to `component.renderSummaryBoilerplate()` below fails without that check.

src/components/summary/SummaryComponent.tsx
```diff
@@ -132,13 +137,27 @@ export function SummaryComponent({ summaryNode, overrides }: ISummaryComponent)
...
+          <>
+            {component.renderSummaryBoilerplate() ? (
+              <SummaryBoilerplate
+                {...change}
```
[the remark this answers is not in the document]
olemartinorg: Hah, yeah, not just that - but I messed this up so the feature it provides was broken! 😆 (Meaning, the "change" button disappeared when rendering a summary for a `Group`). Fixed that just before you wrote your review.

src/components/summary/SummaryComponent.tsx
```diff
@@ -132,13 +137,27 @@ export function SummaryComponent({ summaryNode, overrides }: ISummaryComponent)
...
+                label={label}
+                summaryNode={summaryNode}
+                targetNode={targetNode}
+                overrides={overrides}
+              />
+            ) : null}
+            <RenderSummary
+              {...change}
+              summaryNode={summaryNode}
+              targetNode={targetNode as any}
```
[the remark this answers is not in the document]
olemartinorg: Whoops, hoped nobody saw that thing.. 🙄 At this point, typescript thinks both `RenderSummary` and `targetNode` is a union-type of all possible component types, but our typings specify that `RenderSummary` takes in a `targetNode` of its own type. Typescript can't guarantee that here, but it's self-evident from the code that we're not calling the render function for _some component with an entirely _unrelated_ `targetNode`.

In other words, the component utility class and `node` object are not sufficiently bound together. I have some thoughts on how we could do that, but I haven't figured that one out quite yet.

src/layout/Group/index.tsx
olemartinorg: > But again, maybe some other classes that are also inherent in the FormComponent actually use hooks within the useDisplayData method.

Spot on! 😉 "most" components fetch their form data in this step, which many of them get from redux, meaning these methods are usually using hooks to fetch the data they need. This is why I had to disable the hook rules for these files, since the eslint rules thinks these files contain class components (which they don't, but it kinda looks like it). I would agree that it's a bit funky, though, and I think we can move the target a bit further and find a way to work around this while keeping our code somewhat nice.

I added a `getFormData()` function to `LayoutNode`, since formdata is an input when resolving the node tree anyway, so `LayoutNode` can find data in data bindings. That still doesn't work for a `list` binding in, for example, the `FileUpload` component - so that has to figure out the list binding on its own. There's still a way to go here, in other words!

src/components/summary/SummaryComponent.tsx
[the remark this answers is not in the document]
framitdavid: Great! We do not need to fix this within this PR. Hopefully, we could talk about it and figure out how to fix it to avoid this next time. 👍

src/layout/Group/index.tsx
framitdavid: Thanks for the clarification! Anyway, we are in the right direction to make the codebase more maintainable. Great work! 🚀 For now I'm resolving this comment and we can discuss this later on. Maybe we somewhere in the future can fetch data and handle states differently. :)

PR #990, opened 2023-03-10
DevTools: Node inspector (with live data model lookups, schema inspector)

olemartinorg:
## Description
Updated 2023-05-19: This PR now contains a node inspector for DevTools (renaming the layout inspector in the process), which show a read-only view of the node hiearchy.

Some more details and a screenshot in the comment below:
- https://github.com/Altinn/app-frontend-react/pull/990#issuecomment-1554388549

List of features:
- DevTools now have an option to navigate to pages that are hidden in the form
- You can also use a toggle to display all hidden components in the form (either fully, or somewhat greyed out)
- The `Komponenter` tab in DevTools has been renamed `Layout`, and a new `Komponenter` tab has taken its place. Here you can view the resolved state for all components in the form (along with instances of repeated components)
- You can click between the `Layout` and `Komponenter` tabs to switch between editing the properties to simulate layout file changes, and viewing the results in the `Komponenter` tab.
- In the component inspector, you can now view:
  - Most properties on components
  - The form data for each data model binding (updates automatically when changed)
  - The schema information for each data model binding
  - Expressions, with their evalulated results (with click-to-edit in expression tester)
  - Text resource bindings, with both their raw value and lookup result for the current language

List of technical details for these changes:
- The type definition for `state.instanceData.instance?.process` has now been set to undefined, as I encountered a runtime error once where this was undefined.
- The `data-componentid` attribute on DOM elements now use the full node ID, not just the base component id.
- Code to highlight instances of components (and get their refs) have been extracted out into `src/features/devtools/hooks`
- We now have a type definition for the `multiPageIndex` property that is set on multiPage components in repeating groups.
-  The `HierarchyGenerator` classes now gets the responsibility to generate a list of children to return when calling `node.children()`.

<details>
<summary>Show the older PR description, with images</summary>
## Description
This is not a normal pull request.

I've looked into the possibility to create some tooling for figuring out what the data model actually says about the data model bindings for our components. We could use this information to enhance our understanding of the app, and/or possibly create some developer tools in the future.

To summarize how this all works, to make sure we're all on the same page:
- Every component with `dataModelBindings` point to a certain location in the data model, where the data for that component is stored (the data model is basically just a big JSON/XML document).
- Each location is defined with some kind of data type. Previously, our source of truth has been an XSD uploaded to Altinn Studio (which in turn generated a JsonSchema), but the new data modelling tools seem to generate JsonSchema directly.
- It is expected that a `simpleBinding` and such is bound to a `string` in the data model, but we haven't had any way to verify that ahead of time. We use the JsonSchema for automatic validation in app-frontend, so that we can generate error messages for simple things such as "you typed 10 characters here, 5 is the maximum". We're using [`ajv`](https://ajv.js.org/) for this validation.
- While `ajv` is great, it provides us nothing until validation fails. Looking up "how is the data supposed to be stored here" can actually be difficult, because you can have JsonSchemas specifying `anyOf`, which makes for a complex type of any one of many possible types.
- For a `group` binding, it is expected the target is an `array` for repeating groups and `object` for non-repeating groups. We don't validate this currently, and while the JsonSchema could tell us about it, we rely on the `maxCount` property (which may be different in the data model!).

Anyway, with all of this said, the results of using these tools are interesting. I rendered out the JsonSchema for each data model binding above each component in our test app:

changename-with-data-model

group-with-data-model
</details>

## Related Issue(s)

- #136
- #1021

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
  - [x] No changes made
- Support in Altinn Studio
  
- Sprint board
- Labels


ivarne:
Another “source of truth” is the C# classes that are used in the backend for (de)serializing json and xml, writing server side code (and crashing everything if a numerical input is too big, or a datetime is invalid).

olemartinorg:
Updated this PR now to make it possible to merge against main. I moved this tool into DevTools, so now it's possible to inspect the node hierarchy (where live results from expressions are displayed), along with resolved text resource bindings and data model bindings (with details from the data model schema, and the current data from the data model).

Relevant screenshot of the inspector:
20230519_12h39m37s_grim

Sadly, my backend developer background is showing - I tried, but feel like I failed to make this look good. 🤷 It's got loads of data in it, though. Feel free to make changes. Opening this up for review now.

olemartinorg:
> The node tree and layout list looks a bit cluttered when using UUIDs for IDs. Not sure how to mitigate this though.
Whoops.. :-/ I changed this from the layout inspector so that the ID floats up directly to the right of the node type, because the `display: table-cell` stuff didn't work all that well with a hierarchy. I might get it to work better with `display: grid` though - I'll look into it.

> I am getting `null` for all of my datamodel-bindings:
Hmm! Strange. Which app is this? And does it have a Json schema for the data model? Not all apps have that, but when using the new data modelling tools it _should_ work.

> Clicking child components causes a "refresh", not a huge problem, but resets any local changes made in the layout inspector
Ahh, yup - I cheated and used a `<a href="#" />` there, so that must be the reason.

> Why is `number` clickable? Is it just to collapse the object? This also causes a "refresh" which seems a bit buggy.
Yep, just to collapse it. Let me know if you have better suggestions for how to make it communicate as collapsible!

> Would be cool to see the expression that resolves to the value
Yesss! I thought the same. "Click to examine in expression tester" or something like that. I'm hoping to make the expression engine work on top of the layout hierarchy soon, such that changes that just affects the expressions inside components will re-run the expressions instead of re-creating the entire hierarchy. For that, I'd need to store a un-evaluated version of the node configuration alongside the result, which would make this easier to implement. But yeah, I think I can make this work with a few shortcuts! 🤔

bjosttveit:
> > I am getting `null` for all of my datamodel-bindings:
> 
> Hmm! Strange. Which app is this? And does it have a Json schema for the data model? Not all apps have that, but when using the new data modelling tools it _should_ work.

It is my [sogndal app](https://altinn.studio/repos/bjosttveit/newcomer-sogndal/src/branch/master/App/models/datamodel.schema.json). It does have a json schema, but it was created back in ye olden days of september (v6), using the example xsd with some manual modifications.
 
> > Why is `number` clickable? Is it just to collapse the object? This also causes a "refresh" which seems a bit buggy.
> 
> Yep, just to collapse it. Let me know if you have better suggestions for how to make it communicate as collapsible!

That is fine, I was just confused that clicking it caused the page to "refresh".

> > Would be cool to see the expression that resolves to the value
> 
> Yesss! I thought the same. "Click to examine in expression tester" or something like that. I'm hoping to make the expression engine work on top of the layout hierarchy soon, such that changes that just affects the expressions inside components will re-run the expressions instead of re-creating the entire hierarchy. For that, I'd need to store a un-evaluated version of the node configuration alongside the result, which would make this easier to implement. But yeah, I think I can make this work with a few shortcuts! 🤔

🥳 🥳

bjosttveit (approved):
Very nice! since this is devtools, i will just approve it now and improvements can come later.

olemartinorg:
@bjosttveit I agree, the devtools-specific changes can probably be skipped in review. However, I ended up making a lot of other changes in the process of implementing these features. I'd much appreciate a review of those! 🙏 

Btw, I think I fixed the issues you mentioned above (although not always beautifully). Take another look if you get a chance!

bjosttveit:
# 🤯 🤯 

image
image

PR #1002, opened 2023-03-14
Fix layout sets schema

bjosttveit:
## Description

The definitions were incorrectly positioned inside of properties, and the property `type` should be `dataType`. I have tested these changes locally.


olemartinorg (approved):
Looks good. I checked, and no current apps I could find are setting `"type"` in their `layout-sets.json`.

PR #1006, opened 2023-03-14
Selected Language stored in localstorage should be saved per user

### LINKED ISSUE #1005: Selected Language stored in localstorage should be saved per user.
### Description of the bug

Language preference should be stored per user and should not be shared.

### Bakground
When multiple users are taking a survey on a public computer it is highly likely users have different language preferences. The language one user picks should not change the default for another user.

### Steps To Reproduce

1. Log in as user 1
2. Change language
3. Log inn as a different user
4. The selected language should be default but is the same as the first user language

### Additional Information

I suggest changing the localStorage key to include a unique key for the user

lasseklovstad:
## Description
selectedAppLanguage state is moved to profileState in Redux store (It made more sense to me). When the selectedAppLanguage is changed the language is saved to localStorage as before. BUT the localStorage key include the userId (New). 

I believe this is a breaking change for the users who has changed their language (For them it will default back to profile preferences). 

## Related Issue(s)

- closes #1005

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
  - [x] No changes made
- Support in Altinn Studio
  
- Sprint board
- Labels

### DIFF
--- a/src/shared/resources/profile/profileSlice.ts
+++ b/src/shared/resources/profile/profileSlice.ts
@@ -1,3 +1,6 @@
+import { put } from 'redux-saga/effects';
+
+import { OptionsActions } from 'src/shared/resources/options/optionsSlice';
 import { fetchProfileSaga } from 'src/shared/resources/profile/fetch/fetchProfileSagas';
 import { createSagaSlice } from 'src/shared/resources/utils/sagaSlice';
 import type {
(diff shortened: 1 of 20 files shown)


src/shared/resources/profile/profileSlice.ts
```diff
@@ -7,14 +10,23 @@ import type {
...
+const altinnWindow = window as Window as IAltinnWindow;
+const getLanguageStorageKey = (userId: number | undefined) => `selectedAppLanguage${altinnWindow.app}${userId ?? ''}`;
+
+export const initialState: IProfileState = {
   profile: {
     profileSettingPreference: {
       language: 'nb',
     },
   } as IProfile,
+  selectedAppLanguage: '',
```
lasseklovstad, line 29: Maybe the default language should be:
```suggestion
  selectedAppLanguage: getLanguageStorageKey(undefined),
```
This will then support remembering the selected language for stateless app (no user logg in). But is it neccassary to store the selected language for a stateless? Maybe it is better to store this in sessionStorage...

framitdavid: First of all, thanks for your great contribution.  🚀 I agree with you that we should use "`getLanguageStorageKey(undefined)`" to persist the selected language on stateless apps. It's more of a nice-to-have feature than a need-to-have for stateless apps if you ask me.

Since the user is not logged in and we do not know if the next user is the same person, it might make sense to persist this within session storage instead of local storage to ensure other persons/users get the default language.

Within this PR I think it's good enough **not** persist the language within stateless apps. Because if we set the key `selectedAppLanguage${altinnWindow.app}` and later on move this to session storage, we need to write logic to clean up local storage by removing the key.  :)

lasseklovstad: I agree. It makes sense not to save it to localStorage if no user is logged inn.

framitdavid (approved):
🚀 🥇

PR #1027, opened 2023-03-22
Added test case for custom receipt

bjosttveit:
## Description

Added a test case for custom receipt to prevent it from unknowingly breaking again.

## Related Issue(s)

- https://github.com/Altinn/app-frontend-react/pull/1024


olemartinorg (approved):
🚀

PR #1047, opened 2023-03-24
Check if current layout is visible after fetchsettings

### LINKED ISSUE #1044: Conditionally showing pages in a step using expressions seems to be broken
### Description of the bug

This case is relevant to signing, where you would want to use permissions to determine which page to show for a given step. The easiest way to do this is to use hidden expressions for the pages to show either one or the other.

There seems to be a problem with this currently however, where it will not be able to show the page unless it is first in the page order. I would like this to be possible without having to ask the backend for a new page order. It seems that the `currentView` is not updated even though that page has been added to the list of hidden pages.

image

### Steps To Reproduce

1. Create two pages for a step
2. Add a hidden expression for each such that only one page will be hidden at a time
3. Test both cases of the expression and observe that only the page first in the order is shown

### Additional Information

_No response_

bjosttveit:
## Description

The source of the problem was that after settings are fetched, it would set the current page to what was stored in localStorage, and if that was not set; the first page in the order. It does this without checking if the page was hidden, leading to the issue. The fix simply calls `findAndMoveToNextVisibleLayout` after each time settings are fetched. This saga has no effect if the current page is visible, but will move to the next visible page if it is hidden.

I have not added a specific test case for this, but the signing-tests will cover this functionality.

## Related Issue(s)

- closes #1044 

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
  - [x] No changes made
- Support in Altinn Studio
  
- Sprint board
- Labels

### DIFF
--- a/src/features/form/layout/formLayoutSlice.ts
+++ b/src/features/form/layout/formLayoutSlice.ts
@@ -118,6 +118,7 @@ export const formLayoutSlice = createSagaSlice((mkAction: MkActionType<ILayoutSt
       saga: () => watchFetchFormLayoutSettingsSaga,
     }),
     fetchSettingsFulfilled: mkAction<LayoutTypes.IFetchLayoutSettingsFulfilled>({
+      takeEvery: findAndMoveToNextVisibleLayout,
       reducer: (state, action) => {
         const { settings } = action.payload;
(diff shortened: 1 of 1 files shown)


olemartinorg (approved):
One line of greatness is always worth far more than a thousand lines of spaghetti! 🚀

PR #1061, opened 2023-03-29
Fix group panel schema

bjosttveit:
## Description

Quick fix to remove incorrect warnings on group panel configuration:

image

### DIFF
--- a/schemas/json/layout/layout.schema.v1.json
+++ b/schemas/json/layout/layout.schema.v1.json
@@ -556,7 +556,6 @@
       }
     },
     "groupPanelOptions": {
-      "additionalProperties": false,
       "allOf": [{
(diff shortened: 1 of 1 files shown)


schemas/json/layout/layout.schema.v1.json
```diff
@@ -556,7 +556,6 @@
       }
     },
     "groupPanelOptions": {
-      "additionalProperties": false,
```
olemartinorg, line 559: Removing this line just makes it possible to type these in without having any schema validation. The root cause of this must be something else, because these properties are defined in `#/definitions/panelComponent` that is referenced in the `allOf` below..

olemartinorg (approved):

PR #1063, opened 2023-03-29
Fixing flaky cypress tests, stale data use when clicking checkboxes/radios very quickly

olemartinorg:
## Description
- Removes most usages of `{ force: true }`, only keeping them for file uploaders. This flag was used to trick Cypress into checking/unchecking hidden inputs (radio/checkboxes), but we should make tools for these things instead, to make our tests behave like our users would. Implemented `dsCheck()` and `dsUncheck()` for this purpose.
- Fixing a bug in `useDelayedSaveState()` where a slow re-render by React using a stale `formValue` would overwrite data after a user had made a newer change. Fixing this makes it possible for us to avoid the workarounds we had for this in our tests.
- Upgrading Cypress to `12.9.0` while I was at it
- Removed lots of `.should('exist').and('be.visible').click()` logic. We don't have to check if an element exists and is visible before trying to click it, because Cypress will retry the click until the element is visible -- or fail trying. Starting off with these verbose 'should be visible' assertions actually makes Cypress more flaky, as it can't retry finding the element to click if an older one (that was visible) became invisible during the chained run. Removing these makes for leaner test code, and makes the tests more stable. Read more here: https://docs.cypress.io/guides/core-concepts/retry-ability

## Related Issue(s)

- closes #{issue number}

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
  - [x] No changes made
- Support in Altinn Studio
  
- Sprint board
- Labels

### DIFF
--- a/package.json
+++ b/package.json
@@ -127,7 +127,7 @@
     "@altinn/altinn-design-system": "0.27.6",
     "@babel/polyfill": "7.12.1",
     "@date-io/moment": "1.3.13",
-    "@digdir/design-system-react": "0.9.0",
+    "@digdir/design-system-react": "0.10.2",
     "@material-ui/core": "4.12.4",
     "@material-ui/icons": "4.11.3",
     "@material-ui/pickers": "3.3.10",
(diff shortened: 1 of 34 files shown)


bjosttveit (approved):
# 🧙‍♂️

PR #1071, opened 2023-03-31
added showAddButton property

lassopicasso:
## Related Issue(s)

- closes #{issue number}

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
  - [ ] No changes made
- Support in Altinn Studio
  
- Sprint board
- Labels

### DIFF
--- a/src/layout/Group/types.d.ts
+++ b/src/layout/Group/types.d.ts
@@ -17,6 +17,7 @@ export interface IGroupEditProperties {
   openByDefault?: boolean | 'first' | 'last';
   alertOnDelete?: ExprVal.Boolean;
   saveAndNextButton?: ExprVal.Boolean;
+  showAddButton?: ExprVal.Boolean;
 }
 
 export interface ILayoutGroup extends ILayoutCompBase<'Group'> {
--- a/schemas/json/layout/layout.schema.v1.json
+++ b/schemas/json/layout/layout.schema.v1.json
@@ -537,6 +537,12 @@
           "description": "Boolean or expression indicating whether add new button should be shown or not under the table.",
           "$ref": "expression.schema.v1.json#/definitions/boolean"
         },
+        "showAddButton": {
+          "title": "Show add button on open group",
+          "description": "Boolean value indicating whether add new button should be shown or not under the table when a group is open.",
+          "type": "boolean",
+          "default": false
+        },
         "openByDefault": {
           "title": "Open by default",
           "description": "Boolean or string indicating if group should be opened by default. If no items exist: 'first', 'last', and true adds a new item. If items exist already, true does not open anything, but 'first' opens the first item, and 'last' opens the last item in the group.",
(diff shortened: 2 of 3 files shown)


src/layout/Group/types.d.ts
```diff
@@ -17,6 +17,7 @@ export interface IGroupEditProperties {
   openByDefault?: boolean | 'first' | 'last';
   alertOnDelete?: ExprVal.Boolean;
   saveAndNextButton?: ExprVal.Boolean;
+  showAddButton?: ExprVal.Boolean;
```
olemartinorg, line 20: Be sure to call it `alwaysShowAddButton` in the implementation as well as the documentation! 😁

src/layout/Group/types.d.ts
```diff
@@ -17,6 +17,7 @@ export interface IGroupEditProperties {
   openByDefault?: boolean | 'first' | 'last';
   alertOnDelete?: ExprVal.Boolean;
   saveAndNextButton?: ExprVal.Boolean;
+  alwaysShowAddButton?: ExprVal.Boolean;
```
olemartinorg, line 20: With the type `ExprVal.Boolean` you indicate that it's possible to use a dynamic expression here. I'm not sure if that's requested (or even needed) here, and you haven't implemented support for it in the expressions language by adding a configuration for it in `ExprConfigForGroup` (also, as far as I could see, it wasn't documented in the list of places expressions could be used).

In other words, I think the type here should be just `boolean`.

src/layout/Group/GroupContainer.tsx
```diff
@@ -166,6 +164,13 @@ export function GroupContainer({ id }: IGroupProps): JSX.Element | null {
     );
   }
 
+  const displayBtn = () =>
+    edit?.mode !== 'showAll' &&
+    edit?.addButton !== false &&
+    'maxCount' in node.item &&
+    repeatingGroupIndex + 1 < (node.item.maxCount === undefined ? -99 : node.item.maxCount) &&
+    (editIndex < 0 || edit?.alwaysShowAddButton === true);
```
olemartinorg, line 172: This doesn't have to be a function, you can just assign the value to the `displayBtn` constant instead.

The check for `edit?.mode !== 'showAll'` is just because we render `<AddButton>` in two different places (see the other code that renders AddButton below here). It's a bit misleading to say that we don't display the add button when the mode is `showAll`, because we absolutely do. Maybe you could re-use the `displayBtn` boolean there as well, so that we get this functionality when using `edit.mode === 'showAll'` (as an app developer I would expect the functionality to work the same regardless of edit mode).

olemartinorg (changes requested):

lassopicasso:
@olemartinorg, I encountered a bug with this issue that was mentioned in today's daily meeting. See the note in description. Was talking with David about this and came to a conclusion to create a new issue on this. However, after given this some thought, the disabled add-button when validating can make this issue bigger and would like to hear what you think about this.

olemartinorg (approved):
I think it looks good! I agree that we should open a new issue for the bug you found. Usually these things are released without everyone starting using them all at the same time, so it might be a while until others find the same bug and it starts being a big problem.

StianVestli:
Testet dette nå i skjema med gruppe i gruppe og det ser ut som det funker som det skal, buggen som er nevnt har jeg ikke snublet i enda.

PR #1072, opened 2023-03-31
WCAG: Update Dropdown and MultipleSelect to use @digdir/designsystem

### LINKED ISSUE #510: Replace MultipleSelect with component from design system
### Description

In https://github.com/Altinn/app-frontend-react/pull/505 we implemented a multiple select component in app-frontend. This component was implemented directly as a result of time pressure. Once this component is [implemented in design system](https://github.com/Altinn/altinn-design-system/issues/138), this implementation should be replaced.

### Additional Information

_No response_

### Tasks

- [ ] Replace existing implementation with new
- [ ] The current multiple select only supports storing data as a comma-separated string. A useful feature here would be to store this as an array of selected values, reusing the likert-component implementation with "list" binding. Consider adding this.
- [ ] Update user docs if relevant

### Acceptance Criterias

- Custom implementation is replaced with component from design system

bjosttveit:
## Description

Updated the Dropdown and MultipleSelect component to use the Select component from the designsystem. Removed the dependency of `react-select` as this is no longer used anywhere. There is a known issue with the select since the last design-system version https://github.com/digdir/designsystem/issues/314. However this only has an effect when running in webpack-devserver and not when built and served statically.

Also added support for description for options using the formattedLabel option of the new select.

## Related Issue(s)

- closes #510 

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
  - [x] No changes made
- Support in Altinn Studio
  
- Sprint board
- Labels


test/e2e/support/custom.ts
```diff
@@ -17,6 +17,13 @@ Cypress.Commands.add('dsUncheck', { prevSubject: true }, (subject: JQueryWithSel
   }
 });
 
+Cypress.Commands.add('dsSelect', { prevSubject: true }, (subject: JQueryWithSelector | undefined, name) => {
+  cy.wrap(subject).click();
+  cy.wrap(subject).parents('[data-testid="select-root"]').findByRole('option', { name }).click();
+  cy.get('body').click();
+  cy.wrap(subject);
```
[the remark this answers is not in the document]
bjosttveit: The body click is to call blur (due to delayed save state) and the last wrap is to make it chainable again, otherwise it would be the body that would get the chained operations

bjosttveit:
Multitasking 💪 😅 

https://user-images.githubusercontent.com/47412359/231437331-96b32b0b-e790-4fdb-95e4-08cec2d1c1c1.mov

olemartinorg (approved):
🙏

test/e2e/support/index.ts
```diff
@@ -49,3 +49,11 @@ afterEach(function () {
     });
   }
 });
+
+const resizeObserverLoopErrRe = /^[^(ResizeObserver loop limit exceeded)]/;
```
olemartinorg, line 53: CodeQL complains a lot here, and the regex looks a bit odd. Maybe this could work instead?

```suggestion
const resizeObserverLoopErrRe = /^ResizeObserver loop limit exceeded/;
```

And, we should probably have a proper description in a comment here on why we need to ignore the error (it's not just for forcing the test to pass). It should probably also be noted that it should be removed if possible, when the design system (hopefully) fixes the root cause.

olemartinorg: Otherwise, if the `[` and `]` were supposed to be literal, they should be escaped in the regex.

bjosttveit: this is straight up copied, and I don't know why everyone uses a regex like that

bjosttveit: Turns out someone else pointed that out too https://github.com/cypress-io/cypress/issues/8418#issuecomment-992564877

olemartinorg: Ah, nice! That's an even better solution.

Issue #1021, opened 2023-03-21
Developer tools


olemartinorg:
### Description

We're creating a platform used by developers to create apps. Let's give them tools.

I've had this marinating in the back of my head for a while, so when the opportunity arose to write a list of features I'd want on Slack, it's probably best to detail this in an issue as well.

### Some nice-to have features
- Visualizing the data model while filling out the form. There could be multiple ways to show the data model.
  - Resolved bindings for each component, aka `node.getFormData()` (although that does not currently work well with a `list` binding)
  - The entire data model in dot-notation (aka the redux state)
  - The entire data model as JSON (aka what we send off to save on the backend)
  - The entire data model as XML (we could probably ask the backend for the raw data)
  - See #136
- Functionality to inspect properties on all components
  - See expressions and what they resolve to, for each component in the form
  - Bonus points if we can visualize every evaluation step in an expression (in `["equals", ["component", "something"], "yes"]` we evaluate two discrete expression functions, first `["component", "something"]` and `["equals", "value", "yes"]`)
- Inspect the data model targets for every component binding
  - I've made a prototype for this, see #990
- Showing/hiding hidden components, and why they are hidden
- Showing/hiding hidden pages, and why they are hidden
- Showing errors/warnings (from expression evaluation and from validations as defined in #648)
- Navigating the current form pages without having to add a navigation bar while developing
- One button-click to preview PDF
- Show indicators to see which components trigger validations (and what will be validated when) in the GUI
- Adding functionality to load other versions of `app-frontend-react` (it should be possible to implement with some trickery). That means we could easily switch the version in tt02 and production as well.
- Keyboard shortcut to enable/display the developer tools, and alternatively another way to open them on mobile.
- Show the developer tools (or just a button for it) automatically when the domain is `local.altinn.studio`. That way we don't even have to advertise that we're shipping developer tools to app developers; they'll find out. Just make sure to inform them about the keyboard shortcut in production as well.

bjosttveit:
Could it be an idea to add a "console"? The "real console" gets a bit cluttered with random warnings and such. For example, expressions log pretty errors to the console if they fail to resolve, maybe we could have a cleaner console in the developer tools with only intentional logs?

bjosttveit:
Yeah, maybe just a log, we dont need to be able to send commands really. Maybe hard to make it display things correctly?

PR #1076, opened 2023-04-03
Developer Tools MVP

### LINKED ISSUE #1075: Keyboard shortcut to enable/display the developer tools, and alternatively another way to open them on mobile.
Using this as a starting point for developer tools. I thought starting with a keyboard shortcut is best so that people are not too confused by a new button without much functionality.

bjosttveit:
## Description

This is the starting point for developer tools. This PR adds the developer tools panel and a few simple features to begin with.

**[Related epic](https://github.com/Altinn/app-frontend-react/issues/1021)**

- It can be opened with the hotkey `Ctrl+Shift+K`/`⌘+Shift+K`.
- There is a button hiding away in the bottom right corner that will also open it. (yes I tested it on mobile)
  - This is only visible in `local.altinn.cloud`.
- The panel is resizable (also works on mobile with touch events 😩 )
- **Functionality**:
  - PDF preview button
  - Navigate pages that are in the current page order.

Video of the current capabilities:

https://user-images.githubusercontent.com/47412359/229771638-60ab60aa-02fa-49bf-b76d-1f9f17da3739.mov

## Related Issue(s)

- closes #1075 

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
  - [x] No changes made
- Support in Altinn Studio
  
  - [?] This change/feature does not require any changes to Altinn Studio
- Sprint board
- Labels


olemartinorg (approved):
Great stuff! 👏 🚀 Maybe we could add a message here informing newcomers about what this is, and how to open/close it via the keyboard for tt02/production apps (so they'll get a hint on `local.altinn.cloud`, but knows that the feature is available in other environments as well).

Going forwards, I think the firefox/chrome devtools layout is nice, and something we could copy. By that, I mean less padding, tabs to navigate between different devtools features, etc.

src/features/devtools/DevTools.tsx
```diff
@@ -0,0 +1,44 @@
...
+  children: ReactNode;
+}
+
+export const DevTools = ({ children }: IDevToolsProps) => {
+  const [panelOpen, setPanelOpen] = useState(false);
+
+  useEffect(() => {
+    if (devToolsEnabled) {
+      return window.addEventListener('keydown', (event) => {
+        if ((event.ctrlKey || event.metaKey) && event.key === 'k') {
```
ivarne, line 19: Isn't this shortcut rather short and likely for normal users to press inadvertently.

bjosttveit:
Just hide the button in tt02/prod? So you think it should be available in prod as well?

How do you think I should give the information about the availability? a pop-up message?

Totally agree on the layout stuff, I had to stop myself from going down that rabbit hole now since there are only two features currently 😅

ivarne (commented):
This seems like a nice place to start using code splitting. If devtools grows to provide half of the suggested features, it would be nice if everyone didn't need to download functionality they won't use.

src/features/devtools/DevTools.tsx
bjosttveit: I will gladly take suggestions for other shortcuts, maybe just add `Shift`?

ivarne: I think i missunderstood this. It's currently just a shortcut for when the icon is already visible?

I'm thinking about when this is activated in prod, the mode should probably be activated by something else than a key combination that others might have configured for other purposes (or inadvertently type). Personally I'd suggest setting a custom cookie (devtoolsEnable:true) from javascript when the user adds `?devtools=` in their url. The devtools can provide a button to delete the cookie, and docs entries can link to any app to help users set the cookie. 

PS: I don't think code splitting should be dependent on `Index.cshtml`, but I have not tried setting it up myself. I just used Create React App.

bjosttveit:
A message like this? @olemartinorg 

image

src/features/devtools/DevTools.tsx
olemartinorg: I believe we'll actually want to have the devtools somewhat easily accessible in prod as well. The goal here is not to hide anything, even from competent users (they could figure all of this out without devtools if they wanted), and using a cookie or custom URL just adds more friction than we really need. My goal is to make these tools available on mobile as well some day, and making mobile users inject a cookie will make it very hard for them to use the tools. (Keyboard shortcuts isn't all too easy on mobile either, but we'll figure something out in time.)

PR #1086, opened 2023-04-11
Devtools: Basic component inspector

bjosttveit:
## Description

This adds a "component inspector" that is currently just a list of your components from the current layout file. Note, this is not a hierarchy inspector, that could be a separate thing later. This inspector should be extended to also show the component properties when clicking a component from the list. This is relatively easy to implement, but we should consider adding some sort of panel resizing component to make this work well. This also adds tabs to the devtools panels, since it would get very cluttered to include this on the same page.

### DIFF
--- a/src/features/devtools/components/LayoutInspector/LayoutInspector.tsx
+++ b/src/features/devtools/components/LayoutInspector/LayoutInspector.tsx
@@ -1,0 +1,25 @@
+import React from 'react';
+
+import classes from 'src/features/devtools/components/LayoutInspector/LayoutInspector.module.css';
+import { LayoutInspectorItem } from 'src/features/devtools/components/LayoutInspector/LayoutInspectorItem';
+import { useAppSelector } from 'src/hooks/useAppSelector';
+
+export const LayoutInspector = () => {
+  const { currentView } = useAppSelector((state) => state.formLayout.uiConfig);
+  const layouts = useAppSelector((state) => state.formLayout.layouts);
+
+  const currentLayout = layouts?.[currentView];
+
+  return (
+    <div className={classes.container}>
+      <ul>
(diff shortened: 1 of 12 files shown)


olemartinorg (approved):
🚀

src/features/devtools/components/LayoutInspector/LayoutInspector.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+import classes from 'src/features/devtools/components/LayoutInspector/LayoutInspector.module.css';
+import { LayoutInspectorItem } from 'src/features/devtools/components/LayoutInspector/LayoutInspectorItem';
+import { useAppSelector } from 'src/hooks/useAppSelector';
+
+export const LayoutInspector = () => {
+  const { currentView } = useAppSelector((state) => state.formLayout.uiConfig);
+  const layouts = useAppSelector((state) => state.formLayout.layouts);
+
+  const currentLayout = layouts?.[currentView];
```
olemartinorg, line 11: Btw, getting the entire layout hierarchy is as easy as calling `useExprContext()` instead. Using the flat layout in the developer tools dooms us into re-creating the hierarchy in the devtools, so we should continue using the hierarchy instead of the flat layout (and showing the computed properties are much more useful). The `LayoutNode` objects also have methods for getting form data (aka the data model) and validation messages for each component.

bjosttveit: The thought behind this inspector is more to look at the component declaration and possibly make modifications there (similar to modifying css properties in the chrome dev tools, where computed properties are a separate thing). Then you could play with expressions in the devtools and see the changes happening in real time 😃 

So in my opinion both the raw layout components and computed hierarchy are very useful

PR #1092, opened 2023-04-13
Some cleanup of material ui

lassopicasso:
## Description
Replaced some material-ui code with css-modules. Mainly on summary page.

I've checked that the changes haven't affected the styling, but there is a chance that I've overlooked something. Would appreciate it if someone else could check this out as well.

Components used on following pages in the frontend-test app:
### Summary Page
- ErrorPaper.tsx for groups.
  - E.g. force an error and see the error card in summary.
- SummaryGroupComponent.tsx
  - The elements when summaryNode.item.largeGroup is false. 
- SummaryBoilerPlate.tsx
  - Label and edit button in summary component when renderSummaryBoilerPlate is found in the component.
 - SummaryItemSimple.tsx
  - E.g. The text below SummaryBoilerPlate.tsx in a summary component. 
- SummaryComponent.tsx
  - Only replaced useStyles from Material-ui. 
  The Grid-tag with dynamic breakpoints makes this a bit complex whereas user can set the width (number of columns) of each component for different breakpoints. Not sure what the best way is to replace it with css-modules without making it a big bunch of code.
- SummaryItemCompact.tsx
  - Rendered when largeGroup is set to false and there are rows in group.

### Other
- EditIconButton.tsx in PanelGroupContainer that is used on repeating page. 
  - Edit button in a active row in reference group.
- AltinnError.tsx in InstantiatonErrorPage
  - E.g. when running a new instance fails.

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
  - [x] No changes made
- Support in Altinn Studio
  
- Sprint board
- Labels


src/layout/Group/SummaryGroupComponent.module.css
```diff
@@ -0,0 +1,44 @@
...
+.link {
+  background: none;
+  border: none;
+  border-bottom: 2px solid #008fd6;
+  cursor: pointer;
+  padding-left: 0px;
+}
+
+.emptyField {
+  font-style: 'italic';
```
bjosttveit, line 40: Having it inside quotes is actually invalid.

```suggestion
  font-style: italic;
```

src/layout/Summary/SummaryBoilerplate.module.css
```diff
@@ -0,0 +1,13 @@
+.container {
+  display: flex;
+  flex-direction: row;
+  width: 100%;
+  justify-content: space-between;
+}
+.label {
+  font-weight: 500;
+}
```
bjosttveit, line 9: Is the label supposed to look that thicc? It used to have a weight of 400:

image
image

bjosttveit: There is a lot of **bold** text now

image

src/layout/Summary/SummaryBoilerplate.tsx
```diff
@@ -57,16 +33,14 @@ export function SummaryBoilerplate({
   return (
     <>
```
framitdavid, line 34: This fragment is not necessary, because the next div is wrapping everything. :)

src/layout/Summary/SummaryComponent.module.css
```diff
@@ -0,0 +1,13 @@
...
+  border-bottom: 1px dashed var(--file-list-border-color);
+}
+
+.link {
+  background: none;
+  border: none;
+  border-bottom: 2px solid var(--file-list-border-color);
+  cursor: pointer;
+  padding-left: 0px;
+}
```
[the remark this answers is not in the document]
framitdavid: I agree, but should be speak with the UX to find out if this is more a generic thing? In my head, border-bottom on links should be the same within the whole application. We should avoid to create many css-variables with the same values, if we do not think they are going to be different.

Example for the dashed border I would prefeer to create a css variable that look something like this:

`--dashed-border-primary: 1px dashed blue;`

Then we can use this variable where we need dashed border and it stays consitent.

```
.border {
  border-bottom: var(--dashed-border-primary)
}
```

By doing something like I metioned above, we do not need to think about the colors nor the size of the border. Everything is already set to match the design.

What do you think this approch?

src/components/message/ErrorPaper.module.css
```diff
@@ -0,0 +1,19 @@
...
+  box-shadow: 0px 2px 1px -1px rgba(0, 0, 0, 0.2), 0px 1px 1px 0px rgba(0, 0, 0, 0.14),
+    0px 1px 3px 0px rgba(0, 0, 0, 0.12);
+}
+
+.paper {
+  display: flex;
+  justify-content: flex-start;
+  padding: 12px;
+  align-items: center;
+  gap: 7px;
```
framitdavid, line 14: should we add flex-wrap to this, to ensure that it wraps if the screen size is to small to render the icon and the text side by side?

framitdavid (changes requested):

src/layout/Summary/SummaryComponent.module.css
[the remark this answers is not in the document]
framitdavid: I agree, but should be speak with the UX to find out if this is more of a generic thing? In my head, the border-bottom on links should be the same within the whole application. We should avoid creating many css-variables with the same values if we do not think they are going to be different.

For example, for the dashed border, I would prefer to create a CSS variable that looks something like this:

`--dashed-border-primary: 1px dashed blue;`

Then we can use this variable where we need a dashed border and it stays consistent.

```
.border {
  border-bottom: var(--dashed-border-primary)
}
```
By doing something as I mentioned above, we do not need to think about the colours or the size of the border. Everything is already set to match the design.

What do you think about this approach?

src/layout/Group/SummaryGroupComponent.module.css
lassopicasso: True that :sweat_smile:

src/layout/Summary/SummaryComponent.module.css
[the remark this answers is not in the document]
bjosttveit: Yeah that's even better 👍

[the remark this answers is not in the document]
lassopicasso: Also including css-variables for dotted and solid. 
image

src/layout/Summary/SummaryBoilerplate.module.css
lassopicasso: Hm yeah, that didn't look that good. Will fix it, thx!

src/components/altinnError.module.css
```diff
@@ -0,0 +1,36 @@
...
+  font-size: 0.875rem;
+}
+.articleText {
+  font-size: 1.125rem;
+}
+.title {
+  font-weight: 500;
+  color: #022f51;
+}
+
```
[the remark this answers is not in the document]
lassopicasso: I believe so that all h1 uses font-weight 500 when looking at tjenester3.css, but the color is just plain black when I look at H1 in header.

src/layout/Summary/SummaryComponent.module.css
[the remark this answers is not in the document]
framitdavid: @lassopicasso great work with the CSS variables, I think this is a step in the right direction. 👏

bjosttveit (approved):
Very good 🥇

framitdavid (approved):
LGTM! 👏 🥇

Issue #260, opened 2020-08-07
As an appdeveloper I want to give the user a description beneath a specific radio/checkbox label 


Febakke:

xmrsa:
I got this layout by using a <br>-tag in the text definition. The size of the text is of course the same... were you perhaps thinking smaller text for the description under the main text of the Svaralternativ @Febakke ?

In connection with this I discovered a mistake concerning the alignment of such a text (that was for checkboxes), I tested this on radiobuttons now, and the same thing happens there, it is center-aligned and looks very strange.
![RadiobuttonAlignproblem](https://user-images.githubusercontent.com/42466346/183595895-a336b33c-ae43-400f-92cb-034f92d80910.jpg)
 I will add this problem to the bug I just created, https://github.com/Altinn/app-frontend-react/issues/369.

mrosvik:
Denne er vel egentlig utført nå. I alle fall i designsystemet. https://designsystem.altinn.studio/?path=/story/components-radiobutton--default Er det mulig å legge på description i altinn studio @TomasEng  ?

ivarne:
Mangler både i altinn-app-frontend og i app-lib-dotnet (om man bruker `optionId`), så bør vel vente med å aktivere det i studio til det faktisk virker i apper.

TomasEng:
Per nå er det ikke mulig å legge til en slik beskrivelse i Studio.

PR #1101, opened 2023-04-17
Enable description for checkboxes and radiobuttons

Magnusrm:
## Description
Enables the use of descriptions for options in CheckboxGroups and RadioGroups

## Related Issue(s)
- https://github.com/Altinn/app-frontend-react/issues/260

Backend PR: https://github.com/Altinn/app-lib-dotnet/pull/230
## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
  - [ ] No changes made
- Support in Altinn Studio
  
- Sprint board
- Labels

### DIFF
--- a/schemas/json/layout/layout.schema.v1.json
+++ b/schemas/json/layout/layout.schema.v1.json
@@ -662,6 +662,16 @@
           "type": "string",
           "title": "Value",
           "description": "The option value."
+        },
+        "description": {
+          "type": "string",
+          "title": "Description",
+          "description": "A description of the option displayed in Radio- and Checkbox groups. Can be plain text or a text resource binding."
+        },
+        "helpText": {
+          "type": "string",
+          "title": "Help Text",
+          "description": "A help text for the option displayed in Radio- and Checkbox groups. Can be plain text or a text resource binding."
         }
       },
       "required": ["label", "value"]
--- a/src/layout/Checkboxes/CheckboxesContainerComponent.tsx
+++ b/src/layout/Checkboxes/CheckboxesContainerComponent.tsx
@@ -133,7 +133,7 @@ export const CheckboxContainerComponent = ({
           checked: selected.includes(option.value),
           label: getTextResource(option.label),
           description: getTextResource(option.description),
-          helpText: option.helpText && getTextResource(option.helpText),
+          helpText: getTextResource(option.helpText),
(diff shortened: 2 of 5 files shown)


schemas/json/layout/layout.schema.v1.json
```diff
@@ -662,6 +662,16 @@
...
+        },
+        "description": {
+          "type": "string",
+          "title": "Description",
+          "description": "A description of the option displayed in Radio- and Checkbox groups. Can be plain text or a text resource binding."
+        },
+        "helpText": {
+          "type": "string",
+          "title": "Help Text",
+          "description": "A help text for the option displayed in Radio- and Checkbox groups. Can be plain text or a text resource binding."
```
bjosttveit, line 674: Does this also apply to options for multiselect and dropdown? I suppose this also applies to the backend in the AppOption class. It is not necessarily a problem as I assume it will simply have no effect for these components?

olemartinorg (approved):

bjosttveit (approved):
Very good!

schemas/json/layout/layout.schema.v1.json
Magnusrm: True, it does apply to options for multiselect and dropdown. And correct, it has no effect.

bjosttveit: Its probably more of a hassle for users to have to deal with different option classes anyway, and it would probably break backwards compatibility to change the classes in the backend, so I think we should just keep it as is. If people are confused by the layout schema it could be changed later.

ivarne: Or we should just add support for dropdown (and multiselect).

Finanstilsynet has a case where we want to display some information about the choice you made in a dropdown, and this would be a really nice improvement over our current solution with multiple paragraph components below where only one is visible according to your current selection.

ivarne: Also note that there is no requirement for C# types and json schema to match 100%.

bjosttveit: Good point, the new dropdown and multiselect components in the design system actually do support formatted options, so we could at least add description support, maybe not help-text. Maybe I'll add support for that before I merge the updated dropdown/multiselect.

PR #1103, opened 2023-04-18
DevTools: Change component properties

bjosttveit:
## Description

I wanted something cool to show off on Altinn Kaffe, this allows you to modify component properties and see how it affects your layout. I think this makes the developer experience much better since you normally would have to refresh the entire page to try out small changes. Additionally, you don't have to make changes to your actual layout file and risk forgetting to change them back.

https://user-images.githubusercontent.com/47412359/232710147-9eb9ea75-4a5f-4afe-b7e0-df9ee5188e87.mov

### DIFF
--- a/src/features/devtools/components/LayoutInspector/LayoutInspector.module.css
+++ b/src/features/devtools/components/LayoutInspector/LayoutInspector.module.css
@@ -37,3 +37,42 @@
   white-space: nowrap;
   max-width: 0;
 }
+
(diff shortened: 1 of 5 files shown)


olemartinorg (approved):
🚀

ivarne:
Da mangler det bare å skrive endringene tilbake til json filene, så har vi erstattet altinn-studio.

PR #1108, opened 2023-04-19
WCAG/Change structure of summary content

### LINKED ISSUE #1029: WCAG fixes on the summary component
## Reading order
Today the reading order on the summary component is Label - "endre" button - Value. When using a screen reader this might be a little confusing. We should change it so that the button is the last element on summary element. 
**Example:** 
1. Is how it is today
2. How it should be
Frame 1
The screen reader use the order of the code, if we change it we would have hack it to behave how we want it to. Maybe the easiest solution is to change the design so that the button is visually also the last element. 
Skjermbilde 2023-03-22 kl  15 05 25

- [ ] Change the design so that the "endre" button is last in the reading order

## The button is not descriptive enough
Moved to a separate issue:  #1100

### Steps To Reproduce

Go to an app with a summary page and use a screen reader when testing

### Additional Information

_No response_

lassopicasso:
## Description
Changed the HTML-structure of summary content to make it more WCAG-friendly:
Before: `Label -> Edit Btn -> Summary`
After: `Label -> Summary -> Edit Btn`
Used grid-template-areas to make the structure look visually the same as before.

>Needed to update cypress, due to more layers made with the solution I came up with.
>Instead of adding more layers to these paths in cypress, I removed them and kept just one selector. If this isn't a proper way to do it, or/and could risk any harm later on, I will change it back and add extra layer instead :) 

## Related Issue(s)

- closes #1029 

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
  - [x] No changes made
- Support in Altinn Studio
  
- Sprint board
- Labels


src/layout/Summary/SummaryContent.tsx
```diff
@@ -0,0 +1,69 @@
...
+import type { ISummaryComponent } from 'src/layout/Summary/SummaryComponent';
+import type { LayoutNodeFromType } from 'src/utils/layout/hierarchy.types';
+import type { LayoutNode } from 'src/utils/layout/LayoutNode';
+
+export interface SummaryContentProps {
+  onChangeClick: () => void;
+  changeText: string | null;
+  label: React.ReactNode;
+  summaryNode: LayoutNodeFromType<'Summary'>;
+  targetNode: LayoutNode | any;
```
olemartinorg, line 17: Do we need `| any` here? That just removes typescript support for that property..

olemartinorg: I guess it's because the check for `'renderSummaryBoilerplate' in component` is not narrowing the LayoutNode down to just summary-renderable-components anymore. I think it's best to keep the code strict and check that either inside this React component or outside it (and pass in the narrowed type).

lassopicasso: Good point, thanks Ole Martin :)

PR #1114, opened 2023-04-21
DevTools: Basic expression playground

bjosttveit:
## Description

This is the basic expression playground I showed in today's demo. It currently only supports "global" expressions (you cannot select a component to run it from). My current thought is to be able to choose a component in the future, maybe adding a button to the layout inspector that will open the expression playground with that component selected, or maybe just add a dropdown. Even though this is missing in this first version, I think it still adds a lot of value for very little effort so I thought I may as well just release what I have now.

I don't expect a thorough review, but I mostly copied the code from the window.evalExpression, and just hardcoded the forComponentId to null, so it is quite trivial to add support for selecting a component later.

### DIFF
--- a/src/features/devtools/DevToolsControls.tsx
+++ b/src/features/devtools/DevToolsControls.tsx
@@ -3,6 +3,7 @@ import React from 'react';
 import { Tabs } from '@digdir/design-system-react';
 
 import { DevNavigationButtons } from 'src/features/devtools/components/DevNavigationButtons/DevNavigationButtons';
+import { ExpressionPlayground } from 'src/features/devtools/components/ExpressionPlayground/ExpressionPlayground';
 import { LayoutInspector } from 'src/features/devtools/components/LayoutInspector/LayoutInspector';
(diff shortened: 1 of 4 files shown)


olemartinorg (approved):
Very cool! 🚀 🥳 Would be nice to get a more descriptive error message about why the expression is invalid, but I can help fix that. Hang on, I'll look into it before this is merged!

olemartinorg:
With errors now! 🥳 

https://user-images.githubusercontent.com/700139/233940660-11e2f8ec-08c7-4efc-8302-66e657be36fa.mp4

bjosttveit:
Cannot see the video, but I assume it is good 😅 
image

olemartinorg:
Huh, I blame `wf-recorder` - maybe it works better if i re-encode it in `ffmpeg` once before publishing? 🤔 


https://user-images.githubusercontent.com/700139/233994066-75a1f4da-fdfa-4600-95a1-81eeb9b9827d.mp4

bjosttveit:
same thing 😅 

image

Arch must be great  🤣

PR #1119, opened 2023-04-25
Signing action

### LINKED ISSUE #991: Signing action
### Description

Create a new type of action/button that allows a user of a specified role to sign off on the form data and moves the process to the next step.

The sign action needs to be defined in XACML as a access controlled resource. Frontend needs to ask backend for what actions a user has access to on a given step in the process.

This is basically the same as todays submit button, just with additional information that turns it into a signing action.

### Additional Information
Related epic: https://github.com/Altinn/app-template-dotnet/issues/170
Related backend issue: https://github.com/Altinn/app-lib-dotnet/issues/205

### Tasks

_No response_

### Acceptance Criterias

- [ ] The sign button should only be clickable if the logged on user have access
- [ ] Once clicked the process should be moved to the next step
- [ ] Backend should be informed that it was a sign button that was clicked, allowing backend to "capture the signature"

bjosttveit:
## Description

This PR does a number of things:
- It reads out permissions from the endpoints `GET .../process` and `PUT .../process/next` (using the existing sagas) and stores this in processSlice in redux.
- It exposes these properties from the redux store to expressions using the new `authContext`, allowing you to use expressions based on what permissions the user has.
- The existing submit/confirm button gets disabled if write/confirm permissions are not present (only if authPermissions-backendFeature is enabled).
- Sends the performed action in the body of process.next, depending on the button you clicked.
- Adds a new component `ActionButton`, which will call process.next with a given action.

Questions:
- We agreed that actions should be generic, so that new actions can easily be added. Therefore, there is no specific logic that checks any specific actions, but I have still made an `IActionType`, and also specified the ones we know in the schema. There is also a check in the authContext expression if the action is a valid one. Should we still explicitly type them like this? Or should action simply be a string type?
- I was not 100% sure how we should do the button styling. For now I added two styles, one filled green, and one outlined primary color. Maybe we should just expose all of the options? One potential issue with that it is very specific to the implementation of the design system buttons, which could change and is not fully in our control. Feel free to suggest other solutions, or simply other names for the two styles I defined.

## Related Issue(s)

- closes #991 

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
  - [ ] No changes made
- Support in Altinn Studio
  
- Sprint board
- Labels


src/layout/Button/ButtonComponent.tsx
```diff
@@ -46,41 +55,50 @@ export const ButtonComponent = ({ node, ...componentProps }: IButtonReceivedProp
...
+            componentId,
+          }),
+        );
+      } else {
+        if (processActionsFeature) {
+          dispatch(ProcessActions.complete({ action: 'confirm' }));
+        } else {
+          dispatch(ProcessActions.complete());
+        }
+      }
```
olemartinorg, line 75: Maybe we should check the task type here as well? If not, suddenly a new task type is introduced by someone, we forget about this code snippet, and suddenly the 'confirm' action is sent on an entirely different task type.. 🤔 If I understood this correctly, that is. Might be better to have a fallback to just complete the process task without setting an action.

olemartinorg (approved):
Looks good! 🚀 As far as the questions go:

1. I'm not sure. I think I prefer typing, but maybe the schema could allow for any string as well as giving suggestions? With typing, we might end up implementing checks for `if (action === 'sign')` which might not be what you  want to call the signing action - or am I wrong? We should decide if the actions are static or entirely configurable, and probably choose one of them. I have no strong feelings either way.
2. I know people have asked before, so I suggest adding these button variant configurations to all the buttons! 🥳 If the design system pulls the rug out from underneath us, we just have to find a way around it (or influence them to keep the features we need). No need to worry about features disappearing from internal products, I think - that is easily solved with communication.

Some questions from me:
1. Are there plans/issues to implement the `authContext` expression function on the backend?
2. When you click the 'Send in' button now, it runs validations first and stops you from proceeding. However, the more generic `go-to-task` button does not (it can be used to proceed forwards in the process, but it was primarily thought to be used to go backwards). Does this new Action button validate the form first?

src/layout/Button/ButtonComponent.tsx
bjosttveit: good point

bjosttveit:
1. It should probably be implemented in the backend as well, since it can be used for anything really, so it probably needs similar support I guess, but I don't know how they will get these values from the policy yet.
2. It does not run validations as of now. I guess the validation only makes sense when writing data? And write is not defined as an action in this sense.

framitdavid (approved):
👏

PR #1125, opened 2023-04-27
Enable DevTools button in Studio preview

bjosttveit:
## Description

This adds Altinn Studio to the list of hostnames that should display the devtools button in addition to `local.altinn.cloud`. Tested by @standeren .


olemartinorg (approved):
Great!! 🚀

PR #1134, opened 2023-05-03
Render components as a summary by adding "renderAsSummary: true".

### LINKED ISSUE #996: Support summary component in another step
### Description
Currently it is not really possible to use `Summary` on a confirmation step. In the examples in the documentation, a new component that is readOnly is instead used. We would rather like to support the summary view. The problem with this currently, is that the layout of the data step is not available, and so there is no natural `componentRef`.

It is not obvious how this should be solved, some alternatives are:
1. Configure a "new" component, and somehow set its rendering mode to summary instead.
  - Simple "renderAsSummary": true.
2. Have the summary component refer to the datamodel instead of a different component.
  - Challenges: how do you configure the component settings.
3. Reference the layout and components of a previous step
4. Render summary when the task-type is confirmation.
  - Challenges: Goes against our wish to make tasks less "type dependent".

After a brief discussion we agreed to move on with alternative 1. This means we are extending the the actual components with a summary mode in addition to read and write. At the same time pushing some of the summary component specific settings that exists today until we know more about the needs.

Summary on a separate task than the one used to fill in the data serves a slightly different purpose than having a summary as the last page of a form. On the last page of a form the purpose is to quickly review the form and the values allowing to go back and correct the values. Summary on later step is about reviewing the data in order to make a decision and you might not even av access to modify the data. A decision in this case is to sign or reject based on the values provided effectively moving the process.

### Additional Information
- Edit options should be disabled as the previous step is completed. Other actions might set the process back allowing for editing. But the summary on a confirm step should probably be read only.

### Tasks
- [x] Identify all components that should have a summary mode
- [ ] Implement summary mode on identified components
- [ ] Document the setup and configuration and any differences with regards to the summary component

### Acceptance Criterias
- [ ] Ability to create a layout with components setting the mode to summary
- [ ] Trying to pass data updates to the server should return a Forbidden response

framitdavid:
## Description

Render components as a summary by adding "renderAsSummary: true".

## Related Issue(s)

- closes #996 

## Verification/QA

- Automated tests
  - [x] Has been added/updated
  
  - [ ] No changes made
- Support in Altinn Studio
  
- Sprint board
- Labels


lassopicasso (approved):
This looks good 😄

olemartinorg:
@framitdavid Looked over this now, and while the layout schema (and typescript types) indicate this could be using an expression, it is not configured as such. You need to add a configuration to `ExprConfigForComponent` in `src/features/expressions/index.ts` as well. TypeScript should have generated an error for this, but it seems that type has been broken somehow.

PR #1170, opened 2023-05-16
custom text to add-new-button

### LINKED ISSUE #1143: Ability to customize the entire text on the add button in repeating groups
### Description

On the add button-label for repeating groups, instead of adding additional text after "Legg til ny", like this "Legg til ny <custom text>". The ability to customize the entire text would be nice to have in several cases. So the new format will be "<custom text>", without necessarily having "Legg til ny".

Additionally, sometimes it makes more sense to have "Legg til nytt" - this would also be fixed by the ability to customize the entire text of the add button.

### Additional Information

image
Here the add button has gotten an additional "filial", for this case it works fine. In other cases we might want to change the entire text to i.e. "Legg til filial", or something entirely different.

lassopicasso:
## Description
Added new text-resource binding key to app developers. This enables use of `custom_add_button` to customize the whole text of add-button in repeating groups. 
`custom_add_button` overrules the default `add_new` and also `add_button` if it has been added.

## Related Issue(s)

- closes #1143 

## Verification/QA

- Automated tests
  - [x] Has been added/updated
  https://github.com/Altinn/altinn-studio-docs/pull/922
  
  - [x] No changes made
- Support in Altinn Studio
  
- Sprint board
- Labels

### DIFF
--- a/src/layout/Group/GroupContainer.tsx
+++ b/src/layout/Group/GroupContainer.tsx
@@ -50,7 +50,7 @@ export function GroupContainer({ node }: IGroupProps): JSX.Element | null {
   const multiPageIndex = groupState?.multiPageIndex ?? -1;
   const repeatingGroupIndex = groupState?.index ?? -1;
   const formData = useAppSelector((state) => state.formData.formData);
-  const { langAsString } = useLanguage();
+  const { lang, langAsString } = useLanguage();
 
   const filteredIndexList = React.useMemo(
     () => getRepeatingGroupFilteredIndices(formData, edit?.filter),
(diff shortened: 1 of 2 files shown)


src/layout/Group/GroupContainer.tsx
```diff
@@ -86,14 +84,16 @@ export function GroupContainer({ node }: IGroupProps): JSX.Element | null {
...
+              ? langAsString(resolvedTextBindings.custom_add_button)
+              : `${langAsString('general.add_new')} ${langAsString(resolvedTextBindings?.add_button) ?? ''}`
+          }
         />
       )}
-      {`${getLanguageFromKey('general.add_new', language ?? {})} ${
-        resolvedTextBindings?.add_button ? getTextResourceByKey(resolvedTextBindings.add_button, textResources) : ''
-      }`}
+      {resolvedTextBindings?.custom_add_button
+        ? langAsString(resolvedTextBindings.custom_add_button)
```
olemartinorg, line 95: Nice! Two nitpicks:
1. I don't think it has to be a string. We can get away with just using the `lang` function here, so that we support markdown, etc. Better to be flexible out of the box. The else-case here relies on string concatenation, so that still needs to be a string.
2. I'm not sure I like the `custom_add_button` name here. By definition, all text resources to you bind to a component are meant to set some custom value. How about `add_button_full` or something in that direction?

olemartinorg (approved):
🚀

src/layout/Group/GroupContainer.tsx
lassopicasso: Thx for the feedback, I have changed to lang for the if case, and kept langAsString in else case. 
I like the suggestion to change it to `add_button_full`, it makes more sense when you compare it with add_button in the docs.

Issue #1182, opened 2023-05-22
Performance: Make each component responsible for running expressions


olemartinorg:
Right now, the expression engine runs over all nodes in the tree every time something changes (even a value in the data model). Also, the node hierarchy is responsible for running all expressions, and it does not care about the component type when doing this. This is problematic in multiple ways:

- Even if the tree/hierarchy does not change (rows are not added/removed), the entire object hierarchy needs to regenerate for expressions to run again. This is wasteful and resource intensive.
- Because we re-create the tree all the time, we don't have any better indicators to tell rows apart from each other than the row index/component id. This is problematic because it [works against React optimizations](https://articles.wesionary.team/using-index-as-a-key-is-an-anti-pattern-in-react-8e5db3aea3a9).
- It's hard to implement expression support in irregular places, like deep inside the `Grid` component (as the expression engine did not have support for looking for expressions inside arrays at the time of creating the `Grid` component).
- Expression support seems like magic for our developers. I've seen people add schemas that indicate we support expressions without actually adding the expression config needed to make expressions work for a new property.
- ...and even more ways!

To fix this, each `LayoutComponent` class should get the responsibility for evaluating expressions inside the node.

### Relevant issue(s):
- #377

PR #1189, opened 2023-05-24
Feature: Link component

### LINKED ISSUE #665: Button: Add property for redirecting to a URL
### Description

The `Button` component seems to be the one to extend (i.e. when adding support for `go-to-task`). We should implement a mode that makes is possible to redirect the whole page to a specific URL (which should also be possible to define using an expression).

### Background
This need solved an issue for @elsand in an Altinn 3 app they're making. Their concrete use-case is that they need to redirect the user to an external page for signing/consent, which again needs to redirect back to the Altinn 3 app. This can be solved by:
1. Implementing a button that can redirect somewhere. By having it configurable with an expression, an URL can be constructed that passes the current instance ID to the backend, which can facilitate the consent request and redirect further to the actual consent page (and store the current instance ID in that context somehow).
2. Finishing #451 to make it possible to link back to a specific page in the Altinn 3 app (without fiddling with localStorage to inject the requested layout/page to display).

### Additional Information

- Slack dicussion: https://altinndevops.slack.com/archives/C045EB3JA9X/p1667991655772529

bjosttveit:
## Description

New component: `Redirect`. Can be styled as either a link or a button, with the same button styles as in action button. Can optionally open in new tab, default false. Target url is set in textresourcebindings to make expressions work.

Need to write documentation for this component:
- basic config
- create url with expression
- use frontend settings to have different url for different environments

## Related Issue(s)

- closes #665 

## Verification/QA

- Automated tests
  - [x] Has been added/updated
    - https://github.com/Altinn/altinn-studio-docs/pull/935
  
  - [ ] No changes made
- Support in Altinn Studio
  
- Sprint board
- Labels

### DIFF
--- a/schemas/json/layout/layout.schema.v1.json
+++ b/schemas/json/layout/layout.schema.v1.json
@@ -42,7 +42,7 @@
           "type": "string",
           "title": "Type",
           "description": "The component type.",
-          "enum": ["ActionButton", "AddressComponent", "AttachmentList", "Button", "ButtonGroup", "Checkboxes", "Custom", "Datepicker", "Dropdown", "FileUpload", "FileUploadWithTag", "Grid", "Group", "Header", "Image", "Input", "InstanceInformation", "InstantiationButton", "IFrame", "Likert","List", "MultipleSelect", "NavigationButtons", "NavigationBar", "Panel", "Paragraph", "PrintButton", "RadioButtons", "Summary", "TextArea"]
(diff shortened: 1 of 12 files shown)


lassopicasso (approved):
Nicely done! I like that you give appdev the option to redirect to a new tab or not.

olemartinorg (approved):
Looks good! Two things:
- It's more of a link than a redirect. Link is what is _is_, redirect is what it _does_, and most of our other components seem to be named after what they _are_ and not _what they do_. I.e., it's called `Button` and not `Submit` (for better or worse), and `ActionButton` instead of `PerformAction`. I think this component would be better named `Link`, IMO. It does not make sense to 'redirect' in a new tab, but it could be argued it does not make sense to 'link' with a button either.
- Tests are great, but it seems they would't break if the link/button is not clickable, or if nothing happens when you click it. Maybe a simple Cypress test could be used to verify the functionality of the button as well, not just the properties of the DOM elements.

bjosttveit:
Alright I can rename it to `Link`.

I honestly don't think it is useful to test if the native `a`-element is able to open a link specified in href. (currently, there is no way to disable it, see below) Additionally, the button test does check that window.open is called with the correct parameters, I don't think testing that this native function does what it is supposed to has that much value either.

One question though about it not being clickable. I thought about adding a prop for disabling it, but I don't see how this is useful without adding expression support, and due to #1182 I was unsure whether to add any special cases in the existing code or not.

Issue #1000, opened 2023-03-14
CSS order prevents overriding design system


bjosttveit:
### Description of the bug

In apps currently, the CSS from app-frontend is bundled to a separate file `altinn-app-frontend.css` which is loaded into each apps `Index.cshtml` file. The css from the design system is loaded dynamically by the javascript bundle into `style` tags in the head, and importantly, these are placed below the app-frontend css, meaning that it is difficult to override the design system CSS without using either `!important` or having greater specificity.

Example, changing a `Button` margin, when there is a default in the design system.

The following will not work (same specificity, design system takes precedent due to order):
![image](https://user-images.githubusercontent.com/47412359/224942825-ba6c3a31-83a0-4fb2-89f2-41bd03cbe861.png)

The following will work, due to higher specificity:
![image](https://user-images.githubusercontent.com/47412359/224942861-4cdd4a9e-2723-4180-acb4-baee4a692ac8.png)

Issue #952, opened 2023-02-23
Cypress: Visual regression testing from main


olemartinorg:
### Background and description
Today, Cypress is a great contribution to _functionality testing_. Meaning, we'll quickly catch unintended changes in functionality caused by changes. Personally I rely heavily on Cypress tests to uncover problems I may introduce when refactoring. Some examples:

- A bug that was introduced by my refactoring, but since it had no automatic test coverage, nobody noticed the bug until it got released into production:
  - #951
- A major rewrite of lots of smaller things (although it possibly lead to the bug above). During the implementation phase, I ran Cypress tests repeatedly to make sure functionality was kept intact. I found lots of functionality that broke because of the refactor, but because the tests are more complete and run faster than I can manage manually, I found and fixed them before releasing the changes. Faster feedback loop leads to increased productivity and better quality.
  - #861

So, with that in the back of our heads; one of the things our tests doesn't do for us, is verifying the visual integrity. Nor should it, you could argue, as we introduce new features that obviously lead to visual changes all the time. However, we also run these tests before merging upgrades to dependencies, and since we don't test the visuals, those could be broken severely by some unrelated change (unless we're careful and test this manually). I've made it a habit to sometimes carefully expect the visual representation of the app and make sure it is how we want it.

Since we already run Cypress tests _after merging a PR into main_ (but we mostly ignore the results, as they are not reported anywhere), we could use that run to produce "gold-standard" screenshots of how the app is supposed to look, and then diff against those screenshots when running tests in a PR. It would be nice to have a bot on github tell us via a comment which tests produced visual changes, and maybe a GIF showing us the differences. For dependency upgrades, we would expect there to be no visual changes from the latest main. And for new functionality (where there are new/changed tests, obviously 😉) the bot could report the fresh screenshots showing how the new feature(s) look visually - a nice-to-have when reviewing a PR.

### Considerations
- Where should we store the "gold standard" screenshots and metadata? Storing these in git would rack up disk usage pretty quickly, and only the last version of main really counts. Some kind of Azure storage location would probably be best. We would need to download these on every PR test run, and upload new ones after every (successful) main run. Can the Cypress cloud do anything here?
- How do we report these in a PR? My suggestion would be to never fail a test run because of visual changes, but report them so a human reviewing a PR can make an informed decision. We only want to catch _unintended visual changes_.
[...]

PR #1200, opened 2023-05-26
Added grouping indicator to non-repeating group

### LINKED ISSUE #436: Update group component UI to indicate what elements are within the group.
## Screenshots
Following screenshot is from Altinn 2.0: 
image

Screenshot from Altinn 3.0: 
 
image
(Note: Point 2.3 is not included in group 2.2, but is still aligned close to point 2.2.4, which makes it difficult to interpret that it is not in the same group.)

## Specification tasks

- [ ] Development tasks are defined

## Definition of done
Verify that this issue meets [DoD](https://digdir-digitalisering.atlassian.net/wiki/spaces/T3KP/pages/5049246/Definition+of+Done) (Only for project members) before closing.

- [ ] Documentation is updated (if relevant)
  - [ ] Technical documentation (docs.altinn.studio)
  - [ ] User documentation (altinn.github.io/docs)
- [ ] QA
- [ ] Manual test is complete (if relevant)
- [ ] Automated test is implemented (if relevant)
- [ ] All tasks in this userstory are closed (i.e. remaining tasks are moved to other user stories or marked obsolete)

### Design solution

If a group of input fields consists of a header, body text and input fields, the header and body text should be aligned and the input fields should be indented with a vertical line on the left side starting from the input fields. Conversely, in the case of groups within a group, the header, body text and input fields should be aligned with a vertical line on the left side starting from the header.

Skjermbilde 2023-05-25 kl  13 38 15

For input field components with a header, that are not created as part of groups, the input field is not indented with a vertical line to the left.  

Skjermbilde 2023-05-25 kl  09 42 16

https://www.figma.com/file/Pvti3aRcOwH0k5Z7DrLznk/Arbeidsomr%C3%A5de%3A-Altinn-Studio-APPS?type=design&node-id=4639%3A103560&t=ni8ThCJyqL5s97BF-1

bjosttveit:
## Description

Added optional groupingIndicator, line + indent, also added body text.

Anything about this that need unit/cypress testing?

I made some changes to how non-repeating groups are treated in summary to make it work a bit better. I made it so that largeGroup is true for nested non repeating groups, this will make it look more like how it does when filling out the form. I also made the bottomBorder not visible when not using summaryBoilerplate, this applies to group and grid, which have a custom rendering that makes it look better without it in my opinion. It looked very strange for non-repeating groups at least. TLDR; There were too many borders imo. 😅 

https://www.figma.com/file/Pvti3aRcOwH0k5Z7DrLznk/Arbeidsomr%C3%A5de%3A-Altinn-Studio-APPS?type=design&node-id=4745%3A110013&t=xrvnNTRiNxSh5H4u-1

The use of `!important` is due to #1000 i think, there are at least problems with overriding material-ui as well.

image

## Related Issue(s)

- closes #436 

## Verification/QA

- Automated tests
  - [x] Has been added/updated
    - https://github.com/Altinn/altinn-studio-docs/pull/944
  
  - [ ] No changes made
- Support in Altinn Studio
  
- Sprint board
- Labels

### DIFF
--- a/src/features/pdf/PDFView.tsx
+++ b/src/features/pdf/PDFView.tsx
@@ -11,6 +11,7 @@ import { DisplayGroupContainer } from 'src/layout/Group/DisplayGroupContainer';
 import { ComponentType } from 'src/layout/LayoutComponent';
 import { SummaryComponent } from 'src/layout/Summary/SummaryComponent';
 import { useExprContext } from 'src/utils/layout/ExprContext';
+import type { HNonRepGroup } from 'src/utils/layout/hierarchy.types';
 import type { LayoutNode } from 'src/utils/layout/LayoutNode';
 
 interface PDFViewProps {
@@ -22,7 +23,7 @@ const PDFComponent = ({ node }: { node: LayoutNode }) => {
   if (node.isType('Group')) {
     return (
       <DisplayGroupContainer
-        groupNode={node}
+        groupNode={node as LayoutNode<HNonRepGroup, 'Group'>}
         renderLayoutNode={(child) => (
           <PDFComponent
             key={child.item.id}
--- a/src/features/pdf/PDFView.tsx
+++ b/src/features/pdf/PDFView.tsx
@@ -11,7 +11,6 @@ import { DisplayGroupContainer } from 'src/layout/Group/DisplayGroupContainer';
 import { ComponentType } from 'src/layout/LayoutComponent';
 import { SummaryComponent } from 'src/layout/Summary/SummaryComponent';
 import { useExprContext } from 'src/utils/layout/ExprContext';
-import type { HNonRepGroup } from 'src/utils/layout/hierarchy.types';
 import type { LayoutNode } from 'src/utils/layout/LayoutNode';
 
 interface PDFViewProps {
@@ -20,10 +19,10 @@ interface PDFViewProps {
 }
 
 const PDFComponent = ({ node }: { node: LayoutNode }) => {
-  if (node.isType('Group')) {
+  if (node.isNonRepGroup()) {
     return (
       <DisplayGroupContainer
-        groupNode={node as LayoutNode<HNonRepGroup, 'Group'>}
+        groupNode={node}
         renderLayoutNode={(child) => (
           <PDFComponent
             key={child.item.id}
--- a/src/layout/Group/DisplayGroupContainer.tsx
+++ b/src/layout/Group/DisplayGroupContainer.tsx
@@ -1,34 +1,25 @@
 import React from 'react';
 
-import { Grid, makeStyles, Typography } from '@material-ui/core';
+import { Grid, Typography } from '@material-ui/core';
 import cn from 'classnames';
 
+import { ConditionalWrapper } from 'src/components/ConditionalWrapper';
 import { useAppSelector } from 'src/hooks/useAppSelector';
-import { pageBreakStyles } from 'src/utils/formComponentUtils';
+import classes from 'src/layout/Group/DisplayGroupContainer.module.css';
+import { pageBreakStyles, selectComponentTexts } from 'src/utils/formComponentUtils';
 import { getTextFromAppOrDefault } from 'src/utils/textResource';
+import type { HNonRepGroup } from 'src/utils/layout/hierarchy.types';
 import type { LayoutNode } from 'src/utils/layout/LayoutNode';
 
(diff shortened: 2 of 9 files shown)


src/features/pdf/PDFView.tsx
```diff
@@ -22,7 +23,7 @@ const PDFComponent = ({ node }: { node: LayoutNode }) => {
   if (node.isType('Group')) {
     return (
       <DisplayGroupContainer
-        groupNode={node}
+        groupNode={node as LayoutNode<HNonRepGroup, 'Group'>}
```
olemartinorg, line 26: Using `node.isNonRepGroup()` above would give you this automatically. By casting, you're forcing all groups to appear as non-repeating groups, which is not quite right.

src/layout/Group/DisplayGroupContainer.tsx
```diff
@@ -52,20 +47,45 @@ export function DisplayGroupContainer({ groupNode, id, onlyRowIndex, renderLayou
       data-testid='display-group-container'
       data-componentid={container.id}
     >
-      {title && (
+      {(title || texts.body) && (
```
olemartinorg, line 50: Btw, I just checked, and found no apps that already define a `body` text resource bindings on a `Group` component, so this should not break any existing apps. 🥳

Just for reference, I checked by adding this snippet to the existing one in `all.test.ts`:
```typescript
    for (const layout of Object.values(layouts)) {
      for (const component of layout || []) {
        if (component.type.toLowerCase() === 'group' && (!component.maxCount || component.maxCount <= 1)) {
          expect(component.textResourceBindings?.body).toBeUndefined();
        }
      }
    }
```

src/layout/Group/GroupRenderer.tsx
```diff
@@ -47,7 +49,7 @@ export function GroupRenderer({ node }: GroupRendererProps) {
   return (
     <DisplayGroupContainer
       key={node.item.id}
-      groupNode={node}
+      groupNode={node as LayoutNode<HNonRepGroup, 'Group'>}
```
olemartinorg, line 52: Same as the other comments - I would prefer to wrap the whole thing in `if (node.isNonRepGroup())` so that errors can be caught via the type system instead of just silencing any errors by casting the type.

src/layout/Group/SummaryGroupComponent.tsx
```diff
@@ -79,7 +79,7 @@ export function SummaryGroupComponent({
             <DisplayGroupContainer
               key={`summary-${targetNode.item.id}-${idx}`}
               id={`summary-${targetNode.item.id}-${idx}`}
-              groupNode={targetNode}
+              groupNode={targetNode as LayoutNode<HNonRepGroup, 'Group'>}
```
olemartinorg, line 82: Same here - better to wrap in `node.isNonRepGroup()` than to cast. Also, why the need to check if `targetNode` is a non-repeating group below here if you've already told the type system it always is? 🤔

olemartinorg (approved):
I agree, no need for any tests here - better to implement #952 then.. ;-)

src/features/pdf/PDFView.tsx
bjosttveit: you are absolutely right

src/layout/Group/DisplayGroupContainer.tsx
bjosttveit: Group with panel already has a body so that is a little surprising

src/layout/Group/GroupRenderer.tsx
bjosttveit: 👍  what should the default return be in that case? null?

src/layout/Group/SummaryGroupComponent.tsx
bjosttveit: I guess I moved a little fast with these 😅

src/layout/Group/DisplayGroupContainer.tsx
olemartinorg: Yup, I agree.. 🤔 Maybe nobody has started actually using that, though - it was only re-introduced lately. I updated all the apps to make sure, but still couldn't find any such configuration - so maybe it's just because nobody has released any such thing on tt02 or prod.. I actually found no non-repeating groups with both `panel` and `textResourceBindings` at all. 🤷

olemartinorg (approved):

Issue #1175, opened 2023-05-22
Experiment: Storing the data model in a Context


olemartinorg:
This issue describes an approach to restructure our data model (in-memory) storage for app-frontend, extracted from #345.

### Background

Currently, we're using Redux and some complex sagas to manipulate the data model, and we're storing everything in dot-notation internally. We do this because Redux has trouble with deeper stuctures, and I trust that former developers has gone down that route and landed on Redux and dot-notation for a reason. [It even says this is a good thing in our docs](https://docs.altinn.studio/technology/tools/react/architecture/#normalized-redux-store).

Right now, this structure has multiple downsides, mainly:
- Empty objects cannot exist in dot-notation. This has lead to complex state management for adding a new row to a repeating group (because until you fill out some data in that row, it is essentially an empty object). We cheat by increasing the `index` property on our repeating group state, without ever letting the server know that a new row was added. Changing this, as proposed in #555, would be a breaking change.
- With empty object being deleted by us, we actually change the data model and send our changes back to the server with empty objects removed. This has caused problems in the past for app-developers who wants to initialize empty objects in their data structure first, and use them in later in requests.
- Without proper/good tools, and transparency in code making smells very visible, we've fallen into a routine where state management _has_ to be done using dot notation. That means, if you want to add/remove items in an array, you're in a world of pain (where you have to unset keys in the dot-notation object and 'move up' later values in the array). Proper array tools would have made this much easier for us, and would have made #273 a breeze to implement (instead of being in the backlog for years).
- With complex state handling comes complex workarounds for seemingly simple problems. In #155, I made a `useDelayedSaveState` hook that debounces input from the user so that we _wait a while_ before storing the change in Redux (and subsequently calling the server). In hindsight, our data model storage should more intelligently handle multiple changes in rapid succession, and wait to save the data on the server when our state has stabilized. Multiple/rapid updates to the data model currently works, but has caused multiple problems (see one of the outstanding issues in #925 and the bug reported in #989).

### Proposed solution

Instead of swapping out the entire state manager (as it is sort-of proposed in #345), this issue suggests creating a React Context for the data model storage, and provide tools to manipulate and fetch data in that context. This means we could still have reducers and complex tasks that would happen as the data model updates, but by implementing this ourselves we get better creative control over how this is handled.

A few requirements for this context should be:
- It should expose tools tailored to our needs, such as a hook that binds to a data model binding (or multiple?) and gives you functions to update the state.
[...]

PR #1202, opened 2023-05-26
Refactoring layout/node type mapping

olemartinorg:
## Description
Fixing a few issues:
- Our master-list of components can cause merge conflicts if forks try adding their own layout components, because we all have to maintain the master list. This change moves the list out from git, and instead auto-generates it in an ignored file (`src/layout/components.ts`, which is generated by running `yarn gen`, which should also happen automatically for all relevant operations). This should make a solution as described in #200 more feasible.
- The type definitions for each component before and after node hierarchy generation was confusing. The default used to be `ExprResolved<YourType>`, and if you had special needs (such as adopting children and including other LayoutNode objects in your post-hierarchy type), you had to manually override that in `hierarchy.types.d.ts`. This should now just be set in the component config to take effect.
- Renamed the folder for the `Iframe` folder to match its component type name (`IFrame`)
- Cleaned up and simplified some types. Now `HComponent<T>` and `AnyItem<T>` is practically the same (although the first one does not support groups).. 🤷
- Removing a sneaky `disabled` property that could be optionally set on all layout components, but was never read by any code.
- Moving the `NotInLayout` type. It was now very easy to move these property extensions (such as `baseComponentId`) to indicate that they are only present on nodes in the hiearchy (as that's what add these properties), not some magic extensions to the layout definitions.
- Making it possible for a component type to extend the `LayoutNode` class and construct the extended class in the hierarchy. This will make it possible for components to add functionality here, such as having a `LayoutNodeRepGroup` type.

## Related Issue(s)
- #1175
- #200

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
  - [x] No changes made
- Support in Altinn Studio
  
- Sprint board
- Labels


bjosttveit (approved):
Very cool! I like that you split the config and type config, much cleaner this way!

PR #1204, opened 2023-05-26
Respect `doNotPromptForParty` profile preference

### LINKED ISSUE #268: Party selection is not shown even though the user wants to see party list with every instantiation
## Describe the bug
Party selection is not shown eventhough the user wants to see party list with every instantiation

## To Reproduce
1. Login to AT21 and navigate to profile and uncheck the checkbox 'Jeg ønsker ikke å bli spurt om aktør hver gang jeg starter utfylling av et nytt skjema' under Advanced settings
2. Start a deployed app that can be started by a person, organisation
3. The party selection page is not shown to select a different actor 

## Expected behavior
- [ ] The party selection should be shown with every instantiation (since in profile advanced settings, it is set that user will be asked to choose party with every instantiation). 
- [ ] If the user doesn't have multiple parties to represent the party selection should not be shown.
- [ ] This fix should only apply when creating a new instance, not when opening an existing instance.

## Additional info
Browser: Chrome 76
App: partydisplaytest
Env: AT21

bjosttveit:
## Description

Respect the profile preference that implies that you should always be presented with party selection on new instantiation unless you specifically ask not to have this happen.

What happenes now?
1. doNotPromptForParty is by default unchecked in the altinn profile
2. If you have more than one valid party that can instantiate, it will always show party selection
3. If you have exactly one valid party it will not do this

To make this possible I had to always pull the list of valid parties and store in state.party.parties, before, it would only get this list when entering party selection, and would just put the current party in otherwise. This change will possibly fix https://github.com/Altinn/app-lib-dotnet/issues/244 as well.

I manually updated the profile preferences for all of the test users in tt02 that we use in cypress to have `doNotPromptForParty=true`, that way all the existing tests work without modification. I have also updated this in `app-localtest`: https://github.com/Altinn/app-localtest/pull/42

## Related Issue(s)

- closes #268 

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
  - [x] No changes made
- Support in Altinn Studio
  
- Sprint board
- Labels


Magnusrm (approved):
Amazing work! 🚀

olemartinorg:
Isn't this, in a way, a breaking change? Behavior seems to change from how it was before, at least. I know we'll get complaints if we just roll this out in a patch release, even if it just a bugfix for a property we didn't properly respect before. How do we manage this to properly notify everyone? Calling on @RonnyB71. I know at least some external cypress tests will fail if this selector suddenly interferes - we're not the only ones running e2e-tests on apps.

bjosttveit:
Okey so what should be done about this?

1. I have not added a way for specific apps to ignore this setting, but you want that to be possible @RonnyB71 ? Where should this configuration live? `applicationMetadata`? Should it be opt-in or opt-out?
2. Having some text explaining why the user is seeing the party selection every time and informing them how to change their preference is a good idea. Should we add that too @RonnyB71 ?
3. As for the potential failing external cypress tests @olemartinorg , what do you suggest we do about that? If we make the setting opt-in, it will essentially not exist since I dont see why app developers will bother opting in to something like this 🤷 Should this be opt-in before a new major version is released, and then on by default?

olemartinorg:
> Should this be opt-in before a new major version is released, and then on by default?

That sounds like a good approach to me! 🤔

Another way could be to keep the implementation as it is, but make it a feature toggle in the app-frontend code. I'm doing that for #1175 in [this file here](https://github.com/Altinn/app-frontend-react/blob/experiment/new-form-data-storage2/src/features/toggles.ts), and then just using an if/else everywhere in the code so that I can toggle the new/old functionality. That way, we can remove the constant in v4 and roll it out then - in essence making breaking changes _now_ that are deactivated until everyone are ready to start v4 (and avoiding having to maintain and merge back and forth between a main and v4 branch that are guaranteed to get out-of-sync quickly when we start making large sweeping changes).

bjosttveit:
@olemartinorg let me know if you think the feature toggle strategy is fine

src/hooks/useDoNotPromptForParty.ts
```diff
@@ -0,0 +1,34 @@
+import { useAppSelector } from 'src/hooks/useAppSelector';
+import { useInstanceIdParams } from 'src/hooks/useInstanceIdParams';
+import type { IAltinnWindow } from 'src/types';
+
+export function useDoNotPromptForParty(): boolean | null {
```
olemartinorg, line 5: I don't quite like the double negative (if `doNotPromptForParty === false` it means to prompt for party), but I guess it's closer to the profile setting.

src/features/instantiate/containers/PartySelection.tsx
```diff
@@ -80,21 +80,36 @@ export const PartySelection = () => {
...
     // Clear any previous instantiation errors.
     dispatch(InstantiationActions.instantiateRejected({ error: null }));
   };
 
+  const { org, app } = window as Window as IAltinnWindow;
+  useEffect(() => {
+    if (selectedParty && hasSelected) {
+      navigate('/');
+    }
+  }, [selectedParty, hasSelected, navigate, org, app]);
```
olemartinorg, line 111: Not quite sure what `org` and `app` is doing here

src/language/texts/en.ts
```diff
@@ -232,6 +232,10 @@ export function en() {
       unit_personal_number: 'pers. number',
       show_deleted: 'Show deleted',
       show_sub_unit: 'Show sub units',
+      why_seeing_this: 'Why am I seeing this?',
+      seeing_this_preference:
+        'You can change your profile settings to not get prompted for party selection each time you start a new instance. You can find this setting under **Profile** > **Advanced settings** > **Do not ask what party I represent each time I start to fill in a new form**.',
```
olemartinorg, line 237: Not sure, but can we link to the profile settings as well? 🤔

olemartinorg (approved):
Looks very good! 🥳 I absolutely agree with the feature toggle functionality. In time I guess we should also make it controllable using a cookie (read on app load, set for each app using the `/{org}/{app}` path), and extend the developer tools to let you toggle on/off features in preview for the next major version (which should reload the page using that feature).

src/hooks/useDoNotPromptForParty.ts
bjosttveit: I know, this has been painful to wrap my head around the whole time, but the setting is literally called `doNotPromptForParty`, and `false` means prompt for party every time. 😅

src/features/instantiate/containers/PartySelection.tsx
bjosttveit: Nice catch, probably forgot to remove this after some experimentation

src/language/texts/en.ts
bjosttveit: Probably

src/hooks/useDoNotPromptForParty.ts
bjosttveit: I guess I could change the hook to be easier to understand

PR #1209, opened 2023-05-30
New functions to expressions

### LINKED ISSUE #1176: Expressions: New simple functions
Extracted from #646.

We should implement some new simple functions in the expression language. All of these implementations should be fairly easy, but if any of these presents problems during the implementation phase, new issues should be created for them specifically.

List of functions we need:
- [x] `round`
  - Rounds a number to an integer, or optionally a decimal with a configurable amount of decimal points
- [x] `text`
  - Should look up a value from a text resource, or from a language key. Additional arguments should be considered parameters to the text resource, if any are provided. Returns `string` or `null`
- [x] `strLength` / `stringLength` / `numCharacters`
  - Function taking a sting as an argument and returns the number for characters in the string (note: the number of characters is not the same as the number for bytes)
- [x] `language`
  - Function with no parameters, returns the currently selected language (i.e. `nb`, `en`, etc).
- [x] `contains`
  - String-checking function to check if the second parameter (a `string`) is inside the first parameter (also a `string`)
- [x] `notContains`
  - Synonym for `["not", ["contains", ...]]`
- [x] `endsWith`
  - String-checking function to check if the first parameter ends with the string in the second parameter
- [x] `startsWith`
  - Same as the above, but for the beginning of the text
- [x] `commaContains`
  - Special function for checking if a value is included in a comma-separated list of strings.
  - Can also be represented as a shortcut to `["or", ["equals", string, input], ["startsWith", string, ["concat", input, ","]], ["endsWith", string, ["concat", ",", input]], ["contains", string, ["concat", ",", input, ","]]]`
- [ ] `displayValue` **Created a separate [issue](https://github.com/Altinn/app-frontend-react/issues/1208)**
  - Looks up a component nearby, and presents its display value (which should work for number-formatted Input fields). See comment here: https://github.com/Altinn/app-frontend-react/issues/1105#issuecomment-1536240868
- [x] `upperCase`
  - Takes a string, turns it into UPPERCASE
- [x] `lowerCase`
  - Takes a string, turns it into lowercase

List of functions that require [list support](https://github.com/Altinn/app-frontend-react/issues/1179) in the expression language:
- [ ] `sum`
  - Should take any amount of numbers as input, and also lists of numbers. All these numbers should be summed, and the resulting number should be returned. `null` and non-numeric values should be considered to be `0` (i.e. no effect on the sum).
- [ ] `average`
  - Same as above, but should produce the average of all the numbers summed. `null` and non-numeric values should be considered entries when calculating the average, as we should make a filtering function for those cases where invalid values should not be considered instead.
- [ ] `in`
  - Shortcut function similar to `equals`, but compares the first argument against multiple others to check if the first argument matches either one of them. Should support lists of items as parameters, and should then check inside the lists as well.
- [ ] `notIn`
  - Synonym for `["not", ["in", ....]]`

framitdavid:
## Description

Created new functions to be available for expressions, please see the issue for more details. 

_Working on documentation while this PR is in review._

## Related Issue(s)

- closes #1176 

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
  - [x] No changes made
- Support in Altinn Studio
  
- Sprint board
- Labels


bjosttveit (commented):
Very nice! And great test coverage! Just one question, will expressions using state.textResources or selectedAppLanguage reevaluate if the language changes?

framitdavid:
> Very nice! And great test coverage! Just one question, will expressions using state.textResources or selectedAppLanguage reevaluate if the language changes?

@bjosttveit great question! 👍 I was testing this by console.log **"running language function"** inside the language expression/function and it re-runs every time I change the language within the app.

bjosttveit (approved):

olemartinorg:
Ohh, I missed the review window on this one! 😬 I see you used the window object directly to get the current state instead of properly adding the state we depend on to `dataSources`. That's an immediate problem I think we should get sorted, but I'll do a more thorough review later.

framitdavid:
Docs PR is ready for review here. https://github.com/Altinn/altinn-studio-docs/pull/946 :)

PR #1215, opened 2023-05-31
WCAG: Party selection buttons

bjosttveit:
## Description

This is a quick fix that adds `role=button` to the party selection items, these ones:

image

 This has annoyed me for a while since i use [vimium](https://chrome.google.com/webstore/detail/vimium/dbepggeogbaibhgnhhndojpepiihcmeb) to navigate in the browser and it could not detect these as clickable since it was just a div with no role. It was likely not great to use with screen readers either.

### DIFF
--- a/src/layout/Group/DisplayGroupContainer.tsx
+++ b/src/layout/Group/DisplayGroupContainer.tsx
@@ -9,7 +9,7 @@ import classes from 'src/layout/Group/DisplayGroupContainer.module.css';
 import { pageBreakStyles, selectComponentTexts } from 'src/utils/formComponentUtils';
 import { LayoutNode } from 'src/utils/layout/LayoutNode';
 import { getTextFromAppOrDefault } from 'src/utils/textResource';
-import type { HGroups } from 'src/utils/layout/hierarchy.types';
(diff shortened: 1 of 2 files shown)


src/layout/Group/DisplayGroupContainer.tsx
```diff
@@ -9,7 +9,7 @@ import classes from 'src/layout/Group/DisplayGroupContainer.module.css';
 import { pageBreakStyles, selectComponentTexts } from 'src/utils/formComponentUtils';
 import { LayoutNode } from 'src/utils/layout/LayoutNode';
 import { getTextFromAppOrDefault } from 'src/utils/textResource';
-import type { HGroups } from 'src/utils/layout/hierarchy.types';
```
olemartinorg, line 12: <3

olemartinorg (approved):

src/layout/Group/DisplayGroupContainer.tsx
bjosttveit: I think this broke in main, I have had to do it in another PR as well or tests wont run 😅

olemartinorg: Same here..!

PR #1217, opened 2023-05-31
Expressions: Refinements to new functions

olemartinorg:
## Description
I missed the review window to #1209, but it turns out I had quite a few more comments to add, so I would have just made @framitdavid's day much worse, and bothered him unnecessarily before his holiday...

Here's the changes I would have suggested:
- Do not throw runtime errors unless we have to. I've said before that we should keep the expression language strict, and I still think that it should for comparisons, etc - but we should avoid runtime errors as much as possible. When a runtime error occurs (potentially deep inside a function), the entire expression fails, and a default value is returned. Since the language has no error handling (yet), we force developers to make extremely complex expressions to check if every value is `null` to make sure no runtime errors occcur. We can be sure that a `null` never ends with the given string, so these functions (such as `endsWith`) should just return the expected `boolean` instead of throwing an exception that bubbles up the stack.
- Adding support for these functions to JsonSchema. To make sure we don't miss that for future built-in functions, I added a unit test to check the schema and validate against it.
- Making the last parameter for the `round` function optional (and implementing support for that in the validator), so that we don't have to pass `null` when you want to round to the nearest integer.
- Returning `nb` as the default `language` if the profile does not have any information. That's the default in our apps as well, and `null` as the default here is not really useful.
- Fixing autocompletion for function names in vscode by using `anyOf` instead of `oneOf`

## Related Issue(s)

- #1209 
- #1176

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
  - [ ] No changes made
- Support in Altinn Studio
  
- Sprint board
- Labels


bjosttveit (approved):
Great stuff 🥇

PR #1235, opened 2023-06-07
Update @digdir/design-system-react to 0.14.0

bjosttveit:
## Description

Only other change is removing the z-index workaround for dev-tools and adjusting the z-index of the dev tools panel. It is still not possible to set the z-index for select, but I now get the same results as before without the workaround.


Magnusrm (approved):
Looking good! 👍

PR #1258, opened 2023-06-16
Add new RadioButton and RadioGroup components.

### LINKED ISSUE #1247: Option for displaying RadioButtons as Cards
### Description

We want to be able to optionally display `RadioButtons` with a "Card" design. The content of the card might be solved by the existing solution for label and description, as they support markdown.

Configuring this option could be done by setting a property (e.g. `showAsCard`) on the `RadioButtons` component type in `FormLayout.json`

## Proposed technical solution
As we see it, this needs to be solved with a new `RadioGroup` component which can be composed with a list of `RadioButton` components from the `@digdir/design-system`.   

The `RadioGroup` should:
- Validate that only `RadioButtons` are direct descendants. 
- Make external state handling possible, so that state handling can be customized.
- Be accessible for screen readers. 

### Additional Information

The new `RadioGroup` should be moved to the design system at some point. 

Screenshot:
image

[RadioButton as card design in DGM](https://www.figma.com/file/f3FeSdA0yWgFlJq4AoTNTl/Konseptskisser?type=design&node-id=1384-65842&t=tXihA5ke01tnmevx-4)

mikaelrss:
## Description

- Added a new `RadioButton` component which has the option of being rendered with a `Card` wrapped around the component. 
- Added a `RadioGroup` component which accepts `RadioButtons` as children.
- Refactored the old `ControlledRadioGroup` component to use this new `RadioGroup` component instead of using `RadioGroup` from `@digdir/design-system`. 

## Related Issue(s)

- closes #1247 

## Verification/QA

- Automated tests
  - [x] I have tested with a screen reader/keyboard navigation/automated wcag validator Tested with Orca and TalkBack on Android.
  - [x] Has been added/updated Altinn/altinn-studio-docs#991
  
  - [ ] No changes made
- Support in Altinn Studio
  
- Sprint board
- Labels

### DIFF
--- a/src/components/form/RadioButton.module.css
+++ b/src/components/form/RadioButton.module.css
@@ -1,0 +1,29 @@
+.cardLabel {
+  position: absolute;
+  width: 100%;
+  height: 100%;
+  left: 0;
+  top: 0;
+  cursor: pointer;
+}
+
+.card label:focus-within {
+  box-shadow: none;
+  outline: none;
+}
+
+.card {
+  border-radius: 6px;
+  background-color: var(--fds-semantic-background-subtle);
+  padding: var(--fds-spacing-3);
+  position: relative;
+  display: flex;
+}
+
+.card:focus-within {
+  --fds-focus-border-width: 3px;
+
+  box-shadow: 0 0 0 var(--fds-focus-border-width) var(--fds-inner-focus-border-color);
+  outline: var(--fds-focus-border-width) solid var(--fds-outer-focus-border-color);
+  outline-offset: var(--fds-focus-border-width);
+}
--- a/src/components/form/RadioButton.module.css
+++ b/src/components/form/RadioButton.module.css
@@ -1,3 +1,5 @@
+@import '@digdir/design-system-tokens/brand/digdir/tokens.css';
+
 .cardLabel {
   position: absolute;
   width: 100%;
@@ -5,6 +7,7 @@
   left: 0;
   top: 0;
   cursor: pointer;
+  z-index: -1;
 }
 
 .card label:focus-within {
@@ -18,10 +21,12 @@
   padding: var(--fds-spacing-3);
   position: relative;
   display: flex;
+  z-index: 0;
 }
 
 .card:focus-within {
   --fds-focus-border-width: 3px;
+  background: var(--fds-semantic-surface-neutral-selected);
 
   box-shadow: 0 0 0 var(--fds-focus-border-width) var(--fds-inner-focus-border-color);
   outline: var(--fds-focus-border-width) solid var(--fds-outer-focus-border-color);
--- a/src/components/form/RadioButton.module.css
+++ b/src/components/form/RadioButton.module.css
@@ -1,5 +1,3 @@
-@import '@digdir/design-system-tokens/brand/digdir/tokens.css';
-
 .cardLabel {
   position: absolute;
   width: 100%;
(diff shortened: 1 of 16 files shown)


olemartinorg (approved):
Looks good! 🚀 I couldn't find the unit tests you updated/added - but other than that, I think this is ready to merge. We should possibly also add this new Card view to some radio buttons in the `frontend-test` app we use for Cypress testing, so that we also get screenshots for it (and thus visual testing).

src/layout/layout.d.ts
```diff
@@ -52,6 +52,7 @@ export interface IComponentRadioOrCheckbox<T extends Extract<ComponentTypes, 'Ra
   extends ILayoutCompBase<T>,
     ISelectionComponent {
   layout?: LayoutStyle;
+  showAsCard?: boolean;
```
Magnusrm, line 55: Will the Checkboxes-component now also get this property? If so i think we should not add it to the Checkboxes-component until this is implemented as a feature.

Magnusrm (approved):
Nice work! Just a question below 😊

src/layout/layout.d.ts
mikaelrss: Ah yes. Will have a look and see if I can fix this issue.

mikaelrss:
> Looks good! rocket I couldn't find the unit tests you updated/added - but other than that, I think this is ready to merge. We should possibly also add this new Card view to some radio buttons in the `frontend-test` app we use for Cypress testing, so that we also get screenshots for it (and thus visual testing).

Ah you're correct. I didn't update any unit tests, I had to update some functionality to make the current unit tests pass :sweat_smile:

mikaelrss:
We've decided to make some changes to `@digdir/design-system-react`, and will forward a `ref` for the input-field instead of solving it this way. This means we need to change the implementation in `RadioButton` slightly.

src/components/form/RadioButton.tsx
```diff
@@ -0,0 +1,37 @@
...
+export interface IAppRadioButtonProps extends RadioButtonProps {
+  showAsCard?: boolean;
+}
+
+export const RadioButton = ({ showAsCard = false, ...rest }: IAppRadioButtonProps) => {
+  const inputRef = useRef<HTMLInputElement>(null);
+  if (showAsCard) {
+    return (
+      <Card>
+        {/* eslint-disable-next-line jsx-a11y/click-events-have-key-events,jsx-a11y/no-static-element-interactions */}
```
mikaelrss, line 19: This element is only clickable for visual effects. A screen reader would only want to click the inner `input` element of the `DesignSystemRadioButton`.

Magnusrm: Maybe we could add this explanation in a comment in the code to help future maintainers 😉

Magnusrm (approved):
New changes look good to me! 🚀

src/components/form/RadioButton.module.css
```diff
@@ -0,0 +1,47 @@
+@import '@digdir/design-system-tokens/brand/digdir/tokens.css';
```
olemartinorg, line 1: As we already import the altinn brand tokens, will this affect the global tokens? 🤔 What's the reason we have to import the digdir branding here - I'd at least want to include a comment about it so that we can consider when we can remove it safely again (i.e. when this works with the Altinn branding).

src/components/form/RadioGroup.module.css
```diff
@@ -0,0 +1,13 @@
+@import '@digdir/design-system-tokens/brand/digdir/tokens.css';
```
olemartinorg, line 1: Again here, with the digdir branding

olemartinorg (approved):
Looks good! 🥳

src/components/form/RadioButton.module.css
mikaelrss: Seems like we don't have to import the digdir branding here. Just a mistake from my part :sweat_smile:

PR #1234, opened 2023-06-06
1188 summary component - enable title and accessibleTitle fields in text resource bindings

lassopicasso:
## Description
Added 2 new text bindings for the summary components:
- `title` - behaves the same way as the `title` that is received from the reference component. But this one overrides if both are set.
- `accessibleTitle` - if app developers wants to customize the aria-label of the edit btn this can be used, and overrides `title` that is otherwise used. 

## Related Issue(s)

- closes #1188 

## Verification/QA

- Manual functionality testing
  - [x] I have tested these changes manually
  - [ ] Creator of the original issue (or service owner) has been contacted for manual testing (or will be contacted when released in alpha)
  - [ ] No testing done/necessary
- Automated tests
  - [ ] Unit test(s) have been added/updated
  - [x] Cypress E2E test(s) have been added/updated
  - [ ] No automatic tests are needed here (no functional changes/additions)
  - [ ] I want someone to help me make some tests
- UU/WCAG ([follow these guidelines](https://aksel.nav.no/god-praksis/artikler/utvikling?tema=universell-utforming) until we have our own)
  - [x] I have tested with a screen reader/keyboard navigation/automated wcag validator
  - [ ] No testing done/necessary (no DOM/visual changes)
  - [ ] I want someone to help me perform accessibility testing
- User documentation @ [altinn-studio-docs](https://github.com/Altinn/altinn-studio-docs)
  - [x] Has been added/updated
https://github.com/Altinn/altinn-studio-docs/pull/961
  - [ ] No functionality has been changed/added, so no documentation is needed
  - [ ] I will do that later/have created an issue
  
- Changes/additions to component properties
  - [ ] Changes are reflected in both `src/layout/layout.d.ts` and `layout.schema.v1.json`, and these are all backwards-compatible
  - [x] No changes made
- Support in Altinn Studio
  - [ ] Issue(s) created for support in Studio
  
  - [x] This change/feature does not require any changes to Altinn Studio
- Sprint board
  - [x] The original issue (or this PR itself) has been added to the Team Apps project and to the current sprint board
  - [ ] I don't have permissions to do that, please help me out
- Labels
  - [x] I have added a `kind/*` label to this PR for proper release notes grouping
  - [ ] I don't have permissions to add labels, please help me out

PR #1278, opened 2023-06-22
Fix: Title and Aria-Label on Summary Page

lassopicasso:
## Description
The [new text resource bindings](https://github.com/Altinn/app-frontend-react/pull/1234) in the summary component don't work as intended when referencing a group component.

Since the current solution is not yet deployed, we have opted for an alternative approach. Instead of adding text bindings directly to the summary component, there are introduced two new text bindings in the components that are summarized. 
These bindings are specifically meant to be displayed only in the summary:
- `summaryTitle`: set own title for the summary component. If this field is set, it overrides the title field.
- `summaryAccessibleTitle`: customize the aria-label for the edit button. If this field is not set, the summaryAccessibleTitle will be used. If neither field is set, the title will be used.

## Related Issue(s)

https://altinn.slack.com/archives/C02EVE4RU82/p1687420564190999

## Verification/QA

- Automated tests
  - [x] Has been added/updated
  
  - [x] No changes made
- Support in Altinn Studio
https://github.com/Altinn/altinn-studio-docs/pull/1002
- Sprint board
- Labels


lassopicasso:
@olemartinorg 
I managed to fix the specific issue encountered by SSB. 
One potential problem I noticed when the app is that text resource bindings in the summary component also override the titles in each individual child within the group. 

This is because in the [original solution](https://altinn.slack.com/archives/C02EVE4RU82/p1687420564190999), I set the text resource bindings in the summary component to override text resource bindings it references to.

In my opinion, it might be better if the title key in the referenced component is not overridden by the summary component (I will update the docs accordingly). However, an accessibleTitle in the summary component should still override the title for the aria-label in the edit button.

Do you think this approach could potentially create any problems?

image

olemartinorg:
> In my opinion, it might be better if the title key in the referenced component is not overridden by the summary component

Does that mean that setting a `title` on a `Summary` component will have no effect unless the target component also does not have one? 🤔 I would assume the `Summary` title should override, but of course not for every child within a `Group`..

lassopicasso:
> Does that mean that setting a `title` on a `Summary` component will have no effect unless the target component also does not have one? 🤔 I would assume the `Summary` title should override, but of course not for every child within a `Group`..

Correct. Hm... I kinda agree with you. It's not intuitive that text resource bindings from another source can override what is set in the summary component. I'll see if I can find an alternative solution to this.

lassopicasso:
What about something like this? 

image

src/layout/Group/__snapshots__/SummaryGroupComponent.test.tsx.snap
```diff
@@ -12,11 +12,13 @@ exports[`SummaryGroupComponent SummaryGroupComponent -- should match snapshot 1`
...
       >
-        Mock group
+        <span>
+          Mock group
+        </span>
       </span>
       <button
         aria-label="Change: Mock group"
-        class="Button-module_button__2ZuB7 i__imported_focusable_0 Button-module_small__l39oh Button-module_quiet__1KlhF Button-module_secondary__R0waJ"
+        class="Button-module_button__2ZuB7 utility-module_focusable__1fBKr Button-module_small__l39oh Button-module_quiet__1KlhF Button-module_secondary__R0waJ"
```
olemartinorg, line 21: I believe this was a downgrade - maybe you should merge from main and update the snapshot again? 🤔

olemartinorg (approved):
Look good! 🥳

src/layout/Group/__snapshots__/SummaryGroupComponent.test.tsx.snap
lassopicasso: Thank you ❤️

PR #1279, opened 2023-06-23
Making list of hidden components reactive, static options for Dropdown

olemartinorg:
## Description
- If you had expressions hiding a component which depended on some other component value to also be set, we could end up in a situation where expressions for `hidden` needed to be run multiple times in order to get a stable result. This could cause some components to appear as hidden (or visible) when their expressions clearly said they shouldn't.
- Adding support for static options to the Dropdown component, which was mysteriously missing.

## Related Issue(s)

- https://altinn.slack.com/archives/C047F8FK2NQ/p1687419433771429

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
  - [x] No changes made
- Support in Altinn Studio
  
- Sprint board
- Labels

### DIFF
--- a/src/layout/Dropdown/DropdownComponent.tsx
+++ b/src/layout/Dropdown/DropdownComponent.tsx
@@ -14,9 +14,18 @@ import type { PropsFromGenericComponent } from 'src/layout';
 export type IDropdownProps = PropsFromGenericComponent<'Dropdown'>;
 
 export function DropdownComponent({ node, formData, handleDataChange, isValid, overrideDisplay }: IDropdownProps) {
-  const { optionsId, preselectedOptionIndex, id, readOnly, mapping, source, textResourceBindings } = node.item;
+  const {
+    optionsId,
+    options: staticOptions,
+    preselectedOptionIndex,
+    id,
+    readOnly,
+    mapping,
+    source,
+    textResourceBindings,
+  } = node.item;
   const { langAsString } = useLanguage();
-  const options = useGetOptions({ optionsId, mapping, source })?.filter(duplicateOptionFilter);
+  const options = useGetOptions({ optionsId, mapping, source })?.filter(duplicateOptionFilter) || staticOptions;
   const lookupKey = optionsId && getOptionLookupKey({ id: optionsId, mapping });
   const fetchingOptions = useAppSelector((state) => lookupKey && state.optionState.options[lookupKey]?.loading);
   const hasSelectedInitial = React.useRef(false);
--- a/src/layout/Dropdown/DropdownComponent.tsx
+++ b/src/layout/Dropdown/DropdownComponent.tsx
@@ -25,7 +25,9 @@ export function DropdownComponent({ node, formData, handleDataChange, isValid, o
     textResourceBindings,
   } = node.item;
   const { langAsString } = useLanguage();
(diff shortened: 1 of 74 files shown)


src/layout/Dropdown/DropdownComponent.tsx
```diff
@@ -14,9 +14,18 @@ import type { PropsFromGenericComponent } from 'src/layout';
...
+    preselectedOptionIndex,
+    id,
+    readOnly,
+    mapping,
+    source,
+    textResourceBindings,
+  } = node.item;
   const { langAsString } = useLanguage();
-  const options = useGetOptions({ optionsId, mapping, source })?.filter(duplicateOptionFilter);
+  const options = useGetOptions({ optionsId, mapping, source })?.filter(duplicateOptionFilter) || staticOptions;
```
bjosttveit, line 28: FYI, the Select component from the design system throws an exception if there are duplicated option values, which is why duplicates are filtered out from `useGetOptions`.

src/utils/conditionalRendering.ts
```diff
@@ -51,8 +50,7 @@ export function runConditionalRenderingRules(
         });
 
         if (connection.repeatingGroup.childGroupId) {
-          const childGroup: IRepeatingGroup | undefined =
-            repeatingGroups && repeatingGroups[`${connection.repeatingGroup.childGroupId}-${index}`];
+          const childGroup = repeatingGroups && repeatingGroups[`${connection.repeatingGroup.childGroupId}-${index}`];
           if (childGroup) {
             for (let childIndex = 0; childIndex <= childGroup?.index; childIndex++) {
               const connectionNestedCopy: IConditionalRenderingRule = JSON.parse(JSON.stringify(connectionCopy));
```
bjosttveit, line 56: Is there a reason not to switch to `structuredClone` here as well?

bjosttveit (approved):
Nice work!

src/layout/Dropdown/DropdownComponent.tsx
olemartinorg: Oh.. 🤔 But I guess that's fair? 🤷 Errors in the input should not be silenced, but should throw errors, IMO - although throwing exceptions is not that developer-friendly, I guess.

If I was developing an app with 4 options and only saw 3 of them, I could end up spending way too much time trying to figure out why one of the options 'disappeared', instead of immediately noticing that there are duplicate values (if I just saw the exception with an explanation in the console).

src/utils/conditionalRendering.ts
olemartinorg: Nah, don't think I touched that code - but fixing that now!

src/layout/Dropdown/DropdownComponent.tsx
bjosttveit: It might be a breaking change though

olemartinorg: Right - fixing it properly in `useGetOptions()` now, in such a way that others can't easily do the same mistake.

olemartinorg:
The failing `multipart-save` cypress test seems to be a fluke, from what I can tell. I got it to fail once myself, but now it's running fine every time as far as I can tell.

Issue #1022, opened 2023-03-22
Language selector options are not visible


lasseklovstad:
### Description of the bug

The language selector cuts off option-text, so its impossible to see which language you are selecting. 

### Steps To Reproduce

1. Select a language where the option has not so many characters
2. open the selector again and see that the other options are cut off

### Additional Information
When you choose a option with not so long text.
![image](https://user-images.githubusercontent.com/43168787/226838453-f2bb30ca-ee49-439f-aec9-78f9e92c9ba1.png)

When you choose a longer option, it is okei:
![image](https://user-images.githubusercontent.com/43168787/226844109-c3d70ee1-ed71-4e71-b8d4-966c64d714bd.png)

Issue #832, opened 2023-01-11
Auto-genereate layout schema based on TypeScript interfaces


framitdavid:
### Description

When we are creating new components within app-frontend-react we need to manually add the component interface to the layout schema. This task is time-consuming and it may be errors/mismatches between the schema and the component interface. This task should be automated to avoid the mentioned drawbacks. 

I'm not sure what is the best approach to generate the layout schema but I have listed 3 alternatives below:

1. Generate schema based on TypeScript Interfaces for each component
2. Generate TS Interface based on layout schema
3. Have a third-part format to define the interface and then convert it to schema and typescript.

nkylstad:
Lots of possibilities here! I very much like the idea of 3, though at least in the short term 1. might be the most practical considering that the interfaces are already easily available and very structured. 

Wrt developer experience, 3 might also be more constraining in that there is an extra step that needs to be done when developing a new component. Typically the component is first developed and tested in app frontend, and then added to the jsonschema, so starting in that end and going with 1 might also be the best solution wrt that. But @olemartinorg will have to weigh in here.

One important aspect we have to consider is that the structure of the current `layout.v1.schema.json` file is not optimal. F.ex. a lot of components are defined with properties from the component base that are not relevant. Fex. `dataModelBindings` are expected for components like `Header` and `Paragraph`. 
If we autogenerate the JSON schema from the TS interfaces, it would probably solve that for us.

olemartinorg:
I also gravitate towards 3 here, also because I'd prefer to do this once instead of having a stop-gap solution before that (we have that temporary solution already, by manually trying to keep the files in sync). This is a tooling and developer experience issue, so we shouldn't let that steal focus from delivering functionality to users (which is why I recommend doing this once).

With that said, the immediate problem I see with number 1 is that it might be hard to parse typescript definitions. The types disappear at runtime, so then we'd have to shop around for a typescript parser (for what i perceive is just a temporary solution).

[...]

PR #1283, opened 2023-06-26
Separate JSON schema for each component

nkylstad:
## Description

- Split the large `layout.schema.v1.json` into separate schemas for each component, which are referenced from a new `layout.schema.v2.json`
- Definitions that are common/re-used in many components are added to the `common-defs.schema.json` file, which only contains definitions.
- Basically all the components have the properties `showAsSummary`, `hidden`, `grid` and `pageBreak`, as I'm not sure if there are any components that do not support these. Feel free to comment/suggest changes here.
- I have tried to keep f.ex. `dataModelBindings` and `textResourceBindings` as component specific as possible, but it is possible I have missed/forgotten something - again feel free to update/comment.
- I added a script for generating the base for the component schema and adding reference to this to the new `layout.schema.v2.json`
- One issue I encountered was the "either/or" nature of the `options/optionsId` properties for selection components. I'm sure it's possible to create a conditional schema where:
  - it's not allowed with both properties together
  -  `options` is required when `optionsId` is not present
  - `optionsId` is required when `options` is not present
  However I was not able to get this working. As it's not a feature of the current schema either I left it out for now.

I'm also slightly unsure of whether the relative paths to the linked schemas work in the case of schemas in a different folder. It works when I test it locally, but I imagine it might be different once this is published to CDN - we might have to use complete URLs for schemas that do not live in the same folder 🤔 

I will also need some help updating any pipelines/actions that publish schemas to CDN if there is a need for that.

## Related Issue(s)

- these changes are done as a part of https://github.com/Altinn/altinn-studio/issues/9529
- created a separate issue just for this: https://github.com/Altinn/altinn-studio/issues/10603

## Verification/QA

  - [x] I have tested these changes manually (i have tested creating a layout and seeing that it validates as expected)
- Automated tests
  - [ ] Has been added/updated
  
  - [x] No changes made
- Support in Altinn Studio
  
- Sprint board
- Labels


olemartinorg:
>I'm also slightly unsure of whether the relative paths to the linked schemas work in the case of schemas in a different folder. It works when I test it locally, but I imagine it might be different once this is published to CDN - we might have to use complete URLs for schemas that do not live in the same folder 🤔

Should be easy enough to test locally with `http-server`. I'll look into it!

> I will also need some help updating any pipelines/actions that publish schemas to CDN if there is a need for that.

Should not be needed - the whole `schemas` folder is copied over.

I'll add a list of TODOs I think we should look into after this:
- [ ] Test references to external schemas inside a folder to make sure it works when deploying to CDN
- [ ] Run tests to check existing apps, errors and differences between this schema and the old schema. (See `src/utils/layout/schema.test.ts`)
- [ ] Move this over to `v1` if it's not a breaking change (if it is, work out a strategy for migrating apps to `v2`)
- [ ] Add automatic formatting with prettier? We disabled this for the older schema because we didn't want to reformat that file when adding/changing a few lines. With a clean slate, we can clean it up as we go.
- [ ] #832 🙏

bjosttveit:
> I have tried to keep f.ex. dataModelBindings and textResourceBindings as component specific as possible, but it is possible I have missed/forgotten something - again feel free to update/comment.

Quickly looked over a few components and there are likely many missing textResourceBinding properties. Not sure if this is up-to-date anymore but: https://altinndevops.slack.com/archives/C045EB3JA9X/p1675931223052249?thread_ts=1675871625.344949&cid=C045EB3JA9X

bjosttveit:
> @bjosttveit I think we should merge this as-is, to let @nkylstad carry on - and in a separate job we should look over our text resource bindings for each component and properly type them in both TypeScript and in this schema. Would be a nice way to start off #832, maybe.

Then maybe switch `additionalProperties` to `true` in textResourceBindings for now? Or people will get a ton of warnings and it would be confusing. Some components like `Link` require the target to be set in textResourceBindings.

nkylstad:
> > @bjosttveit I think we should merge this as-is, to let @nkylstad carry on - and in a separate job we should look over our text resource bindings for each component and properly type them in both TypeScript and in this schema. Would be a nice way to start off #832, maybe.
> 
> Then maybe switch `additionalProperties` to `true` in textResourceBindings for now? Or people will get a ton of warnings and it would be confusing. Some components like `Link` require the target to be set in textResourceBindings.

Sure, I have set `additionalProperties: true` for all textResourceBindings now. Then you can restrict it back for the relevant components when you're sure which keys are the correct ones 👍

schemas/json/component/common-defs.schema.v1.json
```diff
@@ -0,0 +1,607 @@
...
+        "multiPage": {
+          "title": "Multi-page",
+          "description": "Boolean value indicating if form components in edit mode should be shown over multiple pages/views.",
+          "type": "boolean"
+        },
+        "addButton": {
+          "title": "Add button",
+          "description": "Boolean or expression indicating whether add new button should be shown or not under the table.",
+          "$ref": "expression.schema.v1.json#/definitions/boolean"
+        },
```
bjosttveit, line 438: You forgot to add `../layout/` in front of the reference to `expression.schema.v1.json#/definitions/boolean` here, so it cannot find the reference.
image

schemas/json/component/list-component.schema.v1.json
```diff
@@ -0,0 +1,115 @@
+{
+  "$id": "https://altinncdn.no/schemas/json/component/list-component.schema.v1.json",
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "title": "List component",
```
bjosttveit, line 4: `List` seems to be missing `dataModelBindings`. In this case, there really are no fixed binding keys, it is up to the developer to select which columns to store:
https://docs.altinn.studio/app/development/ux/components/listcomponent/#how-to-store-data-in-the-data-model

FinnurO:
Har man glemt denne settingen som kom nylig på Radiogroup? https://github.com/orgs/Altinn/projects/39/views/2?pane=issue&itemId=30673603

nkylstad:
> Har man glemt denne settingen som kom nylig på Radiogroup? https://github.com/orgs/Altinn/projects/39/views/2?pane=issue&itemId=30673603

Ja! Glemt er kanskje å ta litt hardt i gitt at PR'en for den radio-endringen ble merget i går 😅 Men tar det med!

schemas/json/component/list-component.schema.v1.json
nkylstad: Added `dataModelBindings` for `List`, with no set property names.

schemas/json/component/common-defs.schema.v1.json
nkylstad: Whoops 🙈 Updated now.

FinnurO:
Vi har vel også en "PrintButton"? Brukes i anonym oppstartsveileder.

olemartinorg:
Marking this as draft because I'm working on adding support for the 3 missing components here. After that is done, I want to compare the results of validating against this file and the old one on existing apps to make sure we know of every case where this might break an existing configuration.

thestokkan:
Hi! I've created Hugo shortcode scripts to automatically list each component's properties in the documentation and have come across a few issues:
* [ActionButton](https://github.com/Altinn/app-frontend-react/blob/3ad9e86634a9f942fcdefc0a5d15053b710a3c1a/schemas/json/component/ActionButton.schema.v1.json#L34): examples should be quoted separately (i.e., `"sign", "confirm", "reject"`, not `"sign, confirm, reject"`).
* [Custom](https://github.com/Altinn/app-frontend-react/blob/3ad9e86634a9f942fcdefc0a5d15053b710a3c1a/schemas/json/component/Custom.schema.v1.json#L28): typo `daataModelBindings`.
* [gridCellComponent](https://github.com/Altinn/app-frontend-react/blob/3ad9e86634a9f942fcdefc0a5d15053b710a3c1a/schemas/json/component/common-defs.schema.v1.json#L374): Should the referenced [tableColumnOptions](https://github.com/Altinn/app-frontend-react/blob/3ad9e86634a9f942fcdefc0a5d15053b710a3c1a/schemas/json/component/common-defs.schema.v1.json#L382) be part of the above properties map?

thestokkan:
@olemartinorg Aha 😅 I was aware of someone working on autogenerating the schema files. I asked @nkylstad before the holiday if anyone was working on autogenerating documentation from the schema files and she said no. 🤷🏻‍♀️  I just merged a PR with the shortcode files (#1022) and a separate PR with drafts for template files and a component example using the shortcode (#1021) (I'm going to ask for feedback on the template on Slack). Why don't you have a look and then we figure out the best solution? I'm a Hugo newbie so you can most likely create something more robust. 😉

olemartinorg:
@thestokkan Very interesting! 🤔🤩 I still have some work left on #832, but I would love to work together with you and @nkylstad to make something work. I believe JsonSchema is just a bit too complicated to parse for this purpose, so I think we can generate a new format that can be used to automatically create markdown for component properties docs, and also for generating a configuration Altinn Studio can use to get some hints at what is needed for the component configuration. If that spins out into 4 different formats (our internal TypeScript types, the JsonSchema used for validation + vscode hinting, a simplified structure for documentation, and another format for Studio) that should be easily possible with the code generator I'm making.

So, I suggest we work together on this one! 🙏 I'll continue working on #832, but I'll make sure to reach out when that's more stable.

thestokkan:
@olemartinorg Sounds great! 
And I can attest to json schema being difficult to parse! 😆 I have managed to do it, but the presentation of the components is probably not the best.

Question is, should I keep the shortcode for now until we have something else? I've used the json schema links from this PR waiting for it to be merged. Not sure if that is ok as a temporary solution?

olemartinorg:
@thestokkan I suspect the new JsonSchema will look radically different, so even though parsing _this_ JsonSchema was possible, parsing every possible valid varation in JsonSchema is close to impossible - so using the JsonSchema itself for parsing to docs makes for a solution that will easily break, and/or limit our creativity when extending the schema.. The same thing goes for using the schema as a basis for an editor (i.e. Studio).

So yeah, keep it as-is, if you have other things to work on, and we'll touch base to figure out a longer-lasting solution. 🙌 Btw, I looked briefly at the hugo shortcode, and it did not look like newbie code to me! 🤷 Nifty solution! 🤩

olemartinorg:
The new (generated) JsonSchema is on its way in #1406. Closing this PR, and we'll figure out a solution to the docs and Studio side of things discussed above.

Issue #1230, opened 2023-06-06
Grid component: Helptext in text cells


olemartinorg:
### Description

There has been multiple requests for helptext support in Grid cells that show text. Theoretically simple to implement, but this isn't a normal FieldSet (because we don't know which field(s) the text cell refers to).

### Additional Information

- https://altinn.slack.com/archives/C02FK32KVKJ/p1685532531583239

RonnyB71:
Alternatives:
1. Help text on text cell, no binding against input filed and text needs to be duplicated on input field for screen readers.
2. Property on text cell that holds a reference to the input field and uses the text and help text from that component

PR #1286, opened 2023-06-27
Text cell with help text in grid

lassopicasso:
## Description

Two new configuration options have been introduced for grid cells:
1. In addition to the `text` prop, the current text cell now also supports the `help` prop, which allows to include help text along with the `text` prop. 
2. A new type of cell has been added, which functions similarly to the text cell, but instead refers to text resources from a component. It uses the property `labelFrom` with the corresponding component ID. In the referenced component, you can access the `title` (which serves as the label for the component), `description`, and `help` to provide help text.

Example 1:
```
            "cells": [
              { "text": "laan-bolig", "help": "Dette er en hjelpetekst til laan-bolig" },
              ...
            ]
```

Example 2:
```
            "cells": [
              { "labelFrom": "<component-id>"},
              ...
            ]
```

## Related Issue(s)

- closes #https://github.com/Altinn/app-frontend-react/issues/1230

## Verification/QA

- Automated tests
  - [x] Has been added/updated
https://github.com/Altinn/altinn-studio-docs/pull/1011
  
  - [ ] No changes made
- Support in Altinn Studio
  
- Sprint board
- Labels

### DIFF
--- a/src/layout/Grid/GridComponent.tsx
+++ b/src/layout/Grid/GridComponent.tsx
@@ -6,8 +6,10 @@ import { Grid } from '@material-ui/core';
 import cn from 'classnames';
 
 import { ConditionalWrapper } from 'src/components/ConditionalWrapper';
+import { Description } from 'src/components/form/Description';
 import { FullWidthWrapper } from 'src/components/form/FullWidthWrapper';
 import { HelpTextContainer } from 'src/components/form/HelpTextContainer';
+import { Label } from 'src/components/form/Label';
 import { useIsMobile } from 'src/hooks/useIsMobile';
 import { useLanguage } from 'src/hooks/useLanguage';
 import { GenericComponent } from 'src/layout/GenericComponent';
@@ -45,6 +47,7 @@ export function RenderGrid(props: PropsFromGenericComponent<'Grid'>) {
             row={row}
             isNested={isNested}
             mutableColumnSettings={columnSettings}
+            node={node}
           />
         ))}
       </Table>
@@ -56,9 +59,10 @@ interface GridRowProps {
   row: GridRow<GridComponent>;
   isNested: boolean;
   mutableColumnSettings: ITableColumnFormatting;
+  node: LayoutNode;
 }
 
-export function GridRowRenderer({ row, isNested, mutableColumnSettings }: GridRowProps) {
+export function GridRowRenderer({ row, isNested, mutableColumnSettings, node }: GridRowProps) {
   const { lang } = useLanguage();
 
   return isGridRowHidden(row) ? null : (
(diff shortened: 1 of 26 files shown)


olemartinorg:
Hmm. A few comments on this one:
1. I don't think we need the logic for `aria-describedby`. Components in Grid are described by their own text resources, as they are required for [mobile support](https://docs.altinn.studio/app/development/ux/components/grid/#mobile-support). Meaning, the components in the table have everything they need for accessibility already. Also the description should come from the `description` text resource binding - help text is something else.
2. We can't make the assumption that a text cell describes all components in the same row. It's entirely valid to have a row that is `[text cell, component, text cell, component]`, and so making assumptions that the first text cell describes the last component is most likely false.
3. The thing I remember we discussed was to avoid duplicating the help text by referencing a component (the other way around from what is done here). Meaning that you could for example implement a row like:

```json
{
  "cells": [
    { "textFrom": "myComponent" },
    { "component": "myComponent" }
  ]
}
```

And such a configuration would fetch the texts (i.e. `title`, `help` and possibly also `description`) from the target component and put them in the text cell. That way you get most of the same effects of the `aria-describedby` feature added here, but also:

- No need to add the text resource bindings twice (both on the component and in the text cell)
- We can implement functionality to make the title/description/label to work as a label (i.e. clicking on it focuses the target field)

lassopicasso:
Understood, thanks for the feedback @olemartinorg ! 👍🙏

src/layout/Grid/GridComponent.tsx
```diff
@@ -76,29 +82,48 @@ export function GridRowRenderer({ row, isNested, mutableColumnSettings }: GridRo
           mutableColumnSettings[cellIdx] = cell.columnOptions;
         }
 
-        if (cell && 'text' in cell && cell.text) {
+        const closestComponent = () => {
+          if (cell && 'labelFrom' in cell && cell?.labelFrom) {
+            const idToFind = cell.labelFrom;
+            return node.flat(true).find((n) => n.item.id === idToFind || n.item.baseComponentId === idToFind);
+          }
+        };
```
olemartinorg, line 90: Nitpicking here, but this code could probably just be moved down into the other `if` below, and it has no need to be a function, so you could just pass the found node to `<CellWithLabel>` directly.

src/layout/Grid/GridComponent.tsx
```diff
@@ -131,6 +156,8 @@ function InternalRow({ header, readOnly, children }: InternalRowProps) {
 interface CellProps {
   className?: string;
   columnStyleOptions?: ITableColumnProperties;
+  referenceComponent?: LayoutNode;
+  help?: string;
 }
```
olemartinorg, line 161: Could probably just make another type for `<CellWithLabel>`. Mixing these two cell types could just be confusing.

olemartinorg (approved):
This looks great! 🚀 Some very minor and optional nitpicks in the comments.

Issue #1023, opened 2023-03-22
role: alert when info-panel is used with dynamics


Febakke:
### Description of the bug
We have some situations today where apps use our panel component in combination with dynamics to create an alert component that triggers based on user input. This gives us an accessebility issue where this panel should have an aria alert role so that screen readers also can get this information.

## conundrum
We cant just add an alert role on our panels because: 

> The most important thing to know about the alert role is that it's for content that is dynamically displayed, not for content that appears on page load.
[Source ](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/alert_role)

One of the main uses for the panel component is to highlight information on a page, and is often used so that is visible on page load. We need to seperate dynamic alerts and those that is only used to highlight static information.

Also, this is pretty similar to our soft-validation functionality. If we expand the functionality of our panels to handle this, we have to seperate components doing almost the same thing. This might be hard to comunicate the difference between the 2. 

One solution is to in some way expand the soft validation so that it meets our appdevelopers needs. Then we can contain this to one consept that is easier to communicate. The other is to add some attribute to the panel that can be set by the app-developer when they use it with dynamics, but then we add alert functionality to a component that maybe should be used as a way of presenting static information, and by having another attribute the app-developer has to worry about, we increase the chance of accessebility issues in apps that use this feature. 

Or maybe a developer have a better suggestion on how to solve this? 😇

Small example: 
![image](https://user-images.githubusercontent.com/33222679/226847850-41460207-0e00-4e74-a730-8816e0c60f57.png)

[aria alert documentation](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/alert_role)

## Specification
The documentation specifies that you need to be careful with the use of alert as it can be very intrusive for screen reader users. Is there a risk that it might be overused, should the developer be able to turn it of if the information is not that important. 
- [ ] Look at relevant cases
- [ ] Define guidelines
- [ ] Added to all variants (info, success and warning)
- [ ] Consider adding the alert that is specified in the designsystem [Issue](https://github.com/digdir/designsystem/issues/78)

### Steps To Reproduce

Go to an app with soft validation and test if it is read by the screen reader when it is triggered.

StianVestli:
org/ssb

StianVestli:
Regner med at hvis man legger på aria-live på denne så vil dette være på alle varianter av denne boksen. dvs alle disse:
info
success
warning

[...]

PR #1289, opened 2023-06-27
feat(Alert): add Alert component which uses @digdir/design-system-react

### LINKED ISSUE #1284: Be able to use the Alert component from Felles Designsystem
### Description

We want to be able to present information in the `Alert`-component from `@digdir/desgin-system-react`. 

## Proposed technical solution
This needs to be solved with a new component which implements the `Alert` from `@digdir/design-system-react`. 

The new component should: 
- Have an option for setting `severity`.  
- Have an option for setting the attribute `role="alert"` on the component. This is used for cases where alerts appears as a consequence of user input. Screen readers will then be able to relay the text in the alert to its users.
- Have an optional `title` attribute
- Have an optional `description` attribute
- ~~Possibly have an option for turning on and off `elevated` property (adding a box-shadow).~~

### Additional Information

Screenshot of the design we want to implement:
image

[The alert-component in the design-system](https://digdir.github.io/?path=/docs/kjernekomponenter-alert--docs)

mikaelrss:
## Description

Added a new `Alert` component, with options for 
- `severity` - sets the style of the alert, and has four valid values `"success" | "info" | "danger" | "warning"`.  
- `title` - sets the title via `textResourceBindings`.
- `description` - sets the description via `textResourceBindings`.
- ~~`useAsAlert` - A boolean which specifies whether the component should have `role="alert"` or not. It should not have this for static elements displaying information, but should have it for dynamic elements that appear after a certain user input. I currently implemented the `aria-live` attribute with `polite`, but I think maybe this should be configurable. Or maybe it should be based on severity label. if `useAsAlert` is set, and `severity` is >= `warning`, `aria-live` should be set to `assertive`.~~ This option is removed and rather determined automatically based on the `hidden` property of `GenericComponent`. If `hidden` is explicitly set to `false` we append `role="alert"` to the `Alert` component. There is however one problem with this. If an element is hidden with an expression, but the expression evaluates to `false` on page load, `role="alert"` will be appended to the `Alert` component on load. The same problem would exist for the old API, where the app developers used a `useAsAlert` flag. 

Screenshot: 
image

## Related Issue(s)

- closes #1284 

## Verification/QA

- Automated tests
  - [x] Cypress E2E test(s) have been added/updated - I have added some alerts to frontend-tests so that the alert is tested with percy. I also updated some failing cypress tests for Soft validations. Stopped using the custom id-based selector, and used `@testing-library/cypress` selectors for testing the content of alert messages instead. 
  - [x] I have tested with a screen reader/keyboard navigation/automated wcag validator - Tested with wave and orca screen reader. The screen reader reads alerts twice for some reason, but I think this is an orca issue. Could somebody test with VoiceOver on Mac?
  - [x] Has been added/updated - [Have submitted a PR with documentation here](https://github.com/Altinn/altinn-studio-docs/pull/1008)
  
  - [ ] No changes made
- Support in Altinn Studio
  
- Sprint board
- Labels

### DIFF
--- a/src/layout/Alert/types.d.ts
+++ b/src/layout/Alert/types.d.ts
@@ -1,7 +1,9 @@
 import type { ILayoutCompBase } from 'src/layout/layout';
 
+export type AlertSeverity = 'success' | 'warning' | 'danger' | 'info';
+
 export interface ILayoutCompAlertBase {
-  severity: 'success' | 'warning' | 'danger' | 'info';
+  severity: AlertSeverity;
   useAsAlert?: boolean;
 }
 
--- a/src/layout/Alert/types.d.ts
+++ b/src/layout/Alert/types.d.ts
@@ -4,7 +4,6 @@ export type AlertSeverity = 'success' | 'warning' | 'danger' | 'info';
 
 export interface ILayoutCompAlertBase {
   severity: AlertSeverity;
-  useAsAlert?: boolean;
 }
 
 export type ILayoutCompAlert = ILayoutCompBase<'Alert'> & ILayoutCompAlertBase;
--- a/src/layout/Alert/types.d.ts
+++ b/src/layout/Alert/types.d.ts
@@ -1,0 +1,8 @@
+import type { ILayoutCompBase } from 'src/layout/layout';
+
+export interface ILayoutCompAlertBase {
+  severity: 'success' | 'warning' | 'danger' | 'info';
+  useAsAlert?: boolean;
+}
+
+export type ILayoutCompAlert = ILayoutCompBase<'Alert'> & ILayoutCompAlertBase;
--- a/schemas/json/layout/layout.schema.v1.json
+++ b/schemas/json/layout/layout.schema.v1.json
@@ -167,12 +167,6 @@
           "description": "The severity of the alert.",
           "type": "string",
           "enum": ["success", "warning", "danger", "info"]
-        },
-        "useAsAlert": {
-          "title": "Use as alert",
-          "description": "Boolean value indicating if the alert should be used as an alert. Setting to true will make screen readers read the alert when it is rendered.",
-          "type": "boolean",
-          "default": false
         }
       },
       "required": ["severity"]
--- a/schemas/json/layout/layout.schema.v1.json
+++ b/schemas/json/layout/layout.schema.v1.json
@@ -42,7 +42,7 @@
           "type": "string",
           "title": "Type",
           "description": "The component type.",
-          "enum": ["ActionButton", "AddressComponent", "AttachmentList", "Button", "ButtonGroup", "Checkboxes", "Custom", "Datepicker", "Dropdown", "FileUpload", "FileUploadWithTag", "Grid", "Group", "Header", "Image", "Input", "InstanceInformation", "InstantiationButton", "IFrame", "Likert","Link", "List", "MultipleSelect", "NavigationButtons", "NavigationBar", "Panel", "Paragraph", "PrintButton", "RadioButtons", "Summary", "TextArea"]
+          "enum": ["Alert","ActionButton", "AddressComponent", "AttachmentList", "Button", "ButtonGroup", "Checkboxes", "Custom", "Datepicker", "Dropdown", "FileUpload", "FileUploadWithTag", "Grid", "Group", "Header", "Image", "Input", "InstanceInformation", "InstantiationButton", "IFrame", "Likert","Link", "List", "MultipleSelect", "NavigationButtons", "NavigationBar", "Panel", "Paragraph", "PrintButton", "RadioButtons", "Summary", "TextArea"]
         },
         "required": {
           "title": "Required",
@@ -128,6 +128,7 @@
       },
       "required": ["id", "type"],
       "allOf": [
+        { "if": {"properties": {"type": { "const": "Alert"}}}, "then": {"$ref": "#/definitions/alertComponent"}},
         { "if": {"properties": {"type": { "const": "ActionButton"}}}, "then": {"$ref": "#/definitions/actionButtonComponent"}},
         { "if": {"properties": {"type": { "const": "AddressComponent"}}}, "then": { "$ref": "#/definitions/addressComponent"}},
         { "if": {"properties": {"type": { "const": "AttachmentList"}}}, "then": { "$ref": "#/definitions/attachmentListComponent"}},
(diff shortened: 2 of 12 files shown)


mikaelrss:
I see that I might need to update this PR and split out the schema for `AlertComponent` into a separate schema once #1283 is merged.

src/layout/Alert/types.d.ts
```diff
@@ -0,0 +1,10 @@
+import type { ILayoutCompBase } from 'src/layout/layout';
+
+export type AlertSeverity = 'success' | 'warning' | 'danger' | 'info';
+
+export interface ILayoutCompAlertBase {
+  severity: AlertSeverity;
+  useAsAlert?: boolean;
```
Magnusrm, line 7: I feel this property maybe needs a new name as the components name is `Alert`. 🤔 What about `useAriaLive` or something similar?

Side thought that might not work: Maybe it is possible to check if the Alert components `required` property in the layout file contains a boolean value or an expression(which is written as a string) and set this value to `true` if an expression is used?

olemartinorg: Agree! 💯 But I assume you mean the `hidden` property? An `Alert` is not a form component, so it cannot have any functionality for `required`.

Magnusrm: @olemartinorg, Yes thats what i meant! Sorry 😄

olemartinorg:
This has a striking resemblance to the `Panel` component we already have, although this looks a bit different. Be sure to document what's the difference between these two components, and what's the use-case for each of them.

Also, this reminds me of #1023, which we closed because we wanted to introduce validations using expressions instead via #726. Does this Alert component have to be a component, or the use-case really just the same as [soft validations](https://docs.altinn.studio/app/development/logic/validation/#soft-validations)? In that case, I think we should probably consider updating the visuals for the soft validations instead of making a new component.

src/layout/Alert/types.d.ts
mikaelrss: Yeah that seems like a good idea. I'll see if that is possible :sweat_smile:

mikaelrss:
> This has a striking resemblance to the `Panel` component we already have, although this looks a bit different. Be sure to document what's the difference between these two components, and what's the use-case for each of them.
> 
> Also, this reminds me of #1023, which we closed because we wanted to introduce validations using expressions instead via #726. Does this Alert component have to be a component, or the use-case really just the same as [soft validations](https://docs.altinn.studio/app/development/logic/validation/#soft-validations)? In that case, I think we should probably consider updating the visuals for the soft validations instead of making a new component.

Thanks for the feedback! Regarding #1023 @Febakke was involved in the startup meeting of this new component, and I think we said that this new component should take over the role of Panel in displaying messages which appear as some sort of user input. Please correct me if I am wrong here @Febakke 

This component could also be used to display static messages which do not appear/disappear based on user input. Soft validations seem to me like they only appear as a validation message whenever the user provides some input. 

Maybe we should have a chat about this component again @Febakke @olemartinorg @Magnusrm

olemartinorg:
I see! 🤔 

> This component could also be used to display static messages which do not appear/disappear based on user input.

The `useAsAlert` flag contradicts that. It opens up for using this as something that works just like a validation message, but a validation message that circumvents all the other (current and future) functionality in validation messages. I would assume there comes a feature request at some point for "having the option to not let the user proceed to the next page in the form when a certain `Alert` is displayed" and at that point we're on our way to re-implement validation with new clothes.

> Soft validations seem to me like they only appear as a validation message whenever the user provides some input.

Ah, yes - so `Alert` is, in a sense, permanent soft validations. 🤔 Again, the functionality overlap here makes me feel uneasy. I don't want to be difficult, and I fully understand why you'd want to use a "fully featured component" for these messages, as it's much more flexible than regular validations (and permanent, as you mention!), but I think we should discuss this some more before letting it out in the wild. I would be very confused as an app developer if we promoted 3 different methods that achieve very similar things.

Maybe implementing an 'initial' trigger for validations could be an alternative (worryingly, validation messages disappear right now if you refresh the app, and you won't get them back until you hit a trigger), or reworking the way triggers on validations work could be another option.

Tagging @RonnyB71 as well.

mikaelrss:
> The useAsAlert flag contradicts that. It opens up for using this as something that works just like a validation message, but a validation message that circumvents all the other (current and future) functionality in validation messages.

The thinking is that `Alert` can be used as both information and an alert that grabs the attention of a screen reader. This does not necessarily _have_ to be a validation message, but can certainly be used as one also :sweat_smile:. It could very well just be additional information that is displayed when making a user action. For instance, in our specific use case in DGM we want to display this piece of extra information to the users when they select "Ja, jeg har informert de andre etterlatte", even though this is not a validation message. 

image

> Ah, yes - so Alert is, in a sense, permanent soft validations.

Yes, `Alert` is both soft validations _and_ permanent soft validations, and just information :sweat_smile: 

In any case, it seems we need to discuss this component a little more in detail before we introduce it to the codebase.

Febakke:
@olemartinorg 
I agree that this will be confusing for the app-developer and we have a lot of overlap here. I guess I did not think this trough when we had the initial meeting on this. 🐥 We should land on a plan on how to divide between alert and panel in a way that is easy to understand for our app-developers. Anyway, here are my thoughts on this. This is a suggestion on how it could work in an ideall world 😅

[Alert](https://www.w3.org/WAI/ARIA/apg/patterns/alert/)
Alert is a defined pattern, and we use the same icons and severity grades as [NAV](https://aksel.nav.no/komponenter/core/alert) does in their component. Our goal should be to use alert in the same way inn Apps. It should only contain text, preferbly short texts. It should not be used for validating a form. But in the future it might be used for giving the user information about a technical fault that is not related to what the user has filled in. "We could not submit your form because our database is full, sorry" 

NB: I see that NAV has changed their documentation to say that you should not give the alert component: `role=alert`. Im not sure why, if the alert is presented after the page has been loaded it must have this role for screen readers to catch it. (They do say that it should not be presented dynamicly)

Summary:
* Should be used for important short messages for the user
* Should have different severity levels
* Should not have buttons or other components inside them
* Can be static or presented based on something that has happened either from the user or something outside of their control

**Panel**
When we first made our panel component, we had many different use cases and we thought it should handle all of them and we did at that time not think about concistensy across public solutions like NAV. This has been something that we have started to focus more on after starting to work on the design system.

The panel should be a grouping of elements that make it stand out from the rest of the page. It has no limitations on what it can contain. Example of uses, highlight parts of a paragraph you want to stand out or grouping form components with a text that is relevant. 

The color should only be a stylistic choice, it should not inform the user of a grade of severity.

If it is presented to the user with dynamics it might need a `live-region: polite` depending on the content. 

I have to 🏃, but there is also some overlap with soft validations that I have to think about in this context.

mikaelrss:
After a discussion in todays daily meeting, we have decided that we will introduce the new `Alert` component and at the same time replace the visual design of Soft validations to use the new `Alert` component from `@digdir/design-system-react`

Magnusrm (approved):
Nice work! This looks good to me 🚀

olemartinorg (approved):
Looks good, thank you! 🚀

Issue #165, opened 2021-04-13
Component: Accordion (feature request brreg)


andreas-didriksen:
## Is your feature request related to a problem? Please describe
I want to use accordion for the Basic form and Text components to better structure the form
 layout. In complex forms, it may be advantageous for the user to read the heading of the different sections first. The accordion can make the form appear less daunting and help to reduce scrolling and shorten forms. 

## Describe the solution you'd like
I want an accordion component in Altinn studio designer. After placing the accordion component in the form layout, I want to place text and basic form component inside the accordion component. 

## Describe alternatives you've considered
An alternative is only to have an accordion component for text/paragraph in Altinn studio designer.

![trekkspill](https://user-images.githubusercontent.com/75472603/114628848-dae01980-9cb7-11eb-8250-7ba1959209bc.jpg)

bjosttveit:
Solved by #2957

PR #1301, opened 2023-06-30
Feat/1285 accordion and accordion group

### LINKED ISSUE #1285: Be able to use the Accordion component from Felles Designsystem
### Description

We want to be able to display information with the use of the `Accordion` component from `@digdir/design-system-react`. 

## Proposed techincal solution
~~This needs to be solved as a new component~~ ~~Can be solved by providing an option to `Group` and `RepeatingGroup` which implements the `Accordion` component from `@digdir/design-system-react`.~~

~~The new component behaviour of `Group` and `RepeatingGroup` should:~~
- ~~Be able to accept a list of options to support multiple items in the `Accordion`.~~ 
- ~~Each option should accept a `title` text resource via the `textResourceBindings`-property.~~ 
- ~~Be able to accept dynamic content. Text body should be provided by placing a  `Paragraph` inside the `Accordion`.~~

After discussing this with the apps team, we have decided that we will introduce two new components. `Accordion` and `AccordionGroup`. 

## Accordion
`Accordion` would be a single stand alone component which can have a title and some contents in a body. The contents of the body should be comprised of other Altinn 3 components. If the `Accordion` should contain only text, this should be solved with placing a `Paragraph` inside it. There will be some contraints on what content could go into an `Accordion`. The title of the `Accordion` will be provided via `textResourceBindings`. 

Closed state: 
image

Open state: 
image

## AccordionGroup
`AccordionGroup` should be a component that groups `Accordion` components together. The only content that would be allowed to place inside an `AccordionGroup` is `Accordions`. 

image

### Additional Information

[The accordion component from digdir's design system](https://digdir.github.io/designsystem/?path=/docs/kjernekomponenter-accordion--docs)

mikaelrss:
## Description

Adds two new components, `Accordion` and `AccordionGroup`.

## Related Issue(s)

- closes #1285 

## Verification/QA

- Automated tests
  - [x] Has been added/updated - [Added a PR here](https://github.com/Altinn/altinn-studio-docs/pull/1024)
  
  - [ ] No changes made
- Support in Altinn Studio
  
- Sprint board
- Labels

### DIFF
--- a/schemas/json/layout/layout.schema.v1.json
+++ b/schemas/json/layout/layout.schema.v1.json
@@ -42,7 +42,7 @@
           "type": "string",
           "title": "Type",
           "description": "The component type.",
-          "enum": ["Alert","ActionButton", "AddressComponent", "AttachmentList", "Button", "ButtonGroup", "Checkboxes", "Custom", "Datepicker", "Dropdown", "FileUpload", "FileUploadWithTag", "Grid", "Group", "Header", "Image", "Input", "InstanceInformation", "InstantiationButton", "IFrame", "Likert","Link", "List", "MultipleSelect", "NavigationButtons", "NavigationBar", "Panel", "Paragraph", "PrintButton", "RadioButtons", "Summary", "TextArea"]
+          "enum": ["Accordion", "AccordionGroup","ActionButton", "AddressComponent", "Alert", "AttachmentList", "Button", "ButtonGroup", "Checkboxes", "Custom", "Datepicker", "Dropdown", "FileUpload", "FileUploadWithTag", "Grid", "Group", "Header", "Image", "Input", "InstanceInformation", "InstantiationButton", "IFrame", "Likert","Link", "List", "MultipleSelect", "NavigationButtons", "NavigationBar", "Panel", "Paragraph", "PrintButton", "RadioButtons", "Summary", "TextArea"]
         },
         "required": {
           "title": "Required",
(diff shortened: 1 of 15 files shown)


bjosttveit:
Does this solve https://github.com/Altinn/app-frontend-react/issues/165?

mikaelrss:
I guess it could solve #165. As of right now, the `Accordion` only supports `Paragraph` and `Button` components as children, and it seems like the issue needs form components to be placed inside an `Accordion`. I guess we could expand to accompany for more components.

schemas/json/layout/layout.schema.v1.json
```diff
@@ -160,6 +162,34 @@
...
+        }
+      },
+      "required": ["children"]
+    },
+    "accordionGroupComponent": {
+      "properties": {
+        "children": {
+          "title": "Children",
+          "type": "array",
+          "description": "An array of the \"id\" of child components belonging to the button group.",
```
olemartinorg, line 184: You should probably update these descriptions 😉

src/layout/AccordionGroup/hierarchy.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  /**
+   * Check if a component can be rendered in an AccordionGroup.
+   */
+  private canRenderInAccordionGroup(generator: HierarchyGenerator, childId: string, outputWarning = true): boolean {
+    const prototype = generator.prototype(childId);
+    const def = prototype && generator.getLayoutComponentObject(prototype.type);
+
+    if (outputWarning && prototype && !def?.canRenderInAccordionGroup()) {
+      console.warn(
```
olemartinorg, line 61: You should probably use `window.logWarn` instead, to make this appear in our DevTools log. `console.log` etc is for temporary debugging messages to _us_, and should never be committed, but messages to application developers should be sent to their DevTools.

olemartinorg (approved):

mikaelrss:
> But right now there's probably no coverage over your hiearchy generator etc, so maybe a full integration test for these components are in order? It might be difficult to do with pure unit tests, but easy to observe the effects of it in a Cypress test.

I have already added one cypress test testing that the accordion opens and closes, and that the content is invisible/visible. This should cover `Accordion/hierarchy.ts` right? It also covers `AccordoinGroup/hierarchy.ts` since `frontend-test` includes an accordion group as well. However, I haven't made any assertions to specifically test `AccordionGroup`, as this is mainly a style change for `Accordion`s.  

@olemartinorg Do you think this is enough coverage, or is there something I am missing?

olemartinorg:
@mikaelrss Sorry for missing the test! 🙈 Yup, sounds like more than enough tests, then. 🙏 

@bjosttveit Ah! I changed something in `frontend-test` a while ago that might affect that. In the text resource, we didn't escape the `2. ` so markdown formatting thought it a numbered list item. That might explain parts of it, I think. It seems to me that Percy lags behind a while after there has been a change, so I guess we should just expect there to be large diffs a while after a large change... :-/

Issue #1169, opened 2023-05-16
Option in select does not get bold if there are no description


RonnyB71:
### Description of the bug

When having option lists with and without description in the same list - all values should be bold, not just the ones that has a description. When having large lists of values it's easy to miss the ones without a description since it's not bold as the rest of the values.

### Steps To Reproduce

1. Create a codelist with a two options
2. Add description to one of them
3. Use the code list in a dropdown field
4. Note that the one with a description the option gets bold, while the other without a description is not.

### Additional Information

![image](https://github.com/Altinn/app-frontend-react/assets/5430430/8f6a1f37-211c-45a3-b052-d870dfc3e9fe)

bjosttveit:
What about when there are only options with no description? Should they all still be bold?

hannefinnoy:
@mrosvik Do you have any thoughts about this?

mrosvik:
Suggestion:

If no options has description:
- Regular text an all options

If some options has description: 
- Semibold on title, regular on description
- Semibold on options without description

PR #1304, opened 2023-07-04
Option select styling with and without description

lassopicasso:
## Description
To improve the visual appearance, the font-weight is adjusted based on whether the option list contains descriptions in any of the options. If some or all options have descriptions, the labels/titles will be styled with semibold font-weight. If no options have descriptions, the labels/titles will be styled with regular font-weight.

I wasn't sure about the specific numeric value for semibold, but according to [the documentation here](https://drafts.csswg.org/css-fonts/#font-weight-numeric-values) it is recommended to use 600. Let me know if this value is wrong to use. 

## Related Issue(s)

- closes #https://github.com/Altinn/app-frontend-react/issues/1169

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
  - [x] No changes made
- Support in Altinn Studio
  
- Sprint board
- Labels

### DIFF
--- a/src/utils/options.ts
+++ b/src/utils/options.ts
@@ -1,4 +1,4 @@
-import { getParsedLanguageFromText, replaceTextResourceParams } from 'src/language/sharedLanguage';
+import { replaceTextResourceParams } from 'src/language/sharedLanguage';
 import {
   getBaseGroupDataModelBindingFromKeyWithIndexIndicators,
   getGroupDataModelBinding,
@@ -8,7 +8,6 @@ import {
 } from 'src/utils/databindings';
 import type { IFormData } from 'src/features/formData';
 import type { IOptionResources } from 'src/hooks/useGetOptions';
-import type { IUseLanguage } from 'src/hooks/useLanguage';
 import type { ILayout } from 'src/layout/layout';
 import type { IMapping, IOption, IOptions, IOptionsMetaData, IOptionSource, IRepeatingGroups } from 'src/types';
 import type { IDataSources } from 'src/types/shared';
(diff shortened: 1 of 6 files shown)


src/utils/options.ts
```diff
@@ -217,12 +216,18 @@ export function duplicateOptionFilter(currentOption: IOption, currentIndex: numb
...
+  listHasDescription: boolean,
+): React.ReactNode {
   if (option.description) {
     return getParsedLanguageFromText(
       `<b>${langAsString(option.label ?? option.value)}</b><br><span>${langAsString(option.description)}</span>`,
     );
   } else {
-    return getParsedLanguageFromText(`<span>${langAsString(option.label ?? option.value)}</span>`);
+    return getParsedLanguageFromText(
+      `<span style="${listHasDescription && 'font-weight:600'}">${langAsString(option.label ?? option.value)}</span>`,
```
Magnusrm, line 230: I would prefer using a css class from a css module file instead of using inline styling.
```suggestion
      `<span className={listHasDescription && classes.boldOptionLabel}>${langAsString(option.label ?? option.value)}</span>`,
```

Magnusrm (commented):
Looking good! 🏅 Just a small suggestion below 😊

src/utils/options.ts
Magnusrm: I see now that we are using `getParsedLanguageFromText` on the whole part of html that is the label and description. Would it be possible to use it on just the `option.label` ?? `option.value` part?

Magnusrm: Or maybe use `lang()` instead of `langAsString()`, since `lang()` uses `getParsedLanguageFromText()` already

bjosttveit: I suppose that `getParsedLanguageFromText` is probably not needed at all if you could just return some JSX?

bjosttveit: > Or maybe use lang() instead of langAsString(), since lang() uses getParsedLanguageFromText() already

Well, since langAsString is used inside of getParsedLanguageFromText, it currently does not format the text string itself

lassopicasso: Using a css class from a module file instead of inline styling doesn't seem to work 🤔

bjosttveit: Only if you make it into a react component instead of using the parsing thing

bjosttveit: I should not have made it like that in the fist place 😅

Magnusrm (approved):
LGTM! 🚀

PR #1330, opened 2023-07-13
Fix CSS injection order

### LINKED ISSUE #1000: CSS order prevents overriding design system

bjosttveit:
## Description

This is the only way I could find to fix the issues with CSS specificity, see the original issue for an explanation. This will override the `document.head.appendChild` method to inject `style` elements before any `link` elements. This means we no longer need to use `!important` or unecessarily great specificity to override styles from external libraries like `material-ui` and `@digdir/design-system-react`.

I also cleaned up the use of `!important` where I could.

Since our own css is now higher priority, this also means that the old `tjenester3.css` now more often gets applied over library styles, therefore a lot of cleanup in this file (-4000 lines) and affected components was necessary. This is a step in the right direction of getting rid of the old global css and the old font-icons. Whenever I broke an old component relying on `tjenester3.css`, I replaced them with design system components and nav icons instead of fixing them, so there are some visual updates as well.

## Related Issue(s)

- closes #1000 

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
  - [x] No changes made
- Support in Altinn Studio
  
- Sprint board
- Labels


olemartinorg (approved):
Phenomenal work!! :godmode:

PR #1358, opened 2023-08-03
Prevent crash on save invalid decimal value

### LINKED ISSUE #1354: If you enter - on number fields the application crashes
### Description of the bug

Dersom vi skriver inn "-" på feltet "Sjøtemperatur" (gjelder alle felter som inneholder heltall) krasjer skjemaet. Men å skrive inn minus etter man først har skrevet inn et tall funker. Det gjelder dette skjema: [mat.apps.tt02.altinn.no/mat/lakselus-rapportering/](http://mat.apps.tt02.altinn.no/mat/lakselus-rapportering/)
Feltet er definert slik:
```
{
        "id": "lusetelling-sjoetemperatur",
        "type": "Input",
        "grid": {
          "sm": 4
        },
        "textResourceBindings": {
          "title": "Lakselusrapportering.lusetelling.sjoetemperatur",
          "help": "Lakselusrapportering.lusetelling.sjoetemperatur.hjelpetekst"
        },
        "dataModelBindings": {
          "simpleBinding": "lakselus.lusetelling.sjoetemperatur"
        },
        "required": true,
        "readOnly": false,
        "formatting": {
          "number": {
            "decimalScale": 1,
            "decimalSeparator": ","
          }
        }
      },



```

### Steps To Reproduce

Enter - on Sjøtemperatur field, os lusetelling fields. 

### Additional Information

_No response_

bjosttveit:
## Description

If the form data we save to the backend cannot be parsed into the strong types of the data model class, the save fails with 400 bad request and due to our handling of such errors, the app "crashes" with "unknown error". Since we use `decimal` as our preferred number type when generating the data model classes in studio (even if you select integer it seems), we should probably make sure that our `Input` component with number formatting is "safe" to use with numeric data types in the data model.

This change will check whether the number can be parsed into the `decimal` type before saving the data when using number formatting. The `react-number-format` already restricts what you can type into the box, but there are a few edge cases that this deals with. For example, you need to be able to type the characters `-` and `.` to support negative and floating point values; but these characters by themselves are not valid numbers and will cause a crash.

There is also a size restriction on `decimal` that will also cause a crash if not respected. So this also prevents a crash if you type a number that is too big. ~~A potential issue is that you do not get any feedback that the data you typed was not saved, this cannot easily be done with validation, since that requires the data to be saved first.~~

**EDIT**: I may have found a solution which is not too hacky. I am now forcing a rerender onBlur if the component has number formatting. Before, it would still show the entire number you typed even though it did not get saved, now you at least see the data you are actually going to submit.

https://github.com/Altinn/app-frontend-react/assets/47412359/a484e4e3-1898-4937-b159-a660edaae20e

## Related Issue(s)

- closes #1354 
- somewhat mitigates https://github.com/Altinn/app-lib-dotnet/issues/197 but only for `decimal` values which is default. It will still crash when used with `int`. 

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
  - [x] No changes made
- Support in Altinn Studio
  
- Sprint board
- Labels

### DIFF
--- a/src/hooks/useReload.ts
+++ b/src/hooks/useReload.ts
@@ -1,0 +1,7 @@
+import { useCallback, useState } from 'react';
+
+export function useReload(key: string): [string, () => void] {
+  const [reloadKey, setReloadKey] = useState(0);
+  const reload = useCallback(() => setReloadKey((k) => k + 1), []);
+  return [`${key}-${reloadKey}`, reload];
+}
--- a/src/hooks/useReload.ts
+++ b/src/hooks/useReload.ts
@@ -1,7 +1,12 @@
 import { useCallback, useState } from 'react';
 
-export function useReload(key: string): [string, () => void] {
-  const [reloadKey, setReloadKey] = useState(0);
-  const reload = useCallback(() => setReloadKey((k) => k + 1), []);
-  return [`${key}-${reloadKey}`, reload];
+/**
+ * Hook used to force a rerender of a component. It works by updating the key whenever you call rerender().
+ * @param baseKey - The base key to use.
+ * @returns A tuple containing the key to set in the component's key prop, and the rerender function.
+ */
+export function useRerender(baseKey: string): [string, () => void] {
+  const [number, setNumber] = useState(0);
+  const rerender = useCallback(() => setNumber((k) => k + 1), []);
+  return [`${baseKey}-${number}`, rerender];
 }
--- a/src/hooks/useMapToReactNumberConfig.ts
+++ b/src/hooks/useMapToReactNumberConfig.ts
@@ -4,20 +4,20 @@ import type { IUseLanguage } from 'src/hooks/useLanguage';
 import type { IInputFormatting } from 'src/layout/layout';
 import type { CurrencyFormattingOptions, UnitFormattingOptions } from 'src/utils/formattingUtils';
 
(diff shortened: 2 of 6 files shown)


lassopicasso (approved):
Nice work on the fix(es)! Regarding the potential issue with fixing the crash caused by long decimal inputs and not receiving any feedback when the data is not saved, do I remember correctly that backend was either already working on or planning to address this in the future?

bjosttveit:
> Nice work on the fix(es)! Regarding the potential issue with fixing the crash caused by long decimal inputs and not receiving any feedback when the data is not saved, do I remember correctly that backend was either already working on or planning to address this in the future?

I think it has been discussed. If so we could just remove this again.

src/hooks/useReload.ts
```diff
@@ -0,0 +1,7 @@
+import { useCallback, useState } from 'react';
+
+export function useReload(key: string): [string, () => void] {
```
olemartinorg: Huh, it's non-obvious to me what this is supposed to be used for.. Force-reloading a component? 🤔 Would love to see some doc comments with an example of use.

bjosttveit: I added some comments, it looks like you reviewed not the latest commit

PR #1360, opened 2023-08-03
Updated design system and tokens, and added ignoreDeps

bjosttveit:
## Description

Updated design system and tokens. Also added design system and tokens to ignoreDeps in renovate config so that it stops trying to update them automatically. This is due to adding the tokens package separately which had to be done a little while ago to fix the version resolution since it conflicted with altinn design system. Therefore we usually have to update these two packages together, which renovate does not.


olemartinorg (approved):
It should be possible to [group them together](https://docs.renovatebot.com/configuration-options/#packagerules) to get PRs from Renovate on updates, and we could also set the `schedule` for these to `at any time` and `minimumReleaseAge` to something small - that way we'd get PRs as soon as new versions are released.

bjosttveit:
> It should be possible to [group them together](https://docs.renovatebot.com/configuration-options/#packagerules) to get PRs from Renovate on updates, and we could also set the `schedule` for these to `at any time` and `minimumReleaseAge` to something small - that way we'd get PRs as soon as new versions are released.

Nice, I did not know that!

PR #1368, opened 2023-08-07
Fix `Checkboxes` to use `lang` instead of `langAsString`

bjosttveit:
## Description

On the `Checkboxes` component, `langAsString` was used for description and help text instead of `lang`, which is what `RadioButtons` uses. I let `label` still use `langAsString` on both checkbox and radio. The reason for this is twofold: 1. If the label is not of type string, then the help text button will [not get an accessible name](https://github.com/digdir/designsystem/blob/main/packages/react/src/components/_CheckboxRadioTemplate/CheckboxRadioTemplate.tsx#L119-L124). This is a design system issue. 2. useDisplayData expects to return a string, and if formatting is used on the label then this will not look good in summary. Is it possible to still get a pure string representation from raw markdown? 

## Related Issue(s)

- https://altinn.slack.com/archives/C02FK32FBR6/p1691395951702349

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
  - [x] No changes made
- Support in Altinn Studio
  
- Sprint board
- Labels

### DIFF
--- a/src/layout/Checkboxes/CheckboxesContainerComponent.tsx
+++ b/src/layout/Checkboxes/CheckboxesContainerComponent.tsx
@@ -114,7 +114,7 @@ export const CheckboxContainerComponent = ({
         disabled={readOnly}
         onChange={(values) => handleChange(values)}
         legend={overrideDisplay?.renderLegend === false ? null : labelText}
-        description={textResourceBindings?.description && langAsString(textResourceBindings.description)}
+        description={textResourceBindings?.description && lang(textResourceBindings.description)}
         error={!isValid}
         fieldSetProps={{
           'aria-label': overrideDisplay?.renderedInTable ? langAsString(textResourceBindings?.title) : undefined,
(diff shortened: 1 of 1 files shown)


lassopicasso (approved):
Well done!🚀 
Can't answer the question about pure string representation from raw markdown, but I guess @olemartinorg knows this. 😃

olemartinorg:
> Is it possible to still get a pure string representation from raw markdown?

Yup, using `getPlainTextFromNode()`

olemartinorg:
I've seen that before.. 🙈 Maybe it's easier to hold off on this until the new checboxes/radios arrive from the design system? I don't think I got to the bottom of it then, but with so many layers of state handling (even in the design system) I'm not surprised there are bugs like that.

bjosttveit:
Ait, I removed the problematic bits, and I am unfortunately only left with one line 😅  On the bright side, this is really what the slack question was about so I think we can still merge this now.

PR #1379, opened 2023-08-14
Avoid page break in alert component (PDF)

bjosttveit:
## Description

Avoid page break inside the alert component in PDF. (also update `page-break-inside` to `break-inside`)

Before:
image

After:
image


mikaelrss (approved):
Looks good!

PR #1383, opened 2023-08-15
Input countdown error validation

### LINKED ISSUE #1263: Text/string countdown error circumvents normal validation logic
### Description of the bug

We have a field in `frontend-test` called `newFirstName` that counts down and only allows 4 characters max. When you type more than 4 letters you get a visual error, however:
1. Even though the field triggers validation, no error message is visible in the error report on the bottom
2. Even when navigating and submitting, the user is not stopped, and submitting the form is entirely possible even with the error present.

This seems to be by design, as this countdown circumvents our normal validation system entirely, but it might be confusing to the user that some errors need to be addressed and fixed by them - but other errors are OK to ignore.

### Steps To Reproduce

See the cypress test `components.ts` -> `should countdown remaining letters`

### Additional Information

- #565
- #1193

lassopicasso:
## Description

Using `maxLength` in the form layout indicates remaining characters and displays a visual error when exceeded, yet doesn't prevent form submission. 

This task involved more research than coding for me. I learned a lot about how validations and JSON schema work, though👍 
After some discussion, we decided to reuse the validation of the `maxLength` property in the JSON schema data model. App devs must include it in both places. I've updated the [documentation accordingly](https://github.com/Altinn/altinn-studio-docs/pull/1061). 

With both configured and exceeding max length, error report shows validation message and prevents submission. However, two messages appear below the component, one from the component itself and one from schema validation. We want to exclusively display the component message below the component. I spent some time to create a logic that checks if `maxLength` is set in both the schema and the component node. If that condition holds true, the rendered validation message is hidden. However, I couldn't find a way to extract this information directly from the schema. As a workaround, I came up with a solution where the component checks if the `maxLength` in the node exists, if it's exceeded, and if it's the only error for the component. In this scenario, I believe it's quite likely that the validation message is from `maxLength` validation in the schema. 

There may be some cases I haven't considered. 🤔 If this workaround isn't accepted as a solution, we might need to retain the validation message below the component, particularly if there isn't a straightforward method to check the schema for `maxLength`.

The component continues to show the visual error when `maxLength` is only set in the layout. This is because the visual error is triggered by the `characterLimit` prop, which overrides `isValid` when the length is exceeded. In the absence of JSON schema's `maxLength`, the component should remain neutral, without displaying a "red alert," while still showing the component message below. I've created an issue regarding this matter at digdir/designsystem#719. They are in the process of completely rebuilding the component, so it might take some time before this issue is resolved.

## Related Issue(s)

- closes #1263 

## Verification/QA

- Automated tests
  - [x] Has been added/updated
  https://github.com/Altinn/altinn-studio-docs/pull/1061
  
  - [x] No changes made
- Support in Altinn Studio
  
- Sprint board
- Labels


lassopicasso:
After the meeting today, I've reflected on this PR. It was mentioned that it could lead to breaking changes. Many apps have already set up `maxLength` in the component layout.

I'm unsure about potential breaking changes. The idea behind the PR is to display only one message instead of two when `maxLength` is configured in both the data model and layout. I haven't found a way to determine the source of the validation message, so it checks if `maxLength` exists in the layout, if it's exceeded, and if there's only one validation message. This likely indicates components with both `maxLength` in the data model and layout. This is a bit risky as it can't guarantee this is the case. There might be exceptions where a validation related to something else exists..

Perhaps it's best if I remove this logic and go with the other approach where two messages appear below the component, but where design removes the validation from the component and keeps it neutral unless the corresponding maxLength is set in the data model?

src/layout/GenericComponent.tsx
```diff
@@ -117,6 +117,14 @@ export function GenericComponent<Type extends ComponentTypes = ComponentTypes>({
     shallowEqual,
   );
 
+  //Hide validation message if component maxLength exists and exceeded (handled by component's own message) and it's the only error
+  const componentMaxLength = node.item?.maxLength;
+  const hideValidationMessage =
+    componentMaxLength &&
+    formData?.simpleBinding &&
+    formData?.simpleBinding?.length > componentMaxLength &&
+    componentValidations?.simpleBinding?.errors?.length === 1;
```
olemartinorg, line 126: I agree - we could remove this check (or make it more robust? It could possibly check if the one error message is the one for maxLength in the model as well)

src/layout/GenericComponent.tsx
```diff
@@ -276,7 +285,7 @@ export function GenericComponent<Type extends ComponentTypes = ComponentTypes>({
       </FormComponentContext.Provider>
     );
   }
-
+  console.log(componentValidations?.simpleBinding?.errors?.length === 1);
```
olemartinorg, line 288: Should probably remove this! 😉

olemartinorg (approved):
I agree, these changes are not breaking after all - I didn't look carefully enough at it last time. The 'breaking' part is in the documentation - as app developers will have to make sure their data models are in sync with the layout.

src/layout/GenericComponent.tsx
lassopicasso: Yeah that is a good point 👍 I pushed an update where the validation messages are filtered for maxLength-messages. This seems to work smooth. 
I'm not sure, but it doesn't look like app devs can customize this validation text. However, if it's possible to customize or we plan it in the future, I found a (hacky?) way to check if maxLength in datamodel exists and its value.
I left it out for now, because I'm unsure if this is "smør på flesk". I can implement it if you think it's a good idea.

```
const schema = useAppSelector((state) => state.formDataModel.schemas);
const flatSchema = Object.assign({}, ...Object.values(schema));
const bindingValue = (node?.item?.dataModelBindings as { simpleBinding?: string })?.simpleBinding;
const schemaKey = bindingValue?.split('.').slice(-2, -1)[0];
const schemaMaxLength= schemaKey && flatSchema.definitions?.[schemaKey]?.properties.value?.maxLength;
```

_note: the tests runs well in local, I just need to update tt02._

olemartinorg: Ah, naah - I don't think that's all that reliable - not all schemas are even built using `definitions`, and looking up in the Schema is unreliable even in `useBindingSchema()` that uses a 3rd party tool that properly parses the schema to do it. Otherwise this looks good! 👍

PR #1406, opened 2023-08-24
Auto-generated TypeScript and JsonSchema for components

### LINKED ISSUE #832: Auto-genereate layout schema based on TypeScript interfaces

olemartinorg:
## Description
A summer project that fell into my lap. For the longest time we've been writing our own JsonSchema for the component/layout definitions, and it has caused us to forget certain properties, under-document others, and even specifying the wrong types/options for some of them. To make sure we always update both our internal types and the external types at the same time, I've written a code generator that lets us configure components in code, and generate the resulting JsonSchema and TypeScript types.

A few new/breaking concepts:
- We don't use ExprResolved and ExprUnresolved anymore. Now we instead have Internal and External type suffixes for our components. External types are those in the layout files, the flat configuration of components with expressions, `children` etc. Internal types are those used internally in `app-frontend-react` after evaluating expressions and building the node hierarchy. In these types, expressions only have their results (`string`, `boolean`), and some properties may change entirely (such as `children` -> `childComponents`/`rows`).
- Our TypeScript types are generated with both internal and external types for all components and all of their properties
- The JsonSchema only generates the external types, because our internal details here are irrelevant
- Each component now should have a `config.ts` file, which exports a `Config` instance. Using the `CG` class (short for code generator) we can start building configurations for each component.
- In `src/layout/*/index.tsx` the main component def class now extends the parent (generated) class. So for example, `export class Group extends GroupDef {...}`. This is done so that the component _category_ (previously named component type) is set only once, in the `config.ts` file, and is what decides what the component should extend (`FormComponent`, `ActionComponent`, `ContainerComponent`, `PresentationComponent`).
- Lots of properties that were possible to set on all components are now only possible to set on those components where they make sense. So for example, a `Header` component cannot have the `readOnly` property - that only makes sense for form components.
- `textResourceBindings` are still always optional, but we should consider if some of them should be required in the future (for instance, it does not make sense to add a `Header` component that does not have any text in it).
- `dataModelBindings` can now be either optional or required, but existing ones have only been set to required in the external types - because not all apps have set them for all components, and we should make sure to still check for that when using our internal types. An `Input` for example, does not work properly without being bound to the data model, so these should be required properties.
- All unknown/additional properties in the JsonSchema are disallowed by default, so existing apps will get a lot of warnings when they've added `"dataModelBindings": {}` etc for components that don't even support data model bindings.
- The `Group` component has been split into 4 different sub-types, so that we can have different `textResourceBindings` and specify properties that can only be set on one of the sub-types:
  1. Repeating groups
  2. Repeating groups with `edit.mode = 'likert'`
  3. Non-repeating groups
  4. Non-repeating groups with a `panel` property

## Related Issue(s)

- closes #832 
- replaces #1283
- https://github.com/Altinn/altinn-studio/issues/9529
- #1315

## Verification/QA

- Automated tests
  - [x] Has been added/updated
    - https://github.com/Altinn/altinn-studio-docs/pull/1084
  
- ~~Changes/additions to component properties~~
  - ~~[ ] Changes are reflected in both `src/layout/layout.d.ts` and `layout.schema.v1.json`, and these are all backwards-compatible~~
  - ~~[ ] No changes made~~
- Support in Altinn Studio
  
- Sprint board
- Labels


olemartinorg:
The failing Cypress seems like an entirely unrelated fluke, and tests pass on my machine. Opening this up for review (I also disagree with SonarCloud here)

bjosttveit (approved):
# :goberserk:

PR #1416, opened 2023-08-30
Making sure error messages on submit are still shown even if they're …

olemartinorg:
## Description

This makes sure we still display all error messages/validations sent by the server when the user is submitting the form, even if we end up mapping them to components that are currently hidden (or on hidden pages), regardless of the method used to hide them.

This should alleviate a few recent cases where users have ended up with broken instances and 'unknown error's/409s when submitting instances, likely because of broken backend logic causing validation messages to appear when they shouldn't.

Extensive tests have been added to make sure this doesn't happen again.

## Related Issue(s)

- https://altinn.slack.com/archives/C02FK32FBR6/p1693219611398259

## Verification/QA

- Manual functionality testing
  - [ ] I have tested these changes manually
  - [ ] Creator of the original issue (or service owner) has been contacted for manual testing (or will be contacted when released in alpha)
  - [ ] No testing done/necessary
- Automated tests
  - [ ] Unit test(s) have been added/updated
  - [x] Cypress E2E test(s) have been added/updated
  - [ ] No automatic tests are needed here (no functional changes/additions)
  - [ ] I want someone to help me make some tests
- UU/WCAG ([follow these guidelines](https://aksel.nav.no/god-praksis/artikler/utvikling?tema=universell-utforming) until we have our own)
  - [ ] I have tested with a screen reader/keyboard navigation/automated wcag validator
  - [x] No testing done/necessary (no DOM/visual changes)
  - [ ] I want someone to help me perform accessibility testing
- User documentation @ [altinn-studio-docs](https://github.com/Altinn/altinn-studio-docs)
  - [ ] Has been added/updated
  
  - [x] No functionality has been changed/added, so no documentation is needed
  - [ ] I will do that later/have created an issue
  
- Support in Altinn Studio
  - [ ] Issue(s) created for support in Studio
  
  - [x] This change/feature does not require any changes to Altinn Studio
- Sprint board
  - [ ] The original issue (or this PR itself) has been added to the Team Apps project and to the current sprint board
  - [ ] I don't have permissions to do that, please help me out
- Labels
  - [x] I have added a `kind/*` label to this PR for proper release notes grouping
  - [ ] I don't have permissions to add labels, please help me out

PR #1437, opened 2023-09-07
Fix: Do not filter backend validations on submit

bjosttveit:
## Description

This is related to the linked issue. I discovered this when I activated `RemoveHiddenDataPreview` in `frontend-test`, since this causes required validation to be run on the backend. This app uses old dynamics to hide fields, and so the required validation fails because the backend cannot take old dynamics into account when doing the validation. This caused an unknown error with the response `Cannot complete/close current task {currentElementId}. The data element(s) assigned to the task are not valid!`. This change makes it so that validationissues are not filtered out by source on submit, which should be fine since the backend validation is not called until all frontend validations are fixed so it should not cause duplicates.

## Related Issue(s)

- https://github.com/Altinn/app-frontend-react/pull/1416

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels

### DIFF
--- a/src/features/formData/submit/submitFormDataSagas.ts
+++ b/src/features/formData/submit/submitFormDataSagas.ts
@@ -74,6 +74,7 @@ function* submitComplete(state: IRuntimeState, resolvedNodes: LayoutPages) {
     false, // Do not filter anything. When we're submitting, all validation messages should be displayed, even if
     // they're for hidden pages, etc, because the instance will be in a broken state if we just carry on submitting
     // when the server tells us we're not allowed to.
+    false, // Do not filter sources. We want to show all validation messages for the same reason as above.
   );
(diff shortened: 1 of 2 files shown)


olemartinorg (approved):
Change looks good. I guess adding a cypress test case for this would not be all that easy, without simply setting the flag on frontend-test as well? Which we might not want for some other reason? Or could we set the flag on at least one of our test-apps and create a cypress test case that would fall over if we break this in a future refactor?

bjosttveit:
Yeah, using `RemoveHiddenDataPreview` in frontend-test would mess up other cases where old dynamics is used, and it cannot be set dynamically at runtime so that would be tricky indeed

PR #1450, opened 2023-09-12
Remove footer feature flag

bjosttveit:
## Description

Assuming v4 requires app-lib version >= 7.6.0, we can remove the featureflag check for footer.

## Related Issue(s)

- https://github.com/Altinn/app-lib-dotnet/pull/183

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels


bjosttveit:
> And save 2 lines of code? 🤔 I'm not quite sure of the benefit of this. And as long as we don't have any other checks to make sure the backend version is supported, I'm not sure we should just pull out all the stops.

I thought we had agreed to have a minimum supported backend version?

olemartinorg:
Possibly (I can't recall specifics), but I think we should implement some kind of sanity check in that case, and warn app developers in the frontend immediately if their backends are not updated.

And I also believe the feature flags from backend were supposed to be just that; flags that could be turned off. Although I guess for this specific feature it just guards against us calling an API endpoint that didn't exist in previous versions. So, what happens after this change - will older app-backends produce the 'unknown error' if the footer request fails? In that case, I agree that removing the feature flag check is a good thing.

bjosttveit:
> And I also believe the feature flags from backend were supposed to be just that; flags that could be turned off. Although I guess for this specific feature it just guards against us calling an API endpoint that didn't exist in previous versions.

Yes this one is hardcoded to `true` only to prevent old apps from getting 404 (as opposed to 204) in their logs.

> will older app-backends produce the 'unknown error' if the footer request fails?

Not unless we also disable the `treat 404 as ok` logic for footer as well, which we could very well do

olemartinorg:
> Not unless we also remove the treat 404 as ok logic for the footer as well, which we could very well do

Yep, do it! 🙌 Old backend = fatal error. That's much better than old backend = multiple fixed bugs/issues creeping back to the surface.

olemartinorg (approved):
Sorry, though I had approved it! 🚀

PR #1455, opened 2023-09-13
Fix validation message with tag title (v4)

bjosttveit:
## Description

Quick fix that uses language to show the text-resource binding for tag-title in validation message. Previously it would just show the raw textResourceBinding.


lassopicasso (approved):

olemartinorg (approved):
Looks good! I usually just cherry-pick instead of creating two PRs, but this is fine as well.

bjosttveit:
> Looks good! I usually just cherry-pick instead of creating two PRs, but this is fine as well.

Cherry-pick directly to v3?

olemartinorg:
Yup, that's what I do. While generally discouraged, we have permissions to commit/push directly to both the `main` and `v3` branches. While making a PR + reviewing is usually preferred, changes such as this (that should be the same on two branches) could just be cherry-picked and pushed, I think.

PR #1516, opened 2023-09-27
Remove `processActions` feature flag

bjosttveit:
## Description

This is another static feature flag that is hardcoded in the backend to allow backwards compatibillity with older versions of app-lib. Assuming that v4 requires v8 of app-lib, we can get rid of this featureflag and simplify some logic.

https://github.com/Altinn/app-lib-dotnet/issues/205


bjosttveit:
Setting as draft. This requires test-apps to be updated to v8.

olemartinorg (approved):
Code looks good, but ping me again if you make any changes after upgrading the test-apps. 🚀

bjosttveit:
> Code looks good, but ping me again if you make any changes after upgrading the test-apps. 🚀

All good 👍

PR #1519, opened 2023-09-27
Switch to `-v3` test apps

bjosttveit:
## Description

- Copied and deployed `-v3` versions of the test apps used for cypress
- I kept the app name the same (e.g. `frontend-test`) so I had to change out some places where the app-id variable (`frontend-test-v3`) was used in place of the app name.
- The update to wcag violation was due to the existing signing test app not being deployed to the latest version in main. I have deployed a new version there and updated this in `main` as well.


olemartinorg (approved):
Thank you! 🚀

PR #1560, opened 2023-10-11
Make PDF tests less hacky

bjosttveit:
## Description

Use `.then()` instead of setting default command timeout.


olemartinorg (approved):
Looks good! Assuming you tested it as well. I would assume you'd need a `.should('exist')` in there as well, but if it works without it, that's all good.

PR #1599, opened 2023-10-24
feat: add a sortOrder property to Dropdown component to be able to so…

mikaelrss:
## Description
Adds a new property `sortOrder` to the `Dropdown` component. This property can take on the values `"asc"` and `"desc"`, specifying whether the options should be sorted alphabetically in ascending or descending order. If the property is not defined, then the options list will not be sorted.

## Related Issue(s)

- closes #1598 

## Verification/QA

- Manual functionality testing
  - [x] I have tested these changes manually
  - [ ] Creator of the original issue (or service owner) has been contacted for manual testing (or will be contacted when released in alpha)
  - [ ] No testing done/necessary
- Automated tests
  - [x] Unit test(s) have been added/updated
  - [ ] Cypress E2E test(s) have been added/updated
  - [ ] No automatic tests are needed here (no functional changes/additions)
  - [ ] I want someone to help me make some tests
- UU/WCAG ([follow these guidelines](https://aksel.nav.no/god-praksis/artikler/utvikling?tema=universell-utforming) until we have our own)
  - [x] I have tested with a screen reader/keyboard navigation/automated wcag validator
  - [ ] No testing done/necessary (no DOM/visual changes)
  - [ ] I want someone to help me perform accessibility testing
- User documentation @ [altinn-studio-docs](https://github.com/Altinn/altinn-studio-docs)
  - [ ] Has been added/updated
  
  - [x] No functionality has been changed/added, so no documentation is needed
  - [ ] I will do that later/have created an issue
  
- Support in Altinn Studio
  - [ ] Issue(s) created for support in Studio
  
  - [x] This change/feature does not require any changes to Altinn Studio
- Sprint board
  - [x] The original issue (or this PR itself) has been added to the Team Apps project and to the current sprint board
  - [ ] I don't have permissions to do that, please help me out
- Labels
  - [x] I have added a `kind/*` label to this PR for proper release notes grouping
  - [ ] I don't have permissions to add labels, please help me out

PR #1614, opened 2023-10-27
feat: add support for sortOption for all selection components

mikaelrss:
## Description

Adds the support for a `sortOrder` property to all selectable components and not just Dropdown.

## Related Issue(s)

- closes #1599 

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels

### DIFF
--- a/src/features/options/useGetOptions.ts
+++ b/src/features/options/useGetOptions.ts
@@ -1,6 +1,7 @@
 import { useRef } from 'react';
 
 import { useGetOptionsQuery } from 'src/hooks/queries/useGetOptionsQuery';
+import { useLanguage } from 'src/hooks/useLanguage';
 import { useSourceOptions } from 'src/hooks/useSourceOptions';
 import { duplicateOptionFilter } from 'src/utils/options';
 import type { IMapping, IOption, IOptionSource } from 'src/layout/common.generated';
@@ -43,6 +44,8 @@ interface Props<T extends ValueType> {
 
   // Fetch options from repeating group
   source?: IOptionSource;
+
+  sortOrder?: SortOrder;
 }
 
 export interface OptionsResult {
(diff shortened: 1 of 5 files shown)


mikaelrss:
Percy fails with a bunch of new snapshots. This might be because I deployed a new version of frontend-test, and later rolled back to the old version. In any case, I don't think any of the screenshots are related to the code in this PR.

src/features/options/useGetOptions.ts
```diff
@@ -52,20 +54,32 @@ export interface OptionsResult {
 
 const defaultOptions: IOption[] = [];
 
+type SortOrder = 'asc' | 'desc';
+const compareOptionAlphabetically =
+  (sortOrder: SortOrder = 'asc') =>
+  (a: IOption, b: IOption) => {
+    const comparison = a.label.localeCompare(b.label, ['nb', 'en'], { sensitivity: 'base', numeric: true });
```
olemartinorg, line 61: We should use the current language instead of `['nb', 'en']` here. You cat get the currently selected language key via:

```typescript
const { selectedLanguage } = useLanguage();
```

olemartinorg (approved):
Looks good! I don't think we need to rush v3 support for this (generically for all components) unless you need it in your project. The way this works in v3 is very different, so it may be hard to get working the same way there.

mikaelrss:
Alright. We have solved it with some backend-code as of right now, so we technically don't _need_ this in v3

olemartinorg:
When testing this against the `reasons` radio list in `frontend-test` and adding language keys for one of the options there, it seems that:

- The options are sorted according to their text resource key, not the actual label (I set the key to something beginning with `z`, so it always came last, even though the text resource value started with `a`)
- The options don't always change according to the changing language (when using the language selector). I assume that problem wasn't introduced here.

https://github.com/Altinn/app-frontend-react/assets/700139/cced6a00-c7fb-4a67-9f5c-7e7726ff9501

I think we should add a Cypress test for this scenario, and fix the underlying issues with it. 🙏

mikaelrss:
Thanks for the thorough testing! I will add a cypress test for this case and fix the issue :smile:

Issue #1601, opened 2023-10-24
Migrate to the new Popover component (from the design system)


olemartinorg:
This issue affects us too:
- https://github.com/digdir/designsystem/issues/974
- https://altinn.slack.com/archives/C02ESDSC929/p1698147060692119

We should start using the new Popover which shouldn't have this problem.

PR #1617, opened 2023-10-30
Upgrade design system to 0.35.0

bjosttveit:
## Description

Switched `DeleteWarningPopover` to use `LegacyPopover`, for migration issue, see https://github.com/Altinn/app-frontend-react/issues/1601.

It is possible that this PR fixes the issues related to `HelpText`, as this should already use the new popover.

## Related Issue(s)

- closes #1615


bjosttveit:
Looks like `LegacyTextField` has been broken...

image

PR #1633, opened 2023-11-03
refactor properties of IOptionSource to have suppport for expressions.

mikaelrss:
In #1614 support for dynamic expressions in code lists based on repeating groups was added. This PR did not include an update to the properties of `IOptionSource` which is the basis for generating schema, types, and future documentation of these properties. 

This PR adds these properties.


olemartinorg (approved):
🙌

Issue #1556, opened 2023-10-10
Summary/PDF - rendered before dynamic options are loaded


lassopicasso:
### Description of the bug

In the context of implementing Tanstack Query in useGetOptions and PDF, we've identified a bug that occurs in certain scenarios where options are not fully loaded/fetched before they are displayed in the PDF or Summary. 

This issue is primarily associated with dynamic options, which do not appear to finish loading before they are rendered in PDF or Summary. It has been observed by @bjosttveit that the `initiallyLoaded` variable within `useAllOptions` is being set to true before it has actually finished loading. This was done by wrapping the reducer of `allOptions`, enabling it to log the state of `initiallyLoaded` for each update. 

![image](https://github.com/Altinn/app-frontend-react/assets/74791975/8e2e1598-e3ba-437f-861c-6653fd355d39)

### Steps To Reproduce

Here's an example to reproduce this issue:
1. Start the frontend test app in the `changeName` form.
2. Reduce the network speed (I set it to d: 200kb/s, u: 200kb/s and latency: 4000ms).
3. Refresh the page
4. In the drop-down lists under "Referanser", change the option in source, then quickly navigate to the summary page.
5. You will probably notice that "Referanse2" hasn't had enough time to fetch before the summary is rendered. It should display the same value as Referanse.
![image](https://github.com/Altinn/app-frontend-react/assets/74791975/a271bec7-3bb9-4c9d-9892-bab7a417aad1)

6. If you refresh the page while on the summary page, the issue won't resolve itself. To fix this, you need navigate back to `form` so that these dynamic options can finish rendering. After that, navigate back to the summary page again.

bjosttveit:
Looks like a different bug, but values in Referanse and Referanse 2 are not getting cleared when the source field changes. This Works in frontend-test-v3.

https://github.com/Altinn/app-frontend-react/assets/47412359/564b3cd9-903a-4c2a-bb8f-1c42cf22d067

RonnyB71:
The different bug mentioned, where Referanse and Referanse 2 doesn't get updated. In my case, Referanse gets cleared when the first dropdown changes, but Referanse 2 does not. Referanse 2 does get cleared when Referanse changes.

Issue #1507, opened 2023-09-25
Rewrite process/task management to hooks/Tanstack query


olemartinorg:
We have some sagas:
- watcherFinishDataTaskIsloadingSaga
- watcherFinishStatelessIsLoadingSaga
- completeProcessSaga
- getTasksSaga
- startInitialInfoTaskQueueSaga
- watchStartInitialInfoTaskQueueSaga

We have a few slices:
- queueSlice
- isLoadingSlice
- processSlice

We have some sagas that fetch data from the backend:
- getInstanceDataSaga
- getUpdatedProcess
- checkProcessUpdated
- getProcessStateSaga

And we already have one hook:
- useProcess

But all of this could be simplified a bit. Currently, `App.tsx` implements a hash-based router, and renders either `Entrypoint` or `ProcessWrapper`, or possibly `PartySelection` if we need to select another party. When looking at the possible results of `Entrypoint` and `ProcessWrapper` there are some overlaps, and some pitfalls. I.e.:

- Some operations only happen in `ProcessWrapper` and will not have any effect in `Entrypoint`. This means that `useApiErrorCheck()`, PDF rendering/preview, receipt or any other task-types than data **will not work in stateless apps** that only renders Entrypoint (as we can tell from the router in `App.tsx`, statless apps with no instance will never reach the path needed to render `ProcessWrapper`).
- For most developers in the project, figuring out Entrypoint, useProcess and ProcessWrapper is daunting. We should strive to rewrite it in a way where logic flows more easily, and we only have one 'entry point' to rule them all.
- We should pool up all the needed requests/jobs for one task/page to load into a structure where we can easily tell that more loading is needed or not. Currently the code in queueSlice and isLoadingSlice attempts to do this, but it slightly broken (for example, the *fulfilled actions in queueSlice will be triggered as soon as loading starts, not when it ends).

PR #1643, opened 2023-11-10
The big cleanup, part 1

### LINKED ISSUE #1504: Rewrite attachment sagas to use hooks + Tanstack Query
This includes the sagas for:
- Uploading attachments
- Updating attachments
- Deleting attachments
- Mapping attachments

And may also include:
- `deleteAttachmentReference` (for supporting attachments inside repeating groups), if not handled in #1175

### LINKED ISSUE #1511: Rewrite instantiationSaga to Tanstack Query
The `instantiationSaga` and associated `instantiationSlice` + `InstantiationActions.instantiate()` should be rewritten to use a Tanstack Query mutation.

olemartinorg:
## Description
Sorry about the HUGE number of changes here. Doing this in smaller batches would make it far more time consuming, and as rewriting lots of what our entire project is built on _is a lot of work_, the size of this PR sadly reflects that.

Architectural changes:
- Some redux slices have now been removed: `attachmentSlice`, `instanceDataSlice`, `instantiationSlice`, `isLoadingSlice`, `processSlice` and `queueSlice`. The not-quite-a-slice `appApi` has also been removed (rewritten to use tanstack query).
- All of the code for controlling when to show the "wait while we're loading your data" has been rewritten, and instead of trying to store 'is loading' state in redux, the frontend is now wrapped in a whole lot of context providers that all `return <Loader />` if they're not ready for the app to render yet. This sadly also makes loading an app synchronous - as one provider has to finish loading before the next one can start. We'll fix that later (in #1507, I hope).
- As some of the state (like the instance data, process data, attachments and all options) are used in sagas still, I've added a `deprecatedSlice` that is used to set and store this state in redux for the time being.
- Attachment mapping, uploading, tagging and deletion of attachments have now been rewritten to store its data inside a context instead of redux. There are also some tools available now for awaiting an attachment upload.
- Attachment state is now much closer to the actual original state in an instance. In fact, when uploading an attachment we have a `TemporaryAttachment` type that describes that temporary state (which is quite different from a finished upload). A temporary attachment does not have an `id` yet, so I found a few cases where we were just happily iterating over attachments without really caring if they were uploaded or not. For example, the subform for tagging an attachment is now disabled until the upload completes. When an attachment is uploaded successfully, it will be injected into the current instance data and will be re-mapped back to the correct component. When uploaded, an attachment is really just `IData`, so it was not really neccessary to construct a different state to store the same thing.
  - This way of structuring the types will make it very difficult to write code that mistakenly tries to (for example) delete an attachment that is not yet uploaded. Also, with mapping every finished upload we make sure mapping works the same way every time, without having to write tests that reload the page after uploading to make sure mapping worked as expected. If a bug is found in the mapping code, it will be found faster this way.
- The `process` state has been entirely ripped out of the `IInstance` state. Previously we had both a `instanceDataSlice` and a `processSlice`, and while the `IInstance` state contains data about the current process as well, it does not contain everything we need (that was put into `processSlice` and the process state was fetched separately). This would lead to some code mistakenly reading outdated process state data from the instance slice instead of the process slice.
- The `useDelayedSavedState` hook caused me a lot of head ache, and 1-2 "wasted" days of debugging. It turns out it depended on timing, and when I made the code faster (by removing some of the stuff that's happening in sagas all of the time) the timing-dependent code broke in strange ways, causing tests to fail because `preselectedOptionIndex` got set and then immediately reset.
- `useMemoDeepEqual()` was slightly broken. I wrote a unit test for it and fixed it up.
- Because of the new `InstanceContext` (i.e. the place where instance data is stored) i renamed the old `InstanceContext` (i.e. the data structure used when filling in data from the instance in text resources and expressions) to 'InstanceDataSources', in order to prevent confusion.
- The new way we're waiting for loading, via rendering the <Loader /> component is visually suboptimal right now. There are cases where it leads to flashes with quickly changing information, but we can fix that later before releasing v4.
- The exported `const allOptions` in `AllOptionsProvider` broke lot of unit tests, as the state was shared between many of them. The same thing happened when `queryClient` was made global, as the query cache was persistent between tests - and they ended up being somewhat dependent on each other (or at least some started failing on me when the cache was not primed).
- A list binding now produces an actual array (list!) when fetching data from `node.getFormData()`. A nice-to-have feature that I ended up implementing when re-implementing mapping of attachments to form data in repeating groups.
- The API for `createStrictContext()` has been slightly touched-up, and I've made a new `createLaxContext()` as well. A 'lax context' is one that is not guaranteed to exist (so just a normal react context), but that we can also check if is provided (regardless of the value it provides being `undefined` or not). This was most useful for `InstanceContext`, where there's a difference between "we don't have an instance, this is a stateless app" and "we don't have an instance, we're still fetching it".
- Most of the tools in `src/utils/appMetadata.ts` now has new signatures, and I've made them all use somewhat similar parameter shapes (single object as first and only parameter).
- Probably a _lot more minor details_.

Unit test changes:
- Lots of unit tests have been rewritten - and pretty much all of them now load with the new set of providers needed to construct all the state for a regular component to run. The `renderWithProviders` function has now been split up into multiple functions, each with its set of goals, providers and tools.
  - `renderWithoutInstanceAndLayout` is the simplest one we have (apart from the plain render function from react testing library, of course). This mocks out queries, renders with the least amount of providers, and is useful for testing components that should work in stateless apps and/or without a layout.
  - `renderWithInstanceAndLayout` is the one that most closely resemble what `renderWithProviders` used to be. It renders will all the usual providers, expects you to have defined a layout (or use the default one from `getInitialStateMock()`.
  - `renderWithNode` can be used to also make sure the node hierarchy is available, and will let you render a component with a node object as a parameter. If the node cannot be found in the hierarchy for some reason, you should get a nice error about what happened.
  - `renderGenericComponentTest` does pretty much what it used to do, but now builds on `renderWithNode` and continues to provide some nice defaults for rendering a component just like `GenericComponent` would.
- Signatures for the `render*` functions have changes. They now all expect one object with a `renderer` function, and a varying amount of extra options for controlling behaviour. Query overrides are no longer passed to a third parameter. All of our render functions are now async, and when awaited will return when all providers have finished loading their data (by default, but this can be turned off if you need to assert on some loading state). A simple `await renderWithInstanceAndLayout({ renderer: () => <MyComponent /> })` call will by bootstrap all providers with sane initial state by default.
- There are now more toys to play with regarding unit tests:
  - The redux `dispatch` function is now always mocked by default, so you can assert on it without supplying a mock in your test. All redux actions are caught and logged, and you can decide if any of them should actually be dispatched for real (thus running sagas and reducers) if you want to.
  - All queries are also attached to mock functions, even if you provided your own overrides, so you can assert stuff like `expect(queries.fetchMyStuff).toHaveBeenCalledTimes(1)`.
  - All mutations are also mocked, and will never actually execute in unit tests. If you want to observe effects of them executing, you can call functions to resolve or reject them inside the tests themselves.

## Related Issue(s)

- #1507
- closes #1504 
- closes #1511
- #1175 
- This might solve #1556, I'm not quite sure

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels


olemartinorg:
Hoping you can all just review this one before the weekend! 😈

\/s

bjosttveit (approved):
# 🤷

src/setupTests.ts
```diff
@@ -31,13 +37,17 @@ Object.defineProperty(document, 'fonts', {
...
+window.logError = (...args) => {
+  throw new Error(args.join(' '));
+};
+window.logWarn = window.logError;
+window.logInfo = window.logError;
+window.logErrorOnce = window.logError;
+window.logWarnOnce = window.logError;
+window.logInfoOnce = window.logError;
+
+jest.setTimeout(env.parsed?.JEST_TIMEOUT ? parseInt(env.parsed.JEST_TIMEOUT, 10) : 10000);
```
mikaelrss, line 50: When I added around 550 unit tests for the `formatDateLocale` function, I had to increase this timeout to 15000, as the unit tests kept on timing out on CI. :innocent:

mikaelrss (approved):
:shrug: :man_dancing:

src/setupTests.ts
olemartinorg: 😱
I'll change the default here to 15000 then. That's probably what causes the merge conflict as well, I guess.

Issue #1508, opened 2023-09-25
Rewrite page navigation using hooks


olemartinorg:
We have these sagas currently:
1. updateCurrentViewSaga
2. calculatePageOrderAndMoveToNextPageSaga
3. findAndMoveToNextVisibleLayout
4. watchInitialCalculatePageOrderAndMoveToNextPageSaga

Number 2 and 4 should disappear with #1491, which leaves us with updateCurrentViewSaga, and the code in findAndMoveToNextVisibleLayout which fixes a problem caused by the fact that updateCurrentViewSaga moves to a specific page instead of mostly moving forwards/backwards through the page order.

The new code should prefer using separate methods for next/prev instead of jumping straight to a specific page, to avoid problems like those solved in findAndMoveToNextVisibleLayout whenever possible.

### Related issue(s):
- #1446
- #1408

RonnyB71:
Based on the assumption that pageOrder will be removed and replaced with expressions.

PR #1663, opened 2023-11-23
Refactoring app flow (the big cleanup, part 2)

### LINKED ISSUE #1507: Rewrite process/task management to hooks/Tanstack query

olemartinorg:
## Description

Changes in arhitecture/structure:

- Removed a few redux slices; `dataListsSlice`, `optionsSlice` and `partySlice`.
- Lots of things have now gotten their own providers and contexts, such as application metadata, the organisation list, application settings, layout-sets, footer, profile, parties, text resources, and even the keep-alive functionality and the window title.
- Removed functionality for rendering an error when there is an Error object stored in one (of the many) redux slices. This is now implemented by rendering the `<DisplayError />` component inside the various context providers instead.
- Instead of every other provider needing to call their `useSomeQuery()` hook, render `<DisplayError />` if there are errors, and render `<Loader />` while the query is loading, I made a tool with `createQueryContext()` which does all of this for you.
- Because all contexts are created on the root (`const {...} = create*Context(...);`) and re-exported, I got a lot of problems with circular imports causing the provider and/or useCtx to be undefined. To solve that, I created a `delayedContext()` tool that delays the realization/actual createContext() call until the provider/useCtx is used the first time - thus delaying the context creation until all imports are done.
- I rewrote the `createStrictContext()` and `createLaxContext()` functions to a single function that differentiates between a `required: true` or `required: false` context. It is even possible to set a default value for a context now that will be the value used if you call `useCtx()` without even having a provider - so even optional/lax contexts can have a non-`undefined` return value.
- I've started moving some utilities into `src/core` instead of having a 'assorted bunch of important files' in `src/utils`.
- `index.tsx` now renders a lot of providers we always need, and then renders `App.tsx` which is simply a router. I expect this to change drastically again in #1508.
- `Entrypoint` has been drastically simplified, and now tries to have a clearer purpose; it decides what to do when you open the entrypoint (the `/` url path). For some apps we redirect to the instance selection (which now has its own route), we also redirect to party selection, or call the `<InstantiateContainer />` when we need to create a new instance. The aim here is to make all of this easier to follow when reading the code to try to understand how apps are initialized.
- `ProcessWrapper` has also been somewhat simplified, and now simply switches between process tasks to render.
- Party selection logic has been altered significantly. I noticed that we fetch available parties with a filter, so that we only get a list of parties that are allowed to instantiate - and then we used to call a separate endpoint to verify that the party we selected is allowed to instantiate. I dropped that request as I deemed it unnecessary, and tried to simplify the logic.
- The current language was set in redux, specifically the `profileSlice`, but I've now moved this to a `LanguageProvider` that sets the language very high up in the provider stack (because it is needed to decide about rendering right-to-left language) and allows the language to be set via the URL, via localStorage, via the loaded profile, and via the language selector.
- I rewrote the Altinn logo fetching to use Tanstack Query, fetching the SVG file and allowing the fill color to be controlled by changing a property to the logo component. We used to have some tricky CSS filtering on top of the logo to achieve this, and we often downloaded both the black and the blue logo while loading an app. We should now only download it once, and re-color it as needed.
- In unit tests, I've brought in a few more of the `get*Mock()` functions to provide default mocks for queries we run. We're still in between a rock and a hard place where some of our components use state directly from contexts/providers, while others use redux (which have entirely separate data in unit tests in some cases), but it's getting better with each cleanup pass.
- I've rewritten a few redux selectors to be hooks instead. Such as `useAppName()`, `useAppOwner()`, etc. I brought their tests along with the rewrite, but I had to rewrite large parts of them too.

## Related Issue(s)

- closes #1507
- #1643

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels


bjosttveit (approved):
# 🤷

PR #1671, opened 2023-11-29
The `<Lang />` component to replace `useLanguage()`

olemartinorg:
## Description

The `useLanguage()` hook is great and all, but it's not really idiomatic React, and we could do better. With the steady decline of Sagas and functions that take `langTools` as a bunch of functions used to _do stuff with later_, we can do the next great thing; referencing language keys using a React component instead.

Right now, this component is an error boundary, which makes it possible for us to render language keys (at least our own) very early in the app runtime, even before the `<LanguageProvider />` runs, and still produce some resemblance of a coherent output. This fixes potential problems I (and @bjosttveit) saw where errors early in the provider tree will throw new errors again when `<UnknownError />` may be rendered even before we have a language.

In time, we can also use this component to do something smarter. Since it takes the language key as a property, we can look up the key and branch out to different implementations in runtime. With `useLanguage()` we kind-of guarantee that we have a toolbox ready-to-go with everything you'll ever need, but that's not something we can really guarantee. For example, you can have a text resource with a variable looking up a value in a data model, and if you don't actually use that text resource we won't have to fetch the data model either. Currently we have to load every data model referenced in any of the text resources when in an info-task (see `FormDataProvider`) for this to work, but in the future we can postpone the loading until `<Lang />` is called with a key that actually needs the data model. In other words, we can't return a reactive set of functions from `useLanguage()`, but a component can be reactive.

## Related Issue(s)

- closes #{issue number}

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels

### DIFF
--- a/src/__mocks__/getLogoMock.ts
+++ b/src/__mocks__/getLogoMock.ts
@@ -1,0 +1,3 @@
+export function getLogoMock() {
+  return '<svg width="10mm" height="3mm" xmlns="http://www.w3.org/2000/svg"><text font-size="14px" y="10" fill="black">Altinn</text></svg>';
+}
--- a/src/components/LandmarkShortcuts.tsx
+++ b/src/components/LandmarkShortcuts.tsx
@@ -9,7 +9,7 @@ export interface ILandmarkShortcutsProps {
 
 interface ILandmarkShortcut {
   id: string;
-  text: string;
+  text: React.ReactNode;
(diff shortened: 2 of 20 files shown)


bjosttveit (approved):
## LGTM 🚀

PR #1694, opened 2023-12-07
Feat/133 magic button

### LINKED ISSUE #133: Form element: Generic button
## Description
Look into if service developers needs a generic button that can be used for different functions with some coding. If there is a need for a button with a specific function decide if the button type should be standard components. (example: cancel or run calculations) 

This was originally included in issue Altinn/altinn-studio#1025, but was seen as out of scope after discussion with developers. 

## Technical considerations
Input (beyond tasks) on how the user story should be solved can be put here.

## Acceptance criterea
- [ ] The button should be a new button type on par with Submit, Navigate and ActionButton. Proposed type is CustomButton
- [ ] It should be possible to place the button anywhere on the form jus as any regular component
- [ ] The button should be disabled if the user don't have access to the button or only has read access to the form
- [ ] The should call a new api endpoint on backend providing the required parameteres for backend to identify which button was used and other context variables, but not the data. Frontend is responsible for storing the data ahead of calling the api, backend is responsible for getting any data it needs to perform the action.
- [ ] In the case of success backend should reply with an object telling frontend what data was changed, what action frontend is supposed to take after execution (_none_ aka stay on the page by default, _navigateNext_).
- [ ] In the case of success, consider also if backend should return a customizable, and localized, success message to tell the user what happened eg. "A message was sent to the persons you specified", "Data updated"
- [ ] In the case of an error, backend should provide a standardized error code/message and the ability to customize and localize that message to inform.

## Out of scope
- [ ] Unless prove very simple to add, the use of a button within a repeating group is out of scope.

## Specification tasks
- [x] Find out if there is a need for a generic button.
- [x] If there is a need, design the generic button as a stand alone component or as a part of the submit  component. 
- [x] If there is a need for a button with a specific function, design this component.
- [x] Write acceptance criterea.

mikaelrss:
## Description

This PR introduces a new component `CustomButton`. This component let's application developers perform a custom function (and predefined functions in the frontend) with the click of a button.

The API looks like this: 
```json
{
  "id": "testing-custom-button",
  "actions": [
    {
      "name": "fill",
      "type": "ServerAction"
    }
  ],
  "type": "CustomButton"
}
```
The button has an array of `actions` that are performed in sequence. Each action could have a type of either `"ClientAction"` or `"ServerAction"`. `ClientActions` are performed entirely on the frontend. This PR implements three different `ClientActions`: `nextPage`, `previousPage`, and `navigateToPage`. An action can specify parameters with the `metadata` property. 

e.g: 
```json
{
  "id": "testing-custom-button-2",
  "actions": [
    {
      "name": "navigateToPage",
      "type": "ClientAction",
      "metadata": {
        "page": "grid"
      }
    }
  ],
  "type": "CustomButton"
}
```
Here, the `navigateToPage` function receives the object ` { page: string }` as an argument. 

A `ServerAction` can return any number of `ClientActions`, and these will be run before the next action specified in the `actions` parameter of `CustomButton`. 

If there is an error in a `ServerAction`, we display a toast with the error message specified in the response. I've implemented the toast in the same way Studio recently did:
image

 There is currently no Toast component in the design system, but it has been suggested as a possible future component. 

## Related Issue(s)

- closes #133 

## Verification/QA

- Automated tests
  - [x] Has been added/updated - https://github.com/Altinn/altinn-studio-docs/commit/be3c11c911f0f18c9c317ea8bd70f47e66510e6d
  
- Support in Altinn Studio
  
- Sprint board
- Labels

### DIFF
--- a/src/hooks/useNavigatePage.ts
+++ b/src/hooks/useNavigatePage.ts
@@ -179,9 +179,6 @@ export const useNavigatePage = () => {
     [processTasks],
   );
 
-  const navigateNext = useCallback(() => navigateToPage(next), [next, navigateToPage]);
-  const navigatePrevious = useCallback(() => navigateToPage(previous), [previous, navigateToPage]);
-
   return {
     navigateToPage,
     navigateToTask,
@@ -196,7 +193,5 @@ export const useNavigatePage = () => {
     currentPageId,
     taskId,
     previous,
-    navigateNext,
-    navigatePrevious,
   };
 };
--- a/src/hooks/useNavigatePage.ts
+++ b/src/hooks/useNavigatePage.ts
@@ -179,6 +179,9 @@ export const useNavigatePage = () => {
     [processTasks],
   );
 
+  const navigateNext = useCallback(() => navigateToPage(next), [next, navigateToPage]);
+  const navigatePrevious = useCallback(() => navigateToPage(previous), [previous, navigateToPage]);
+
   return {
     navigateToPage,
     navigateToTask,
@@ -193,5 +196,7 @@ export const useNavigatePage = () => {
     currentPageId,
     taskId,
     previous,
+    navigateNext,
+    navigatePrevious,
   };
 };
--- a/src/hooks/useNavigatePage.ts
+++ b/src/hooks/useNavigatePage.ts
@@ -185,14 +185,14 @@ export const useNavigatePage = () => {
   const getCurrentPageIndex = () => {
     const location = window.location.href;
     const _currentPageId = location.split('/').slice(-1)[0];
-    return order?.indexOf(_currentPageId) ?? -1;
+    return order?.indexOf(_currentPageId) ?? undefined;
   };
 
   const getNextPage = () => {
     const currentPageIndex = getCurrentPageIndex();
-    const nextPageIndex = currentPageIndex !== -1 ? currentPageIndex + 1 : -1;
+    const nextPageIndex = currentPageIndex !== undefined ? currentPageIndex + 1 : undefined;
 
-    if (nextPageIndex === -1) {
+    if (nextPageIndex === undefined) {
       return undefined;
     }
     return order?.[nextPageIndex];
@@ -200,9 +200,9 @@ export const useNavigatePage = () => {
 
   const getPreviousPage = () => {
     const currentPageIndex = getCurrentPageIndex();
-    const nextPageIndex = currentPageIndex !== -1 ? currentPageIndex - 1 : -1;
+    const nextPageIndex = currentPageIndex !== undefined ? currentPageIndex - 1 : undefined;
 
-    if (nextPageIndex === -1) {
+    if (nextPageIndex === undefined) {
       return undefined;
     }
     return order?.[nextPageIndex];
--- a/src/hooks/useNavigatePage.ts
+++ b/src/hooks/useNavigatePage.ts
@@ -109,6 +109,9 @@ export const useNavigatePage = () => {
       if (!page) {
         return;
       }
+      if (!order.includes(page)) {
+        return;
+      }
       setFocusId(options?.focusComponentId);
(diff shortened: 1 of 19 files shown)


mikaelrss:
The Percy-tests are failing because I added some new components to the bottom of the changename form. No unexpected changes!

src/hooks/useNavigatePage.ts
```diff
@@ -179,6 +182,56 @@ export const useNavigatePage = () => {
     [processTasks],
   );
 
+  const getCurrentPageIndex = () => {
+    const location = window.location.href;
+    const _currentPageId = location.split('/').slice(-1)[0];
+    return order?.indexOf(_currentPageId) ?? -1;
```
olemartinorg, line 188: I always prefer using `undefined` as the default return value. Returning `-1` means that the typescript return hint shows as `number`, and checking it with `typeof returnValue === 'number'` always succeeds - and you'd have to look at the implementation to know what the error value even is.

src/layout/CustomButton/CustomButtonComponent.tsx
```diff
@@ -0,0 +1,202 @@
...
+    },
+  });
+  const { handleClientActions, handleDataModelUpdate } = useHandleClientActions();
+
+  return {
+    mutation,
+    /**
+     * We wrap the mutation in a promise to make it possible to await both the mutation,
+     * and the side effects that are run in the onSuccess or onError callbacks.
+     */
```
olemartinorg, line 121: Are you sure you can't just use `mutation.mutateAsync()`? :thinking: It returns a promise, but you'll have to catch and handle any errors yourself (instead of checking `mutation.error`)

src/layout/CustomButton/CustomButtonComponent.tsx
```diff
@@ -0,0 +1,202 @@
...
+              }
+              if (data.clientActions) {
+                await handleClientActions(data.clientActions);
+              }
+
+              resolve();
+            },
+            onError: async (error: PerformActionMutationError) => {
+              if (error?.response?.data?.error?.message !== undefined) {
+                toast(error?.response?.data?.error?.message, { type: 'error' });
```
olemartinorg, line 139: I would assume this should be using a `<Lang id={error.response.data.error.message} />` as well, so that returned errors can be text resource keys (and thus translated).

src/layout/CustomButton/config.ts
```diff
@@ -0,0 +1,50 @@
+import { CG } from 'src/codegen/CG';
+import { CompCategory } from 'src/layout/common';
+
+export const Config = new CG.component({
+  category: CompCategory.Action,
+  rendersWithLabel: false,
+  capabilities: {
+    renderInTable: false,
```
olemartinorg, line 8: I think it should be possible to render this in a table (i.e. Grid). Don't quite see why that wouldn't be useful? :thinking:

src/layout/CustomButton/config.ts
```diff
@@ -0,0 +1,50 @@
...
+    renderInButtonGroup: false,
```
olemartinorg, line 9: It should absolutely render in a button group as well!

olemartinorg (approved):
Great work! :star_struck:

src/hooks/useNavigatePage.ts
mikaelrss: Yeah, I agree. I'll change the default to `undefined`, and change it at all the places it was used as well

src/layout/CustomButton/CustomButtonComponent.tsx
mikaelrss: Thats a great point! Yeah we can. I'm refactoring it, and also renaming the hook to `useHandleServerAction`  and the function returned as `handleServerAction`. That feels a little more consistent with the other hooks.

PR #1717, opened 2024-01-03
Fixes for option components + tests and adjustments

### LINKED ISSUE #555: Validation trigger on group should re-validate on delete + new row

### LINKED ISSUE #707: Kalkuleringer lagrer ikke data til datamodellfelter som ikke er i skjema
### Description of the bug

Hvis man setter oppe en calculation rule som gjør en beregning og returnere denne til et felt i datamodellen som ikke er benyttet som inputfelt i skjema så lagres ikke verdien i datamodellen. 
Dette gjelder også når en benytter denne som variabel inn i en tekst.

Workaround er å opprette et skjult felt i skjema, da lagres data som det skal, men dette er en grisete måte å løse det på...

### Steps To Reproduce

Opprette en calculation rule som returnere inn i et felt i datamodell som da ikke har noen tilknytning til noe inputfelt i skjema.
benytt datafeltet inn som en variabel i en tekst som vises for å se at den ikke setter noe data.

### Additional Information

_No response_

### LINKED ISSUE #1147: Repeating groups: Open by default blir feil ved bruk av hiddenRow
### Description of the bug

Open by default funker ikke som det skal hvis man setter hiddenRow regel slik at første raden i gruppen er filtert vekk.

Filterfeil

### Steps To Reproduce

se over.

### Additional Information
Blocked by: https://github.com/Altinn/app-frontend-react/issues/1175

### LINKED ISSUE #1686: Layout settings for receipt pages and instance selection
### Description of the bug

Currently. There is no way of setting the `showLanguageSelector` or `hideCloseButton` pages that are not a part of a layout set. #1682 Refactores these settings to use a `UiConfigContext`. This highlighed a new problem. The current way of fetching these values uses the layout settings, which mean that, for instance `showLanguageSelector` can only be specified in a layout set. This means that for instance selection and receipt pages, there is no way of showing the language selector. This worked previously because the settings of the last page was stored in redux and subsequently used in the receipt page.

### Steps To Reproduce

No steps to reproduce, but the app developers do not have a way of specifying whether the language selector should be visible on pages outside layout settings. 

### Additional Information

We need to agree on an API for the app developers to be able to set these settings also outside of the current layout set.

olemartinorg:
## Description

This PR finalizes the issues I thought I had mostly solved in #1712, but had no tests or verification. Most needed some adjustments to work properly.

- Fixing some issues with options components after form data could be more than just strings (i.e. numbers). This especially affected components with `source` options
- Fixing multiple minor issues with and added tests form them (such as #555, #707)
- Removing the need to define a `ruleHandlerHelper` and `conditionalRuleHandlerHelper`. These were not really needed for us to run rules, as you had to define an object in them that had the exact same keys as your already declared input parameters. Now we're just using the input parameters, so `ruleHandlerHelper` and `conditionalRuleHandlerHelper` can be omitted (which they are some places in `frontend-test` now, but intentionally not everywhere).
- When fixing #1686 I needed to add the `uiSettings` key to `layout-sets.schema.v1.json`. Instead of doing that manually, that file (along with `layoutSettings.schema.v1.json`) are now generated by our code generator (just like the layout schema). This works by adding all the types to `Common.ts`, and removing unused common types from the `definitions` in each of the generated schema files. So now, by reading `$ref` in the documents we'll only use those Common definitions that are needed for each schema.
- The new generated schemas mentioned above may also include a new breaking change; some page settings were previously typed to be present on both the `ILayoutSettings` object and the `IPagesSettings` object. This included settings such as `hideCloseButton`, so it may have seemed like it was possible to define that setting both inside and outside of the `pages` key. Now it's only possible inside the `pages` key (except for the deprecated `receiptLayoutName` which was read from the `ILayoutSettings` object directly, but this won't be present any longer when releasing v4). **The most notable thing here though is that it's now fully possible to set some page settings globally for the whole app via `uiSettings` in `layout-sets.json`. This includes i.e. enabling the language selector for all tasks (now including the receipt and instance/party selection screens).**

## Related Issue(s)

- closes #555 
- closes #1147 
- closes #707 
- closes #1686

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels


bjosttveit (approved):

src/features/options/useGetOptions.ts
```diff
@@ -70,33 +76,87 @@ const compareOptionAlphabetically =
...
   const calculatedOptions = sourceOptions || fetchedOptions?.data || staticOptions;
   const { langAsString } = useLanguage();
   const selectedLanguage = useCurrentLanguage();
 
-  const setMetadata = metadata?.setValue;
   const downstreamParameters: string = fetchedOptions?.headers['altinn-downstreamparameters'];
   useEffect(() => {
-    if (!!setMetadata && downstreamParameters) {
-      setMetadata(downstreamParameters);
+    if (dataModelBindings && 'metadata' in dataModelBindings && dataModelBindings.metadata && downstreamParameters) {
```
mikaelrss, line 104: Do you have to check all of these specifically or would this be covered by `dataModelBindings?.metadata && downstreamParameters`

src/index.tsx
```diff
@@ -59,7 +59,9 @@ document.addEventListener('DOMContentLoaded', () => {
           <AppWrapper>
             <LanguageProvider>
               <ThemeWrapper>
-                <RouterProvider router={router} />
+                <UiConfigProvider>
+                  <RouterProvider router={router} />
+                </UiConfigProvider>{' '}
```
mikaelrss, line 64: Remove the `{' '}` if it is accidental :smile:

src/index.tsx
```diff
@@ -82,15 +84,13 @@ function Root() {
...
-                          <DevTools>
-                            <App />
-                            <ToastContainer
-                              position='top-center'
-                              theme='colored'
-                              transition={Slide}
-                              draggable={false}
-                            />
-                          </DevTools>
+                          <App />
```
mikaelrss, line 87: Are we temporarily removing `DevTools`?

src/layout/Checkboxes/CheckboxesContainerComponent.tsx
```diff
@@ -17,33 +17,22 @@ import type { PropsFromGenericComponent } from 'src/layout';
...
   const debounce = FD.useDebounceImmediately();
 
-  const selected = value && value.length > 0 ? value.split(',') : defaultSelectedOptions;
-  const { options: calculatedOptions, isFetching } = useGetOptions({
+  const {
+    options: calculatedOptions,
+    isFetching,
+    setData,
+    current,
+    currentStringy,
```
mikaelrss, line 31: I am a little confused by the name `currentStringy` :sweat_smile: Is it supposed to be the current option stringified? If so, I think a more verbose naming makes it easier to understand.

src/layout/Group/RepeatingGroupContainer.tsx
```diff
@@ -42,6 +42,7 @@ export function RepeatingGroupContainer({ containerDivRef }: IGroupProps): JSX.E
   const { textResourceBindings, id, edit, type } = node.item;
 
   const numRows = visibleRowIndexes.length;
+  const firstIndex = visibleRowIndexes[0];
```
mikaelrss, line 45: Is the lint-warning here correct? If this is unused, maybe we could remove it all together

mikaelrss (approved):
Great work! :raised_hands: I have a couple of minor comments/questions.

src/features/options/useGetOptions.ts
olemartinorg: Yup, sadly. Since the type is `IDataModelBindingsOptionsSimple | IDataModelBindingsSimple` and the `metadata` key does not exist at all in `IDataModelBindingsSimple`. I blame TypeScript for that one.

src/index.tsx
olemartinorg: Nope, it has been moved to `<PresentationComponent />`. That way it is always rendered _inside_ all the providers. It was rendered here as well, possibly because of a borked merge conflict resolution, so it was rendered twice.

src/layout/Group/RepeatingGroupContainer.tsx
olemartinorg: Strange. From SonarCloud? I don't always even read those warnings, they're often funky. It's being used in a `useEffect()` below, so absolutely needed.

mikaelrss: Sorry! I should have specified that i meant the entire function. Github does not allow me to place the comment anywhere without a diff :sweat_smile: 
image

olemartinorg: Ah, yes, I don't know - but I would think that's something @bjosttveit is removing if he doesn't need it in the validation rewrite.

src/layout/Checkboxes/CheckboxesContainerComponent.tsx
olemartinorg: Good point! I added some comments describing them. It's not always a string (for multiple selection components it's an array of strings), so that's why I named it 'stringy' not just 'string'.

Also, most of the point of this refactor was to make sure the source options, which are derived from the data model, actually was represented as string values even if the data model used other types (which it can now). I added an extensive unit test to make sure that's the case now for other configurations (such as static options in layout, along with options from API). Each one of them should now work correctly even if the data in them is a number or boolean. :partying_face: Thanks for nudging me to do that!

Issue #1500, opened 2023-09-25
Easing the transition to the Inter font 


olemartinorg:
When transitioning to the Inter font, we have received feedback that some fixed-width input fields no longer match the content the user is supposed to fill into these fields.

Examples often include:
- Post code / zip code
- Personal identity number
- Some numeric input fields for currency, with a suffix

We have already identified some options to explore:
- For advanced users in forms assumed to be displayed on larger screen sizes, we could implement full-width forms for some pages where more horizontal space is needed. See #1220 
- Options for reducing the gap/"luft" in between form components/tables. In other words, options for a more condensed view.
- Introducing options for allowing horizontal scrolling on displays that are narrower than the minimum needed to display the current form page in the "full size" configured/intended by the app developer (i.e., disabling responsive layout and proper mobile support).
- Consider other fonts that align closer to the text widths we observed in Altinn-DIN.

RonnyB71:
Assumes a font change only, not changing code or components.

olemartinorg:
Removing this from `fe-v4`, as we reverted the Inter font change there.

PR #1718, opened 2024-01-09
Switch back to `Altinn-DIN` font

bjosttveit:
## Description

Switch font back to `Altinn-DIN` until we can solve the issues that came up when switching to `Inter`.

## Related Issue(s)

- #1500


mikaelrss (approved):
:raised_hands:

Issue #1714, opened 2024-01-02
Generate JsonPatch instead of saving the full data model


olemartinorg:
### Description

In https://github.com/Altinn/app-lib-dotnet/issues/383 the backend will support a PATCH method when saving form data, and instead of us sending the full form data on every save, we should instead generate a JsonPatch (see https://jsonpatch.com/) containing the changes to the data model. Along with the changes, we should first add `test` operations to any changed paths to assert that the previous data was as expected. If the tests fail, the backend will not apply our changes, and we should instead fetch an updated data model and present that to the user (along with a message that tells the user that the data had apparently changed, and that they need to re-apply their own change on the new data set).

There are some npm libraries that can help us generate a JsonPatch from two models (by diffing them), but when I tried looking into the different libraries I failed to find one that looked stable, mature, maintained and suitable for us - so I suspect it's easier for us to just generate the patches ourselves (we do have a limited set of form data manipulation functions, and we could augment those to generate diffs, along with de-duplicating the diffs whenever we do rapid repeated changes - for example in `setLeafValue` when typing into a text field).

PR #1721, opened 2024-01-10
On-demand fetching data models used in text resource variables

olemartinorg:
## Description
Until now, data models used in variable references in text resources come from one of two places:
- If you're currently in a form (i.e. a `data` task), we'll let you access the current form data. It doesn't really matter which data model you specify, you can only use either `dataModel.default` or `dataModel.<the specific name of the current data model>`.
- If you're in a `confirm` or `feedback` process step, you could use any data model from any previous process step. We'd melt them all together into one frankenstein-datamodel, so as long as one of them referenced the correct model name you could use `dataModel.default` for the others (it would look up from the same place anyway).

With this PR, we instead:
- Look up in the correct data model. If in a form/data task, you can still use `dataModel.default`, but only for as long as we have the concept of a default data model (that may disappear when we introduce support for multiple data models per process step).
- In all other process steps (and for that matter also in a form step), even the receipt, we'll fetch the data model on the fly as soon as we see the text resource being used somewhere. This means we'll only fetch what we actually need, and you can reference any data model from anywhere. If the request to fetch the data model fails, we'll use the `defaultValue` instead. While fetching, you may see a brief `...` text where the value in the text resource should go.

## Related Issue(s)

- #1714 
  - Related because this flattening/melting together of data models was one of the last hold-outs before we can properly access the data models and use other types than just strings.

## Verification/QA

- Automated tests
   - There are tests for most of this already
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels


bjosttveit (approved):
# 💪

PR #1754, opened 2024-01-30
Fix erroneous schema validation errors for null/undefined values

### LINKED ISSUE #1752: Schema validation fails for `null`/`undefined` values
### Description of the bug

Schema validation returns an error if the field has a `format` constraint and the value of the field is `null` or `undefined`. I think the reason for this is that there was an assumption that empty fields will be removed entierly from the data model (key deleted). This may no longer be the case after all of the data saving changes(?). Fields that do not exist are not validated against the schema and so this used to work as it should. Schema validation should generally not give any errors if the field is empty, as this is `required` validation's job to handle. This should probably be solved by explicitly returning an empty validation array for schema validation if the field is empy, `null` or `undefined`.

- Slack: https://altinn.slack.com/archives/C02EJ9HKQA3/p1706277321354569?thread_ts=1706099835.186329&cid=C02EJ9HKQA3

### Steps To Reproduce

1. Set `"showValidations": ["Schema"]` on a field
2. Have the value of that field somehow be `null` or `undefined`

### Additional Information

_No response_

bjosttveit:
## Description

Fixes error where null/undefined values where evaluated with schema validation. This probably changed with the new data storage in v4, so fields that are null/undefined no longer get their key removed from the data model.

## Related Issue(s)

- closes #1752 

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels

### DIFF
--- a/src/features/validation/frontend/schemaValidation.test.ts
+++ b/src/features/validation/frontend/schemaValidation.test.ts
@@ -103,6 +103,9 @@ describe('schemaValidation', () => {
           { value: '2020-01-01', expected: true },
           { value: '1985-04-12T23:20:50.52Z', expected: false },
           { value: 'asdfasdf', expected: false },
+          { value: '', expected: true },
+          { value: null, expected: true },
+          { value: undefined, expected: true },
         ],
       },
       {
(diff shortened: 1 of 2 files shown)


olemartinorg (approved):
Looks fine! Technically, `undefined` is the same as a missing key. We can't serialize `undefined` in JSON, so `{ "someKey": undefined }` becomes `{}` when saved to the backend. `null`, however, is a supported value in JSON. We can, and probably do, store some `undefined` values in our temporary data model in the frontend. Often though, the backend will respond with a `null` value for things that are nullable in those cases, so I would think `undefined` turns into `null` over time.. :shrug:

PR #1778, opened 2024-02-03
Fix/1730 jaws likert issues

### LINKED ISSUE #1730: Tab order on likert component is wrong
### Description of the bug

Using a Jaws screen reader to answer questions using the likert component the focus is wrong when navigating back to the component from a validation message on a summary page (and possible when entering the component in the first place) making it very hard for users to actually complete the answer.

### Steps To Reproduce

1. On a page with a likert component, skip filling out the answer and navigate to the last page and try to submit
2. Click on the validation issue pointing back to the likert component which should be required
3. Observe that the focus is on the last radio button on the answer and not the headline and that is not possible to easily navigate back

### Additional Information
Potentially related to #1408

mikaelrss:
## Description

This PR fixes an issue with the JAWS screen reader and the `LikertComponent` as referenced in #1730. The issue states that when navigating to a `LikertComponent` by clicking an error message in the `ErrorReport` component, JAWS screen reader places the focus in the incorrect question/alternative and there is no way of navigating to the correct one. 

When I tried to reproduce this error, I found that the `ErrorReport` navigating to `LikertComponents` does not work at all (not just for screen readers), and it seems like this never has worked. So this PR fixes the issue by implementing support for navigating to `LikertComponents` with `ErrorReport`. 

To be able to fix this issue, I needed to be able to pass a `ref` to the `TableRow`-component from `@digdir/design-system-react`. This was not possible in our current version of the design system, so I had to upgrade it to at least `v0.46.0`. ~~This came with quite some errors, and would need further refactoring beyond this bug fix. I therefore decided that the most pragmatic way of solving the problem would be to import the latest version of the design system and keep the older version under an alias `@digdir/design-system-react-old`. I upgraded all uses of the design system to the newest version, except for a couple of files which used components that do no longer exist. These still import their components from `@digdir/design-system-react-old`. under the alias `@digdir/design-system-react-latest` and import `Table` components from this aliased version of the design system. Doing it the first way I described seems more reasonable, but this broke _a lot_ of cypress-tests that seemed to be unrelated to this issue. In the `main` branch, we would of course update the design system to the latest version, and not have duplicate versions of the package.~~ I upgraded the design system to the latest version, and imported the `LegacySelect` component and used this instead of the `Select` component.

I have tested the new behavior with both orca and JAWS, and the component seems to behave like it should. 

 ### Code Changes:
- Updated the types of the `render` function in `AnyComponent` to be able to pass a render function that uses `React.forwardRef`: 
```tsx
   abstract render:
    | ReturnType<typeof React.forwardRef<HTMLElement, PropsFromGenericComponent<Type>>>
    | ((props: PropsFromGenericComponent<Type>) => JSX.Element | null);
```
This means that all `render` functions have to be defined as an instance member property, as opposed to an instance member function i.e `render = () => {}` instead of `render(){}`
- Update all LayoutComponent's `render` functions to be arrow functions
- Update `LikertComponent` to use the latest `Table` components from `@digdir/design-system-react`. 
- Update `GridCopmonent` to use the latest `Table` components from `@digdir/design-system-react`. 
- Update `InstanceSelection` to use the latest `Table` components from `@digdir/design-system-react`. 
- ~~Update all Components that use older design system components to import from `@digdir/design-system-react-old`.~~ 

### Visual Changes: 
As the design on the latest `Table` components differ somewhat from the current table, I have tried to make some slight adaption/configuration of the visual appearance without overwriting the styles the design system provides out of the box.

The visual changes can be seen in a nice side-by-side view in Percy, but I provided a couple examples of visual changes here. 

#### Instance Selection

Original:
image

New: 
image
I added a border on this table, to make it feel as "boxy" as the old one did. The new border is quite a lot "harder" than the old one, and there was no configuration option to make a more subtle border.

#### Group/RepeatingGroups

Old:
image
New:
image
The new `Table` does not have a different background color on the `TableHead` row but instead a thicker border. Left the default styling here, as I think it works well. 

Old:
image
New:
image

Old:
image
New:
image
I added the border property for RepeatingGroupTables that are nested. I also added the a background color on the header `--fds-semantic-surface-action-second-subtle`. 

Old:
image
New:
image
Again, I added the same background color on the header `--fds-semantic-surface-action-second-subtle`. I also added a white background color (`--fds-semantic-surface-neutral-default`) to the rows when they were not in edit mode for nested `RepeatingGroupTable`. I also had to overwrite som of the default stylings of the table header, specifically removing `border-top-(left/right)-radius`, and adding a border-top equal to the border-bottom, as there can be multiple `TableHead` rows inside the same table.

#### Likert

Old:
image

New:
image
For the `Likert` component, I added the border property, as the old `Likert` seemed to have the same "boxy" feel as the `InstanceSelector`. 

What do you think of these changes @olavflar @Edavda? This will only be implemented in version `v3` as of right now, but we need to upgrade the design system on `v4` at some point as well. I think we would need to have a detailed look at how we would present all our different flavours of Table components :sweat_smile: 

## Related Issue(s)

- closes #1730 
- #1779 The "same" fix needs to be implemented for v4.

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels


test/e2e/integration/frontend-test/group.ts
```diff
@@ -460,9 +460,9 @@ describe('Group', () => {
     cy.get(appFrontend.group.subGroup).find(appFrontend.group.delete).click();
     cy.snapshot('group: delete-warning-popup');
 
-    cy.get(appFrontend.group.subGroup).find(appFrontend.group.popOverCancelButton).click();
+    cy.get(appFrontend.group.subGroup).find(appFrontend.group.popOverCancelButton).click({ force: true });
```
mikaelrss, line 463: I don't know why this failed. Cypress was complaining that the popover was clipped by overflow properties of one of its parents, but I could not see that this was the case. The popover was working just fine in the test and locally.

PR #1826, opened 2024-02-14
feat: add support for defining a custom receipt layoutset

### LINKED ISSUE #1468: Custom receipt should render even after reloading the page
### Description

Right now, when configuring a custom receipt layout, it is added to `settings.json` _within_ a layout-set. This creates a few problems:

- It is possible to define it multiple times, each subsequent layout-set could overwrite the previous one
- Frontend has to keep old layouts laying around in redux/state storage, for use when showing the receipt in a future process step
- Reloading a page when looking at the custom receipt does not load the `settings.json` file, so frontend is unable to know about the custom receipt that has been defined, and will simply not display it.

Instead, we should move the configuration for a custom receipt to `layout-sets.json` where it belongs. There we could configure a layout-set for the end-event in the process (or simply name it something like receipt), define the data type frontend should download the data for, and have a folder of one (or more!) layouts to display upon reaching the receipt page.

For this to work flawlessly, we should coordinate this with Studio and have them support this alternative configuration there before we release it in app-frontend.

### Additional Information

- https://altinn.slack.com/archives/C0534SVCJLD/p1694781993448529
- #859
- #823 
- #171

mikaelrss:
## Description
Adds support for defining a custom receipt. The receipt is defined as a layout set which is referred to in  `layout-sets.json` by creating a layout set which points to the custom layout set, and specifies the keyword task `CustomReceipt`:
```json 
{
  "id": "custom-receipt",
  "dataType": "fields",
  "tasks": [
    "CustomReceipt"
  ]
}
```

The custom receipt will not be show if the data model that is specified in the layout set is not available. This means, that if a receipt layout set points to a data model that has `autoDeleteOnProcessEnd`, we show the old receipt and write a warning in the dev console. The dev console does however not work for the simple receipt page at this moment. 

## Related Issue(s)

- closes #1468 

## Verification/QA

- Automated tests
  - [x] Has been added/updated
  [PR with documentation changes](https://github.com/Altinn/altinn-studio-docs/pull/1409)
  
- Support in Altinn Studio
  - [x] Issue(s) created for support in Studio https://github.com/Altinn/altinn-studio/issues/12308
- Sprint board
- Labels


src/components/wrappers/ProcessWrapper.tsx
```diff
@@ -113,21 +122,30 @@ export const ProcessWrapper = () => {
...
     return (
-      <PresentationComponent type={taskType}>
+      <PresentationComponent type={realTaskType}>
+        <ReceiptContainer />
+      </PresentationComponent>
+    );
+  }
+
+  if (taskType === ProcessTaskType.Data && customReceiptDataModelNotFound) {
+    window.logWarn('You specified a custom receipt, but the data model is missing. Falling back to default receipt.');
```
bjosttveit, line 140: You should use `window.logWarnOnce` since this is called inside of a render function. It will potentially log many times, and there is a risk of causing an infinite render loop if the log somehow causes a rerender. It causes state in the dev-tools to update after all. It is debounced, so it should not cause a freeze, but still

mikaelrss: Aha! Thanks! I'll fix this right away!

bjosttveit (approved):
## Looks good 🚀

PR #1829, opened 2024-02-15
Fix options loading and clearing

### LINKED ISSUE #1556: Summary/PDF - rendered before dynamic options are loaded

bjosttveit:
## Description

1. `useAllOptions` will now wait until a node is added to the state through `setNodesFound` before it starts to fetch options. This makes sure that it will not think everything is loaded when the first option is loaded.
2. Clearing options did not work as expected. Since it checked the current (from the options-list) instead of the currentStringy, the value would always be undefined if the string-value was not present in the options list, and so it would not see that it needed clearing. Added a test to make sure it does not clear the value while fetching new options in case the value is still valid in the new option list.

## Related Issue(s)

- closes #1556 

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels

### DIFF
--- a/src/features/options/useAllOptions.tsx
+++ b/src/features/options/useAllOptions.tsx
@@ -143,7 +143,8 @@ export function AllOptionsProvider({ children }: PropsWithChildren) {
   const dummies = nodes
     ?.allNodes()
     .filter((n) => isNodeOptionBased(n))
-    .filter((n) => Object.keys(allOptions).includes(n.item.id))
+    // Until allInitiallyLoaded is true, we want to wait for nodesFound to be set before we start fetching options
+    .filter((n) => allInitiallyLoaded || Object.keys(allOptions).includes(n.item.id))
     .map((node) => (
       <DummyOptionsSaver
         key={node.item.id}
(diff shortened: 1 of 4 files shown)


olemartinorg (approved):
:raised_hands:

PR #1835, opened 2024-02-19
Include empty fields in mapping as empty query parameters

bjosttveit:
## Description

In v3, a `mapping` from an empty field would add the queryParameter `?field=undefined` which is literally sending the value as the string `"undefined"`. If the query parameter is used for filtering on some property this sometimes worked as expected, as long as the filtered property never actually was equal the string `"undefined"`.

Up until now in v4, we have instead ignored and simply not added the query parameter if its value was empty, which is less wrong, and arguably correct. However, now you cannot separate the cases where a mapping parameter is added but empty, or if it is not added to the mapping at all. Since setting the mapping parameter is an active choice you may want a different behavior in the two cases.

This change makes sure to add all mapping parameters to the query string, but treat `null`/`undefined` as empty strings instead of literal `"null"`, `"undefined"` strings.

## Related Issue(s)

- closes #{issue number}

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels

### DIFF
--- a/src/features/formData/FormDataWrite.tsx
+++ b/src/features/formData/FormDataWrite.tsx
@@ -354,7 +354,6 @@ export const FD = {
    *      in that sense.
    *   2. The data is fetched from the debounced model, not the fresh/current one. That ensures queries that are
    *      generated from this hook are more stable, and aren't re-fetched on every keystroke.
-   *   3. Values that don't exist in the debounced model are not included in the output at all.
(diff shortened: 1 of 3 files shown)


olemartinorg (approved):
Nice! :rocket:

PR #1836, opened 2024-02-19
fix: make all layout componenst support receiving refs

mikaelrss:
## Description
Fixes an issue with a warning: `Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?`. I thought I had fixed this, but it needs to be implemented for all `LayoutComponents`. This is because all of our components potentially receive a `ref` if the `directRender` function retruns `true`. 

```tsx
  if (layoutComponent.directRender(componentProps) || overrideDisplay?.directRender) {
    return (
      <FormComponentContextProvider value={formComponentContext}>
        <RenderComponent
          {...componentProps}
          ref={containerDivRef}
        />
      </FormComponentContextProvider>
    );
  }

```

I refactored the code such that all `LayoutComponents` use `forwardRef` an such can receive refs even though they might not use the ref. One issue with this is that our lint rules now complain about the render function not having a display name. I've fixed this simply by disabling the lint-rule for these cases:
```tsx

export class Alert extends AlertDef {
  // eslint-disable-next-line react/display-name
  render = forwardRef<HTMLElement, PropsFromGenericComponent<'Alert'>>((props): JSX.Element | null => (
    <AlertComponent {...props} />
  ));
}

```

This could also be solved by placing the component outside of the class, and giving it a displayName there:
```tsx
const RenderComponent = forwardRef<HTMLElement, PropsFromGenericComponent<'Alert'>>((props): JSX.Element | null => (
  <AlertComponent {...props} />
));
RenderComponent.displayName = 'LayoutComponentAlertRender';

export class Alert extends AlertDef {
  render = RenderComponent;
}
```

What do you think we should go for? In option 1, the code looks a little cleaner, but the component becomes `Anonymous` in dev tools. However this is not a huge issue, as the render "component" ever only renders the equivalent react component. In the case of `Alert`, the hierarchy would be `Anonymous` -> `AlertComponent` vs `LayoutComponentAlertRender` -> `AlertComponent` in the react dev tools. 

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels

### DIFF
--- a/src/layout/Accordion/index.tsx
+++ b/src/layout/Accordion/index.tsx
@@ -11,10 +11,11 @@ import type { ComponentHierarchyGenerator } from 'src/utils/layout/HierarchyGene
 export class Accordion extends AccordionDef {
   private _hierarchyGenerator = new AccordionHierarchyGenerator();
 
-  // eslint-disable-next-line react/display-name
-  render = forwardRef<HTMLElement, PropsFromGenericComponent<'Accordion'>>((props, _): React.JSX.Element | null => (
-    <AccordionComponent {...props} />
-  ));
+  render = forwardRef<HTMLElement, PropsFromGenericComponent<'Accordion'>>(
+    function LayoutComponentAccordionRender(props, _): React.JSX.Element | null {
+      return <AccordionComponent {...props} />;
+    },
+  );
 
   hierarchyGenerator(): ComponentHierarchyGenerator<'Accordion'> {
     return this._hierarchyGenerator;
--- a/src/layout/Accordion/index.tsx
+++ b/src/layout/Accordion/index.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { forwardRef } from 'react';
 
 import { Accordion as AccordionComponent } from 'src/layout/Accordion/Accordion';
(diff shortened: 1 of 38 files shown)


olemartinorg (approved):
Looks good! :raised_hands: I have no real opinions on the rest - code looks good as-is, I think, the rest isn't so important that we can't fix it later if we're annoyed by the missing display name.

Magnusrm:
As a third option I think this also may work for setting `displayName`? 🤔 

```
export class Alert extends AlertDef {
  render = forwardRef<HTMLElement, 
  PropsFromGenericComponent<'Alert'>>(function LayoutComponentAlertRender(props): JSX.Element | null => (
    <AlertComponent {...props} />
  ));
}
```

mikaelrss:
> As a third option I think this also may work for setting `displayName`? 🤔
> 
> ```
> export class Alert extends AlertDef {
>   render = forwardRef<HTMLElement, PropsFromGenericComponent<'Alert'>>(function LayoutComponentAlertRender(props): JSX.Element | null => (
>     <AlertComponent {...props} />
>   ));
> }
> ```

If this works, I'll change it! :raised_hands:

Issue #1776, opened 2024-02-02
Use `@xsdType` hint in JsonSchema when converting types before sending to backend


olemartinorg:
Our current `convertData()` function was implemented based on a few (wrong) assumptions, for example that `integer` in JsonSchema maps to `int`/`int32` in C#, and that `number` maps to `decimal`.

Studio maps types a little differently when generating models, and we should account for that. See this clarification:
- https://altinndevops.slack.com/archives/CDU1S3NLW/p1706875045501569

Also, search for `@xsdType` in existing schemas for good examples.

bjosttveit:
Ran a quick search of `@xsdType` over the apps from [altinn-fetch-apps](https://github.com/olemartinorg/altinn-fetch-apps), which of these needs to be checked? @olemartinorg @mirkoSekulic 

```bash
❯ awk '{if ($1 ~ "xsdType") {if ($2 == ":") {print $3} else {print $2}}}' **/*.schema.json | sort | uniq -c | sort
   1 "Integer",
   1 "short",
   2 "XmlSimpleContentExtension",
   4 "XmlAttribute",
   6 "int",
  14 "nonNegativeInteger",
  15 "#ref"
  17 "dateTime",
  22 "gYear",
  44 "token",
  52 "date",
  57 "decimal",
 144 "double",
 155 "long",
 171 "#ref",
 195 "positiveInteger",
 355 "boolean",
 674 "XmlSimpleContentExtension"
 694 "integer",
 991 "XmlAttribute"
4997 "string",

```

olemartinorg:
Seems like a bit of a mess, but we'd need to figure out a mapping from those `@xsdType` values to their corresponding generated types in C#. Do you have such a mapping somewhere in Studio we can look at for inspiration, @mirkoSekulic? I would think this mostly counts for numeric types, but we should also account for booleans, and possibly dates.

I would assume that the `XmlAttribute`, `XmlSimpleContentExtension` and `#ref` values could safely be ignored (i.ie, if not a valid numeric/boolean/date type, assume the C# type is string).

olemartinorg:
@mirkoSekulic Did you get a chance to look at this?

olemartinorg:
Rapport fra testing:
> int (-2147483648 til 2147483647)  og short (-32768 til 32767) gir feil på rette grenseverdier

PR #1865, opened 2024-02-23
Validation for "invalid data" that cannot be saved + refactoring

### LINKED ISSUE #1749: Invalid data in data model should show up as validation errors to the user
In the v4-rc1, if you type something into a field that doesn't validate according to the JsonSchema basic data type (such as typing something other than `true` or `false` in a field bound to a `boolean` in the data model, or typing `-` or a very large number in a `number` or `integer` field) the data will just not be saved anywhere, and you'll lose that data without any warning. In v3, it would cause a backend error, as we'd pass a string that wasn't possible to deserialize.

Before releasing v4 stable, we should implement a new validation class for these problems. This sounds like it should be possible via the current `JsonSchema` validation, but it's not - it's not really that the values are not valid according to JsonSchema, it's errors where data cannot be cast to their respective types (it's impossible to cast `-` to a number without turning it into `-0`, but `-` is the start of any negative number). The same goes for very large (positive or negative) numbers, which we assume won't be possible to cast to `decimal` or `int32` in C#, and therefore are marked as invalid.

As this data cannot be saved at all until valid, it breaks with the model that refreshing the page should re-introduce all the same validation messages.

Right now, we assume that all fields marked as `numeric` in JsonSchema maps to `decimal` in C#, and all fields marked as `integer` in JsonSchema maps to `int`/`int32` in C#, along with their respective range of valid values. We might want to support using a hint in JsonSchema to map to the real type used in the C# model instead.

See these links for C# data types:
- https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/integral-numeric-types
- https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/floating-point-numeric-types

### Related issue(s)
- https://github.com/Altinn/app-lib-dotnet/issues/197
- https://github.com/Altinn/app-lib-dotnet/issues/134
- https://altinndevops.slack.com/archives/C045EB3JA9X/p1706629447618599?thread_ts=1706627814.812689&cid=C045EB3JA9X

### LINKED ISSUE #1847: Refreshing expression validation configuration
The current configuration for expression validation works as-is, but is suboptimal after the more recent changes to how single field validation and backend validation works (on save, no more `*FIXED*`). Before releasing the feature fully, and releasing the documentation for it, we should look over the configuration format and adjust it to more closely match (and work like) the newer validation logic works.

### LINKED ISSUE #1848: Backend validation groups are "forgotten" if not included in every request
### Description of the bug

The `useBackendValidation` no longer keeps track of the validation groups through incremental validation. It now only checks the initialValidations and the last validated validationgroups causing it to forget previous incremental validations if they are not included in each request.

### Steps To Reproduce

1. Implement a custom validator in the backend with a `HasRelevantChanges` implementation that only runs when a specific field changes.
2. Notice that if any other field is changed, any validations from this group disappear.

### Additional Information

_No response_

bjosttveit:
## Description

- This adds support for validating the contents of the `invalidData` object to give feedback to the user when data is not able to be stored in the strongly typed models in the app backend. https://github.com/Altinn/app-frontend-react/issues/1776 Adds proper support for more data types and should work with this.
- Fixes backend validation being broken by forgetting previous incremental validations
- Expression validations are now always run, meaning that there should not be any "field dependency" issues.
- I refactored most of the validation logic to align more with the structures introduced in invalid data validation. In my opinion the validation logic is much simpler and easier to read. And most likely less error-prone. Possibly at some performance cost, however it is hard to say exactly, as the checks it previously had to do to avoid unecessary validation before was a bit expensive.
- This refactor resolves a potential edge-case from the Datepicker format validation conflict. Schema validation should not be much less efficient, as the entire schema had to be validated before as well.

## Related Issue(s)

- closes #1749 
- closes #1848 
- closes #1847 

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels


src/features/validation/invalidDataValidation/useInvalidDataValidation.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export function useInvalidDataValidation(): FieldValidations {
+  const invalidData = FD.useInvalidDebounced();
+
+  return useMemo(() => {
+    if (Object.keys(invalidData).length === 0) {
+      return __default__;
+    }
+
+    return Object.keys(dot.dot(invalidData)).reduce((validations, field) => {
```
olemartinorg, line 19: Hmm, I would think it's possible for the invalidData object to a "shadow" of the full data model after a while, with empty objects, etc. It might be necessary to clean up/ignore such values before adding validation messages for everything in `dot.dot(invalidData)`?

olemartinorg (approved):
Fantastic!

olemartinorg:
> Expression validations are now always run, meaning that there should not be any "field dependency" issues.

Does this mean that this also fixes #1847?

bjosttveit:
> Does this mean that this also fixes #1847?

Well yeah, it is now a more naive implementation that runs much more often, so that should not be an issue now.

src/features/validation/invalidDataValidation/useInvalidDataValidation.ts
olemartinorg: Yeah, I'm not 100% sure either. If you look at the result of `dot.dot(invalidData)` and it always only contains stringy values with invalid data (and not `null`, `{}`, etc), this isn't a problem after all. In older times we had the `flattenData()` function that did a `dot.dot()`, but also removed some values (such as empty objects), but it's worth noting that `dot.dot()` and that old `flattenData()` does not produce the same result (I've made that blunder a few times).

bjosttveit: Looks like empty objects will stay behind when removing nested properties, so I need some filtering I guess

Issue #1812, opened 2024-02-09
Unique IDs in repeating groups


olemartinorg:
### Introduction
After the JsonPatch change, and the new form data engine in v4, combined with `openByDefault` actually creating objects for automatically added rows, it's becoming obvious that we need some better way to uniquely differentiate between repeating groups rows than the indexes alone.

After a meeting the 9th of February 2024, with participants from Apps and Studio, we landed on a proposed solution where Studio generates partial classes in data models (and the CLI tool to upgrade existing apps rewrites the models to partial), while the backend uses a source generator to extend the classes with our own metadata/values. This should allow for unique IDs to be added, serialized and stored in Storage, and kept all the way through form filling.
```[tasklist]
### Tasks
- [ ] https://github.com/Altinn/app-lib-dotnet/issues/435
- [ ] https://github.com/Altinn/altinn-studio-cli/issues/23
- [ ] https://github.com/Altinn/altinn-studio/issues/12277
- [ ] https://github.com/Altinn/app-frontend-react/issues/1750
- [ ] https://github.com/Altinn/app-lib-dotnet/pull/441
```

PR #1870, opened 2024-02-27
Unique row IDs for repeating groups

### LINKED ISSUE #1750: JsonPatch: Better support for uniquely identifying rows in repeating groups
With the new JsonPatch saving method, we still allow users to make changes to the data model in the time between when we send a save request to the backend and the point at which we receive a reply from the backend with the full (and possibly changed) data model. To make this work, we compare the data we got from the backend with the current data we have on the frontend, and we try to make a decision based upon some rules:

1. If a value such as a `string`, `number` or `boolean` changed in the backend while it also changed on the frontend (i.e. the user made a change to it) we overwrite the changes the backend did - which will also schedule a new save request for later. If the backend did not make the changes, but another (or the same) user made a change outside of the current app-frontend session, the save request will fail when the backends expectations in the `test` operation fails in a later request - so we can only possibly overwrite (perceived stale) data received from the backend such as those made automatically by `ProcessDataWrite` and automatic population of `null` or default values from the typed data model.
2. If there are changes in a repeating group (i.e. `Object[]`), we employ a LCS (Longest Common Subsequence) algorithm using the `fast-array-diff` package, along with our own custom comparison function in order to figure out if a row (object in the array) looks like they represent the same row as the one we already had locally, or if the backend meant to remove one row and add another.

It's the second part here that is currently very brittle and sub-optimal. With more details in a [lengthy thread on Slack](https://altinndevops.slack.com/archives/C045EB3JA9X/p1706104091281899), we're hoping to implement a way to add metadata for these objects to uniquely identify them and differentiate them. In frontend then, we should be able to:

1. Reduce the complexity of our custom `isSimilarEnough()` function by comparing the unique IDs
2. Add code to automatically generate unique IDs for new rows, and for those rows we get from the backend that are missing unique IDs (which we should assume are new rows in `isSimilarEnough()`).
3. Remove the code in the `group.ts` Cypress test that introduces waiting until rows are saved before continuing the tests. Running this test on github + tt02 should reveal if the code fixes the core issue, as these tests failed often before these workarounds.

olemartinorg:
## Description

The main headline here is that app-frontend now uses (and requires) the `altinnRowId` property in repeating group rows, as generated automatically by either frontend (when rows are added by the user) or backend (when rows are added by custom code, prefill, etc). These row IDs are now used when generating a JsonPatch so that we're 100% sure which row needed a change (regardless of the row index).

This also includes:
- Using the row UUID in internal state for which group row we're editing. This means the backend can sort the rows, and that app-frontend will keep the correct row open for editing after.
- Some code cleanup, moving component definitions out from other components (`ExtraRows` in `RepeatingGroupTable`)
- A few improvements to `ttd/frontend-test`
  - Adding a new page for 'Pets' that can be used to test large repeating groups (with lots of rows), along with sorting rows. The page is intentionally ugly to point out some styling problems in our apps. :wink: 
  - Using a gateway that makes it possible to jump straight to a given task, skipping the previous ones. This should make it much easier to use `ttd/frontend-test` when working locally (I'm sick of filling out the changename form just to get to the repeating groups).
- Adding support for CustomButton responses returning validation issues (which it now does when the data model is updated).
- Showing a custom error page when the backend did not give us row IDs. This page should nudge app developers into reading the docs and running the migration tool (`altinn-studio upgrade ...`)
- Adding a `cy.waitUntilSaved()` function that will wait until the next save request has finished, and the page is no longer marked as having unsaved changes.
- Alerting the user if you try to close the page (or navigate away) when there are unsaved changes in the form.
- Updating tests, especially shared expression tests
- Removing previous workarounds in the `frontend-test/group.ts` Cypress test, as JsonPatch should now work fully with repeating groups.
- Minimum version of nuget packages are now v8.0.0-rc1

## Related Issue(s)

- closes #1750
- #1812

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
    - https://github.com/Altinn/altinn-studio/issues/12277
- Sprint board
- Labels


olemartinorg:
Some tests are failing now, and it seems to be a problem with v8-rc1. Thread here: https://altinndevops.slack.com/archives/C045EB3JA9X/p1709039345902389

olemartinorg:
I skipped the failing tests. They should be possible to run again after this is merged, released, and `ttd/frontend-test` is deployed with the new version:
- https://github.com/Altinn/app-lib-dotnet/pull/462

This should be ready for review now. :+1:

src/features/devtools/DevToolsControls.tsx
```diff
@@ -34,9 +36,9 @@ export const DevToolsControls = () => {
       <Tabs.List className={classes.tabList}>
         <Tabs.Tab value={DevToolsTab.General}>{DevToolsTab.General}</Tabs.Tab>
         <Tabs.Tab value={DevToolsTab.Logs}>{DevToolsTab.Logs}</Tabs.Tab>
-        <Tabs.Tab value={DevToolsTab.Layout}>{DevToolsTab.Layout}</Tabs.Tab>
-        <Tabs.Tab value={DevToolsTab.Components}>{DevToolsTab.Components}</Tabs.Tab>
-        <Tabs.Tab value={DevToolsTab.Expressions}>{DevToolsTab.Expressions}</Tabs.Tab>
+        {isInForm && <Tabs.Tab value={DevToolsTab.Layout}>{DevToolsTab.Layout}</Tabs.Tab>}
+        {isInForm && <Tabs.Tab value={DevToolsTab.Components}>{DevToolsTab.Components}</Tabs.Tab>}
+        {isInForm && <Tabs.Tab value={DevToolsTab.Expressions}>{DevToolsTab.Expressions}</Tabs.Tab>}
```
bjosttveit, line 41: 🤯

src/features/devtools/components/DevToolsLogs/DevToolsLogs.tsx
```diff
@@ -103,12 +103,31 @@ export const DevToolsLogs = () => {
...
-                <>
-                  {line}
-                  <br />
-                </>
-              ))}
+              {log.message.split('\n').map((line) => {
+                // Find URLs and make them clickable as links
+                const urlRegex = /(https?:\/\/[^\s]+)/g;
+                const split = line.split(urlRegex);
+                return (
```
bjosttveit, line 110: 😍

src/features/devtools/DevToolsControls.tsx
olemartinorg: Ahh, yes, forgot to mention that in the description, but DevTools should work outside of form-tasks now, with this change. It was very much needed when I added a custom error page when row IDs are not provided by the backend, and you need to run `altinn-studio` CLI to fix it - and the details about that should only be visible to developers, thus logging it at the same time as the (more generic) error page is shown was the best way I found to ensure that. Still, very useless when devtools crashed there.. :sweat_smile:

bjosttveit (approved):
## Excellent 🥇 

Thats a lot of uuid 😅 

bilde

src/features/datamodel/useBindingSchema.tsx
```diff
@@ -37,7 +37,7 @@ export function useCurrentDataModelGuid() {
   return getCurrentTaskDataElementId({ application, instance, taskId, layoutSets });
 }
 
-export function useCurrentDataModelUrl() {
+export function useCurrentDataModelUrl(includeRowIds: boolean) {
```
ivarne, line 40: Do you ever set includeRowIds to false here? I assumed it was only added for backwards (and schema) compatibility.

ivarne: Currently the backend will just generate new random guids (on every fetch) in that case, so no real need to load with `includeRowId=false`.

Do you think that is a reasonable behaviour for backend, or should I be more restrictive and only generate random guids when the element is unlocked?

olemartinorg: Naah, that works, but also - as long as we don't need those UUIDs, I don't see why we should fetch them. Making sure our e2e tests work _without_ them when referenced in text resources is a strength in and of itself. It's a lot more difficult to take something away in the future than never providing it in the first place, and if app developers go building on those random UUIDs it will just cause confusion anyway.

PR #1874, opened 2024-02-27
Multiple performance enhancements

### LINKED ISSUE #1867: Find and fix simpler performance issues in v4
Test-cases:
- `rbi-all-things`
- https://altinndevops.slack.com/archives/C045EB3JA9X/p1709110533852109
- The `Pets` page in `ttd/frontend-test` with a large repeating group (with `Dropdown` components in the table)
- General re-rendering on component updates

bjosttveit:
## Description

List of changes:
- Faster duplicate options filtering
- Caching calls to `getParsedLanguageFromText()` to avoid having to parse HTML to react for already parsed strings
- Making sure `useQueryWithStaleData()` does not force rendering twice (used for fetching text resources)
- Implementing more nifty selectors in zustand contexts, such as `useDelayedMemoSelector()`. More about this below. (Not quite sure why @FTLems got credit for that commit, though.. :shrug:) 
- Re-using old node objects when creating new ones and the old ones are identical. This, together with passing node IDs from `form.tsx` to `GenericComponent` instead of objects, and a selector to get the given node to render, this avoid re-rendering some top-level components even when all nodes are refreshed/changed.
- Rewritten multiple contexts to use zustand, along with more targeted selectors for selecting the states/functions you need. If contexts rely on `useEffect()` etc to update their state, these effects have now been put into isolated components rendered beside the rest of the DOM tree, so that they don't affect their children. Rewritten contexts include:
  - Validation
  - Nodes/ExprContext
  - LayoutValidation
  - Attachments
  - PageNavigationContext
- Moving some effects and generated state out from Form.tsx to a state-updating childless child component, causing the Form component to only re-render when any of that state changes (not when anything it relies on possibly changes, such as nodes being re-generated).
- Minor optimization when calling `node.isHidden()`, where traversal is limited when called multiple times (that function is called very often some places).
- Fixing problems where outdated form data models could be passed into the save function, making us save already outdated data.
- Now that everything is faster, some bugs also came bubbling up to the surface, such as form data being saved multiple times in a row with an identical generated patch (which would reliably crash with a 409).
- Lots more, minor fixes and speedups, and probably also some things that did not result in any better performance but seemed sensible at the time.

### Wait, `useLaxDelayedMemoSelectorFactory()`? What?!
One of the problems that have caused _lots_ of unnecessary re-rendering, is that we've had a tendency to create these huge 'data sources' objects for looking up data that we _might_ need. This has been instrumental when working with redux in v3 and migrating towards the code we have now in v4, but piling more and more state into these objects comes with an obvious downside; if any of the pieces change, everything has to re-render. This is part of the reason the node hierarchy (for example) is re-created all the time.

In an effort to quickly get better performance, and instead of turning everything upside down in order to get rid of these data sources objects properly, the 'delayed memo selector' was born. For some background, we already had the concept of 'memo selectors' in zustand contexts, which lets you select some data, process it a bit, and the hook would make sure your component only has to re-render if that selected data (after processing) is `!deepEqual(...)` to the previous value.

A 'delayed memo selector' is just a function you get, that you can call to get some data. So, everywhere we passed around the `formData` object before, we can now pass around a `formDataSelector` function that lets you simply select a given path to immediately get the data for that path. The big difference, of course, is that your component would now only have to re-render _if the data that was previously selected using the selector was changed_. So, in principle, the node hierarchy now only has to change when _one or more paths in the form data which has been used inside expressions_ actually change (as opposed to _every time anything in the form data changes_, as it's been previously). The same goes for other state that was previously dumped into data sources, but performs much better when being selected out, such as option lists, and to a limited extent the list of hidden components.

## Related Issue(s)

- closes #1867 

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels


src/language/languages.ts
```diff
@@ -4,17 +4,24 @@
...
-      return nn();
-    default:
-      return nb();
+  if (cachedLanguages[languageCode]) {
+    return cachedLanguages[languageCode];
+  }
+
+  const langFunc = langFuncMap[languageCode];
+  if (langFunc) {
+    const language = langFunc();
```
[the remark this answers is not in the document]
olemartinorg: Oooh, veldig bra spotta! :star_struck: Potensielt mulig [kalle en av disse](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object#instance_methods) via langCode. Vanskelig å utnytte til noe, men fortsatt interessant! Skal fikse det.

olemartinorg (approved):
Read through all the code, and found no debug statements or out-of-place code I forgot about. Approving my own PR, but leaving it open so we can go through the code later together if needed.

src/components/form/Form.tsx
```diff
@@ -139,3 +155,62 @@ function useSetExpandedWidth() {
...
+ */
+function ErrorProcessing({ setFormState }: ErrorProcessingProps) {
+  const currentPageId = useCurrentView();
+  const topLevelNodeIds = useNodesMemoSelector(
+    (nodes) =>
+      nodes
+        .findLayout(currentPageId)
+        ?.children()
+        .map((n) => n.item.id) || emptyArray,
+  );
```
Magnusrm, line 176: Is there a reason the empty array is stored in a const and only referenced? Is it because we dont want to create a new empty array every time?

olemartinorg: Yup, correct. Because `[] !== []`, every time it renders the empty array would be a different one. When looking it over once more, I'm not sure it does much here, but I usually do that as a safer default when in doubt.

PR #1879, opened 2024-02-29
Feat/1104 store selected codelist value in datamodel

adamhaeger:
## Description

Adds the ability to save chosen labels in option inputs as a list of strings in the datamodel.

This PR is a work-in-progress, we might have found a bug in the PATCH logic.

## Related Issue(s)

- closes #{issue number}

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels


src/codegen/Common.ts
```diff
@@ -128,7 +128,7 @@ const common = {
 
   // Data model bindings:
   IDataModelBindingsSimple: () =>
-    new CG.obj(new CG.prop('simpleBinding', new CG.str()))
+    new CG.obj(new CG.prop('simpleBinding', new CG.str()), new CG.prop('label', new CG.str().optional()))
```
olemartinorg, line 131: Not all components with a simple binding should have the option to use a `label` binding (for example, it does not make sense for an `Input` component), so this should just be set on the `IDataModelBindingsOptionsSimple` type.

test/e2e/integration/frontend-test/dynamics.ts
```diff
@@ -1,10 +1,23 @@
...
+  cy.goto('changename');
+  cy.get(appFrontend.changeOfName.newFirstName).type('geir');
+  cy.get(appFrontend.changeOfName.newLastName).type('hoel');
+  cy.get(appFrontend.changeOfName.confirmChangeName)
+    .findByRole('checkbox', {
+      name: /Ja[a-z, ]*/,
+    })
+    .check();
+  cy.get(appFrontend.changeOfName.reasons).should('be.visible');
+};
```
olemartinorg, line 19: I don't see the need to extract this one into a function, unless that function is used multiple places. It just adds one level of indirection.

olemartinorg (approved):
Code looks good! :rocket: We should look into those failing Cypress tests.
- I think the `all-process-steps` one fails because you changed the data model in `ttd/frontend-test`, but didn't update that test to match it. The easiest way I've found is to run the test locally until it fails, then click on the error, open up developer tools in the browser, then compare the models with something like jsondiff.com
- The other failing test (`hide-row-in-group.ts`) looks a bit stranger. It might be that the currently selected value may have been affected by this change to `useGetOptions()`. I can help out with this one.

For now, btw, I've rolled back the latest deploy of `ttd/frontend-test` as it fails on other pull requests - just so you know. Whenever you push code to be tested on github/tt02 again, just re-deploy version 157 via Studio.

olemartinorg:
One more thing! According to the Percy snapshots, there are quite a few errors in the DevTools log now. We should take a look at those (unless you already know what they're about) and see if we can avoid printing them out (or fix the frontend-test app if there's anything wrong in it).

test/e2e/integration/frontend-test/dynamics.ts
adamhaeger: Agree, used it previously, but not needed after all :)

olemartinorg:
Tested now, and this is working wonderfully! :raised_hands: I'm still missing layout validation for the bindings, but I created an issue for that here:
- https://github.com/Altinn/app-frontend-react/issues/1894

PR #1935, opened 2024-03-14
Give certain schema validations `Required` visibility, and filter duplicates

bjosttveit:
## Description

Give `minItems` errors from schema validation the same visibility as required. Make sure to filter out duplicate errors if both `minItems` is defined in schema and `minCount` is defined on the repeating group.

I added a tab in the node inspector to more easily observe the validation state of nodes. This will show validations for each dataModel binding, which type of validation it is, if it is visible and or filtered out as a duplicate. It also shows which validation types are visible for the node at the top (Schema, Component, CustomBackend,  etc.).

**A small digression**
I also explored the possibility of doing the same with required validations from the data model, as it would be nice to show these even if the component is not marked as required. Required validations in schema are currently ignored in the frontend but can occur in the backend. There were two main problems with this:
1. In JSON schema, required is actually defined on the parent object, not the field itself, so we would need to add a special case when figuring out what field the error should map to. This is trivial to solve, as the error object returned from AJV contains this information.
2. Data models often have their fields deeply nested, and whether or not a field is required is defined in the parent object. If the parent object that contains the definition of what is required does not exist, then the missing field is not really an error in the JSON schema world. This can easily occur since intermediate objects are not created before the datamodel is written to.
3. (Even more digression here, by @olemartinorg) Also, the form data context will remove properties when clearing fields, as the backend may define non-obvious default values, which will only be set when the property is removed (i.e. not when the field is set to empty string, `0` anything else - technically setting `null` may even cause problems if the field cannot be `null` according to the backend). We remove the property to avoid dealing with this, and the backend will respond with the default value for the field immediately when saving.
4. (Yet more digressions by @olemartinorg). Lastly, required on a property in JsonSchema is not the same as our required-validation for form fields. We produce en error if the value is "empty" (i.e. empty string), but technically an empty string is still a value and thus fully valid (and the requirement is fulfilled) according to JsonSchema.

## Related Issue(s)

- https://altinn.slack.com/archives/C02EJ9HKQA3/p1710332770814879?thread_ts=1710320934.677599&cid=C02EJ9HKQA3

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels

### DIFF
--- a/src/features/validation/schemaValidation/schemaValidationUtils.ts
+++ b/src/features/validation/schemaValidation/schemaValidationUtils.ts
@@ -5,6 +5,9 @@ import addAdditionalFormats from 'ajv-formats-draft2019';
 import type { ErrorObject, Options } from 'ajv';
 import type { JSONSchema7 } from 'json-schema';
 
+import { ValidationMask } from '..';
+import type { ValidationCategory } from '..';
+
 /**
  * Create a new ajv validator for a given schema.
  */
@@ -159,4 +162,17 @@ export function getErrorParams(error: ErrorObject): string | null {
   }
   const errorParams = error.params[errorType.paramKey];
   return Array.isArray(errorParams) ? errorParams.join(', ') : errorParams;
+}
+
+/**
+ * Get the category of an error object.
+ */
+export function getErrorCategory(error: ErrorObject): ValidationCategory {
+  switch (error.keyword) {
+    case 'required':
+    case 'minItems':
+      return ValidationMask.Required;
+    default:
+      return ValidationMask.Schema;
+  }
 }
--- a/src/features/validation/schemaValidation/schemaValidationUtils.ts
+++ b/src/features/validation/schemaValidation/schemaValidationUtils.ts
@@ -5,8 +5,8 @@ import addAdditionalFormats from 'ajv-formats-draft2019';
 import type { ErrorObject, Options } from 'ajv';
 import type { JSONSchema7 } from 'json-schema';
 
-import { ValidationMask } from '..';
-import type { ValidationCategory } from '..';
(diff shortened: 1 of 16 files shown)


src/features/validation/schemaValidation/schemaValidationUtils.ts
```diff
@@ -5,6 +5,9 @@ import addAdditionalFormats from 'ajv-formats-draft2019';
 import type { ErrorObject, Options } from 'ajv';
 import type { JSONSchema7 } from 'json-schema';
 
+import { ValidationMask } from '..';
+import type { ValidationCategory } from '..';
```
olemartinorg, line 9: Strange! Our eslint rules should rewrite these to absolute paths, but that apparently doesn't work with `..` :shrug: Whatever, I don't mind.

src/layout/RepeatingGroup/index.tsx
```diff
@@ -93,6 +93,22 @@ export class RepeatingGroup extends RepeatingGroupDef implements ValidateCompone
     return validations;
   }
 
+  /**
+   * Repeating group has its own minCount property, so if set, we should filter out the minItems validation from schema.
+   */
+  schemaMinItemsFilter(validation: BaseValidation): boolean {
```
olemartinorg, line 99: This one could probably be `private`. I guess we don't expect it to be used outside the class itself?

olemartinorg (approved):
Great stuff! :raised_hands:

src/layout/RepeatingGroup/index.tsx
bjosttveit: good point

PR #1948, opened 2024-03-17
feat: Add function for creating links to pages/components in the form…

### LINKED ISSUE #1648: Text: Mulighet for å legge til ankerpunkt/lenke i tekst til sider i skjema
### Description

Vi har skjema der man må referere til hva utfyller her fylt ut på foregående side, på en side lengre ut, og hvor man da ber utfyller om å besvare et eller flere spørsmål i forhold til hva en er svar før. 
image

Det er ikke noen god måte å gjøre dette i skjema i dag, eneste er via summering. Man kunne trengt en mulighet for å legge til lenke til sider tilbake i skjema inn i tekst elementer slik at en bruker kan klikke seg tilbake til sidene, dette vil også være bedre med tanke på UU.

### Additional Information

_No response_

mikaelrss:
## Description
This PR creates to new dynamic expression functions:
- `linkToPage` - This function can be used to create links that point to specific pages in the form that can be placed inside text in the form. 
- `linkToComponent` - This function can be used to create links that point to a specific component in the form.

To place such a link inside a big paragraph right now, would require the use of the `concat` function, but I see this as a good first implementation until we are able to provide dynamic expressions inside resource files. 

## Related Issue(s)

- closes #1648 

## Verification/QA

- Automated tests
  - [x] Has been added/updated
https://github.com/Altinn/altinn-studio-docs/pull/1471
  
- Support in Altinn Studio
  
- Sprint board
- Labels


src/layout/GenericComponent.tsx
```diff
@@ -88,6 +89,27 @@ function ActualGenericComponent<Type extends CompTypes = CompTypes>({
...
+      maybeInput.focus();
+    }
+  }, []);
+
+  React.useEffect(() => {
+    if (nodeId === node.item.id) {
+      containerDivRef.current?.scrollIntoView();
+      focusInput();
+      clearSearchParam(SearchParams.FocusNodeId);
+    }
```
olemartinorg, line 111: This only works if the node is visible, so it wouldn't work if the node is in a repeating group. Look into `useNavigateToNode()` instead, which will open the group for editing, etc, before setting focus on it.

Also, with this method, the query parameter would linger in the URL until the node is suddenly visible, so you're not even guaranteed it will get focus when the page renders. Let's say you link to page `A` and nodeId `inPageB`. With this code the parameter would stay in the URL until you navigate to page `B` manually, at which point the `inPageB` node will suddenly (and unexpectedly) get focus.

mikaelrss: That's a very good point! I'll have to think about a better way to solve this :thinking:

olemartinorg: Yup, look into `useNavigateToNode()` and how to re-use that, which already solves the issue. Also, it should make sure to clear the param regardless of the focus request actually succeeding.

mikaelrss: I solved this by creating two different functions `linkToPage` and `linkToComponent`. Then, using `useNavigateToNode` solved the issues for navigating to a specific component.

src/layout/GenericComponent.tsx
```diff
@@ -89,6 +90,18 @@ function ActualGenericComponent<Type extends CompTypes = CompTypes>({
...
+  const componentId = searchParams.get(SearchParams.FocusComponentId);
+  const focusNode = useResolvedNode(componentId);
+  const navigateTo = useNavigateToNode();
+
+  React.useEffect(() => {
+    if (focusNode != null) {
+      navigateTo(focusNode);
+    }
+    clearSearchParam(SearchParams.FocusComponentId);
+  }, [componentId, clearSearchParam, navigateTo, focusNode]);
```
olemartinorg, line 103: Why does every component have to do this? :thinking: Doesn't that mean that when rendering the page, N components (possibly hundreds) will call `navigateTo(...)` and clear the search params? I would think this functionality is better placed in `Form.tsx`, as it's a peculiarity of a form, not for each component.

olemartinorg (approved):
Looks very good! :raised_hands:

src/layout/GenericComponent.tsx
mikaelrss: You're right! I'll see if I get it to work in Form.tsx

src/components/form/Form.tsx
```diff
@@ -31,6 +31,11 @@ interface FormState {
 export function Form() {
   const currentPageId = useCurrentView();
   const { isValidPageId, navigateToPage } = useNavigatePage();
+  const { searchParams, clearSearchParam } = useNavigationParams();
+  const componentId = searchParams.get(SearchParams.FocusComponentId);
+  const focusNode = useResolvedNode(componentId);
+  const navigateTo = useNavigateToNode();
+  const location = useLocation();
```
olemartinorg, line 38: Did you check if this made `<Form />` re-render more often? I know I struggled a bit to reduce the amount of hooks this component uses in order to minimize re-renders, as re-rendering this might trigger re-rendering of all components (which is costly).

olemartinorg: It could also be possible to do this in a component like the `<ErrorProcessing />` component (declared below here), which doesn't render any children below it, so it won't affect performance of `<Form />` when only it has to re-render.

src/layout/GenericComponent.tsx
```diff
@@ -112,6 +114,7 @@ function ActualGenericComponent<Type extends CompTypes = CompTypes>({
     if (targetNode.item.id !== id) {
       return undefined;
     }
+    clearSearchParam(SearchParams.FocusComponentId);
```
olemartinorg, line 117: This might not be necessary if it happens in the `<Form />` component as well? :thinking:

olemartinorg (approved):

src/layout/GenericComponent.tsx
mikaelrss: You're right. There was some timing issues with providing the `clearSearchParam` function from the `useNavigationParams` hook. Removed the function and cleared the parameters directly in `HandleNavigationFocusComponent` instead.

olemartinorg (approved):
🙌

src/components/form/Form.tsx
olemartinorg: Yep, creating a new component for this was what I meant - I see now that my suggestion didn't make that entirely clear.. 🙈

PR #2039, opened 2024-04-23
Add function select to dynamic expressions

mikaelrss:
## Description

_This is a feature request, and I would like some feedback as to whether you think this should be implemented or not._

This PR introduces a new function `_experimentalSelectAndMap` to the dynamic expression language. The function selects string properties in objects of an array in the data model and concatenates them to a single string. The reason for this is to be able to create a list of something in a form's data model. 

Usage: 

Say you have this underlying data structure: 
```json 
{
  "path": {
    "to": {
      "arrayOfObjects": [
        {
          "name": "Prop1",
          "value": "Value1"
        },
        {
          "name": "Prop2",
          "value": "Value2"
        },
        {
          "name": "Prop3",
          "value": "Value3"
        }
      ]
    }
  }
}
```

```json 
["_experimentalSelectAndMap", "path.to.arrayOfObjects", "name"]
``` 
Would result in the string `Prop1 Prop2 Prop3`

The function accepts two optional parameters that are string constants to prepend and append to each item in the list. 

```json 
["_experimentalSelectAndMap", "path.to.arrayOfObjects", "name", "<li>", "</li>"]
``` 
Would result in the string `<li>Prop1</li> <li>Prop2</li> <li>Prop3</li>`

Appending to the last element in the list is optional but on by default. 

```json 
["_experimentalSelectAndMap", "path.to.arrayOfObjects", "name", "", ",", false]
``` 
Would result in the string `Prop1, Prop2, Prop3`

A typical use case for this would be to show a list of names from some data structure. The need arose in Digital gravferdsmelding, where there is a need to be able to list deceased people who already have somebody to take care of their funeral. 

image

I realize I can probably already achieve this by creating the output string on the backend and show the string in clear text in the form. However, I think it would be nice to have this kind of function as this would allow service owners to implement something like this in Altinn Studio without having to write C# code. 

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
  - [x] I will do that later/have created an issue - This feature will be undocumented as it is not intended to be removed in the future. It will be replaced by a more comprehensive list operation workflow in the dynamic expression language. 
  
- Support in Altinn Studio
  
- Sprint board
- Labels

### DIFF
--- a/schemas/json/layout/expression.schema.v1.json
+++ b/schemas/json/layout/expression.schema.v1.json
@@ -77,6 +77,7 @@
         { "$ref": "#/definitions/func-language" },
         { "$ref": "#/definitions/func-lowerCase" },
         { "$ref": "#/definitions/func-upperCase" },
+        { "$ref": "#/definitions/func-_experimentalSelectAndMap" },
         { "$ref": "#/definitions/func-argv"}
       ]
     },
@@ -465,6 +466,20 @@
       "items": [
         { "const":  "upperCase" },
         { "$ref": "#/definitions/string" }
+      ],
+      "additionalItems": false
+    },
+    "func-_experimentalSelectAndMap": {
+      "title": "Experimental Select and map function",
+      "description": "This function takes a data model path which points to an array in the data model, and selects a specific property from the object in the containing array. It then concatenates the selected values into a single string separated by a space. The values can be prepended and appended with a string constant. This function is experimental and will be removed in future versions. Do not use unless you know what you're doing.",
+      "type": "array",
+      "items": [
+        { "const":  "_experimentalSelectAndMap" },
+        { "$ref": "#/definitions/string" },
+        { "$ref": "#/definitions/string" },
+        { "$ref": "#/definitions/string" },
+        { "$ref": "#/definitions/string" },
+        { "$ref": "#/definitions/boolean" }
       ],
       "additionalItems": false
     },
--- a/src/features/expressions/index.ts
+++ b/src/features/expressions/index.ts
@@ -760,6 +760,30 @@ export const ExprFunctions = {
     args: [ExprVal.String] as const,
     returns: ExprVal.String,
   }),
+  _experimentalSelectAndMap: defineFunc({
+    args: [ExprVal.String, ExprVal.String, ExprVal.String, ExprVal.String, ExprVal.Boolean] as const,
+    impl(path, propertyToSelect, prepend, append, appendToLastElement = true) {
+      if (path === null || propertyToSelect == null) {
+        throw new ExprRuntimeError(this, `Cannot lookup dataModel null`);
+      }
+      const array = this.dataSources.formDataSelector(path);
+      if (typeof array != 'object' || !Array.isArray(array)) {
+        return '';
+      }
+      return array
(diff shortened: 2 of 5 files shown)


bjosttveit:
Have you thought about what the function should look like with multiple data models?

For example, the dataModel expression will have this syntax for referencing different data models explicitly: `["dataModel", "path.to.field", "dataType"]`

mikaelrss:
> Have you thought about what the function should look like with multiple data models?
> 
> For example, the dataModel expression will have this syntax for referencing different data models explicitly: `["dataModel", "path.to.field", "dataType"]`

Ah, no I did not think of this. But if the `dataModel` function is going to have an extra field to specify which data model to read from, then I guess the `select` function would need some way to do this as well. Is the explicit `dataType` argument required in the new function? If so, we could just add it to `select` as the second argument as well, and require 3 args instead of 2.

bjosttveit:
> Is the explicit `dataType` argument required in the new function? If so, we could just add it to `select` as the second argument as well, and require 3 args instead of 2.

For the `dataModel` expression the data type is an optional parameter to maintain backwards compatibility. Not specifying the data type will use the one defined in `layout-sets` as before. But in this case, since the function does a lot and there are already many optional args, it will get very cluttered. Maybe this could be split into simpler expressions?

mikaelrss:
> > Is the explicit `dataType` argument required in the new function? If so, we could just add it to `select` as the second argument as well, and require 3 args instead of 2.
> 
> For the `dataModel` expression the data type is an optional parameter to maintain backwards compatibility. Not specifying the data type will use the one defined in `layout-sets` as before. But in this case, since the function does a lot and there are already many optional args, it will get very cluttered. Maybe this could be split into multiple simpler expressions?
> 
> **Edit**: What you suggested in [slack](https://altinndevops.slack.com/archives/C045EB3JA9X/p1713940741117039?thread_ts=1713938589.049099&cid=C045EB3JA9X) was exactly what I was thinking 😅

I think something like this would be the best solution. `["map", ["pluck", "[path.to](http://path.to/).dataModel", "name"], ["concat", "<li>", ["argv", 0], "</li>]]` but as @olemartinorg pointed out. This would be a larger job to implement. I don't think we have the time to spare to implement something like this right now. But since my `select` (which is going to be renamed) function is a temporary solution, do you still think it is necessary to implement support for multiple data models? Could we get by with only operating on the data model of the current layout-set or are there any issues with this when we start to support multiple data models?

mikaelrss:
> For a "temporary" thing like this, that we do not want to support in the future, but rather replace with something more general. Could we maybe leave this as an undocumented feature, and require a feature flag for it to work? Maybe also a warning in the schema that this is an experimental feature not to be used in prod by anyone else?

I think this is a great solution! Do we need to put the functionality behind a feature flag if we're naming it `_experimentalSelectAndMap`? :sweat_smile:

bjosttveit:
> I think this is a great solution! Do we need to put the functionality behind a feature flag if we're naming it `_experimentalSelectAndMap`? 😅

No probably not, I was brainstorming a little 😆

bjosttveit (approved):
Very good

PS: I added `ignore-for-release` as well as I dont think we should advertise this as a new feature

PR #2043, opened 2024-04-24
Feat/1736 payment

adamhaeger:
## Description

This PR adds the frontend support for payment.

## Related Issue(s)

- closes #{1736}

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels


test/e2e/integration/payment-test/payment.ts
```diff
@@ -0,0 +1,105 @@
...
+import path from 'path';
+
+import { AppFrontend } from 'test/e2e/pageobjects/app-frontend';
+const appFrontend = new AppFrontend();
+
+let paymentUrl: string;
+
+describe('Payment', () => {
+  beforeEach(() => {
+    cy.startAppInstance(appFrontend.apps.paymentTest, { authenticationLevel: '2' });
```
olemartinorg, line 13: Not sure how the authentication level will work on tt02, but it seems like the test app is not yet deployed there? :thinking: That should be working before we can merge this, otherwise we'll have constantly failing tests.

test/e2e/integration/payment-test/payment.ts
```diff
@@ -0,0 +1,105 @@
...
+  });
+
+  it('Should fill out the form, landing on the payment page', () => {
+    cy.intercept({
+      method: 'GET',
+      url: '**/ttd/payment-test/*',
+      query: {
+        paymentid: '*',
+      },
```
olemartinorg, line 22: Doesn't this just catch all requests for the entire app? I'd assume that using a url like `**/payment` is preferable and less flaky.

olemartinorg: Or maybe this request isn't for payment at all? Are we requesting the app itself and expecting a redirect? :thinking: I'm not sure I'm 'getting' this, but I would think it could be more specific.

test/e2e/integration/payment-test/payment.ts
```diff
@@ -0,0 +1,105 @@
...
+      method: 'GET',
+      url: '**/ttd/payment-test/instances/**/**/payment',
+    }).as('paymentInfoRequest');
+
+    cy.wait('@paymentInfoRequest').then((_) => {
+      cy.url().then((url) => {
+        cy.log(`Current URL is: ${url}`);
+        paymentUrl = url;
+      });
+    });
```
olemartinorg, line 41: So this test is storing a URL in a local variable that is then used by the next test? This would break if using `it.only()`. I think if the tests depend on each other, they're not really testing two different things, so you could just keep them in the same `it()`.

src/layout/Payment/SkeletonLoader/SkeletonLoader.tsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+
+import { Heading, Skeleton } from '@digdir/designsystemet-react';
+
+import classes from 'src/layout/Payment/SkeletonLoader/SkeletonLoader.module.css';
+
+export const SkeletonLoader = () => (
```
olemartinorg, line 7: Nice! Maybe this should replace our own `Loader.tsx`? Or maybe that could be used instead of this? It's not clear to me what's the difference between the two. Anyway, this should possibly not be tied to the payment component, and it might be something we'd want to move to `src/core/loading/`?

src/layout/Payment/SkeletonLoader/SkeletonLoader.tsx
```diff
@@ -0,0 +1,29 @@
...
+    <div className={classes.titleContainer}>
+      <Skeleton.Circle
+        width='30px'
+        height='30px'
+      />
+      <Heading
+        asChild
+        size='medium'
+      >
+        <Skeleton.Text>En medium tittel</Skeleton.Text>
```
olemartinorg, line 22: Is this hard-coded text supposed to be here?

src/layout/PaymentDetails/PaymentDetailsComponent.tsx
```diff
@@ -0,0 +1,33 @@
...
+  const { partyId, instanceGuid } = useInstanceIdParams();
+  const { title, description } = node.item.textResourceBindings || {};
+  const hasUnsavedChanges = FD.useHasUnsavedChanges();
+  const queryClient = useQueryClient();
+  const { data: orderDetails } = useOrderDetailsQuery(partyId, instanceGuid);
+
+  useEffect(() => {
+    if (!hasUnsavedChanges) {
+      queryClient.invalidateQueries({ queryKey: ['fetchOrderDetails'] });
+    }
```
olemartinorg, line 23: Seems like the backend should notify us when order details change as a result of form data changes instead. Invaliding the request and fetching it again after each save operation seems excessive.

At some point, if we want to provide an expression function for getting payment details, we'd have to move this effect to work on all form pages in every app, and then we'd just double the amount of requests we send back and forth on form saving.

src/layout/Payment/queries/usePaymentInformationQuery.tsx
```diff
@@ -0,0 +1,23 @@
...
+  const { fetchPaymentInformation } = useAppQueries();
+  const { next } = useProcessNavigation() || {};
+  return useQuery({
+    queryKey: ['fetchPaymentInfo', partyId, instanceGuid],
+    queryFn: async () => {
+      if (partyId && instanceGuid) {
+        const result = await fetchPaymentInformation(partyId, instanceGuid);
+        if (result?.status === PaymentStatus.Paid) {
+          next && next({ action: 'confirm', nodeId: 'next-button' });
+        }
```
olemartinorg, line 17: Yikes - a get query that sometimes fires a mutation effect.. So if someone tries to render the `Payment` component with `renderAsSummary: true` in any other task type, this will end up calling `process/next` for them automatically? I would move this out from the query function and have that as an effect in the `Payment` component.

src/layout/Payment/config.ts
```diff
@@ -0,0 +1,54 @@
...
+  .addTextResource(
+    new CG.trb({
+      name: 'description',
+      title: 'Description',
+      description: 'Description, optionally shown below the title',
+    }),
+  )
+  .addProperty(
+    new CG.prop(
+      'paymentSettings',
```
olemartinorg, line 42: Settings on the Payment component are, by their definition, 'payment settings' - why the need to require an object here? I would just have this one setting on the root of the Payment component.

src/layout/Payment/config.ts
```diff
@@ -0,0 +1,54 @@
...
+    new CG.prop(
+      'renderAsSummary',
+      new CG.expr(ExprVal.Boolean)
+        .optional({ default: false })
+        .setTitle('Render as summary')
+        .setDescription(
+          'Boolean value or expression indicating if the component should be rendered as a summary. Defaults to false.',
+        ),
+    ),
+  )
```
olemartinorg, line 25: Don't define this yourself, extend `SummarizableComponentProps` instead (you can do that by calling `Config.inner.extends(CG.common('SummarizableComponentProps'))`).

In the branch I'm working on I've changed that type, because it claims that `renderAsSummary` can be an expression - but support for that was never been implemented, and nobody has asked about it, so apparently nobody ever needed for it to support expressions anyway.

olemartinorg (approved):
Solid piece of work here!! :godmode: I've still got quite a few comments, and I think it's important the tests passes on tt02 before we merge this.

src/layout/PaymentDetails/PaymentDetailsComponent.tsx
adamhaeger: How would backend notify us though? Theres no websockets/SSE stuff right?

olemartinorg: Nah, but when we're saving form data (i.e. a data model, with the PATCH endpoint) we already get a list of current validation messages from the backend. This, of course, has the effect of collecting pretty much all functionality into the same endpoint - but I think it's better that the backend at least hints that cache should be invalidated than just always invalidating the cache every time we save.

test/e2e/integration/payment-test/payment.ts
adamhaeger: We need this intercept because cypress has a hard time following the redirect back to our app. I've added a comment to explain what happens.

PR #2072, opened 2024-05-03
feat: added new summary component with config, can render child nodes

adamhaeger:
## Related Issue(s)

- closes #{issue number}

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels


src/layout/Summary2/CommonSummaryComponents/MultipleValueSummary.tsx
```diff
@@ -0,0 +1,79 @@
...
+  displayData,
+  showAsList,
+}: MultipleValueSummaryProps) => (
+  <div className={classes.checkboxSummaryItem}>
+    <div className={cn(classes.labelValueWrapper, { [classes.error]: errors.length > 0 })}>
+      <Label weight={'regular'}>{title}</Label>
+      {displayData && showAsList && (
+        <List.Root>
+          <List.Unordered>
+            {displayData?.split(',').map((item) => (
```
olemartinorg, line 33: I'm pretty sure splitting display values by comma will break. Introducing a code list with values such as `Stavanger, Rogaland` would split up into two list items when it was supposed to only be one. I would suggest taking the raw option values (that can be split by commas) and get the option labels for each of them, and then pass that as an array.

src/layout/Summary2/config.tsx
```diff
@@ -0,0 +1,67 @@
...
+        new CG.union(
+          INPUT_SUMMARY_PROPS,
+          CHECKBOX_SUMMARY_PROPS,
+          RADIO_SUMMARY_PROPS,
+          DROPDOWN_SUMMARY_PROPS,
+          MULTIPLE_SELECT_SUMMARY_PROPS,
+        ),
+      ).optional(),
+    ),
+  );
```
olemartinorg, line 67: Ahh, now I get why you have `componentId` in overrides. How about making this a map instead, where the key is the component ID? That would make for faster lookups and it would be impossible to mess things up by adding multiple overrides for the same component. Also, nitfy having a union for all of these - but does it work? I.e. do you only get possible options in vscode for an Input component when the type has narrowed down to that type?

src/layout/Summary2/config.tsx
```diff
@@ -0,0 +1,67 @@
...
+            .setTitle('Mode')
+            .setDescription('Config for what should be rendered'),
+        ),
+        new CG.prop('id', new CG.str()),
+      ),
+    ),
+  )
+  .addProperty(
+    new CG.prop(
+      'overWriteProperties',
```
olemartinorg, line 56: How about just calling this `overrides`? This property is overly long and complex for something we expect app developers to use a lot - and it's not effectively overwriting anything. Also, they're all properties, thats implicit - but by using 'properties' in the name it may look like you can override any property in the component configuration (which you can't).

src/layout/Summary2/config.tsx
```diff
@@ -0,0 +1,67 @@
...
+      'children',
+      new CG.arr(new CG.str())
+        .optional()
+        .setTitle('Children')
+        .setDescription('Array of component IDs that should be displayed in the summary'),
+    ).onlyIn(Variant.External),
+  )
+  .addProperty(
+    new CG.prop(
+      'whatToRender',
```
olemartinorg, line 41: "Rendering" is a term we use, but which app developers don't. For them, I think it makes more sense to call this property `target`.

src/layout/Summary2/config.tsx
```diff
@@ -0,0 +1,67 @@
...
+    ).onlyIn(Variant.External),
+  )
+  .addProperty(new CG.prop('childComponents', new CG.arr(CG.layoutNode)).onlyIn(Variant.Internal))
+  .addProperty(
+    new CG.prop(
+      'children',
+      new CG.arr(new CG.str())
+        .optional()
+        .setTitle('Children')
+        .setDescription('Array of component IDs that should be displayed in the summary'),
```
olemartinorg, line 36: I would love to get a clearer description about how this works as opposed to `whatToRender`.. Also, if it's meant to be used together with `whatToRender` - how does that play out? This is what I would expect to see in a description, so I don't have to try it out or just guess when configuring this component.

src/layout/Summary2/CommonSummaryComponents/SingleValueSummary.tsx
```diff
@@ -0,0 +1,53 @@
...
+      <Paragraph
+        asChild
+        className={cn({
+          [classes.error]: errors.length > 0,
+          [classes.emptyValue]: !displayData,
+          [classes.formValue]: displayData,
+        })}
+      >
+        <span>
+          {!displayData && <Lang id={'general.empty_summary'}></Lang>}
```
olemartinorg, line 39:
```suggestion
          {!displayData && <Lang id={'general.empty_summary'} />}
```

I'm sure prettier or eslint does this for us automatically? 🤔 I saw this at least one other place as well.

src/utils/layout/NodesContext.tsx
```diff
@@ -27,7 +27,7 @@ interface NodesContext {
   setHiddenComponents: (mutator: (hidden: Set<string>) => Set<string>) => void;
 }
 
-function initialCreateStore() {
+export function initialCreateStore() {
```
olemartinorg, line 30: There's probably no need to export this one. I found no imports either, but now it indicates that it's being used somewhere else.

src/layout/Summary2/SummaryComponent2/SummaryComponent2.tsx
```diff
@@ -0,0 +1,63 @@
...
+    return (
+      <ResolveComponent
+        summaryProps={summaryNode.item}
+        summaryOverrides={summaryNode.item.overWriteProperties}
+      />
+    );
+  }
+
+  if (summaryNode.item.whatToRender.type === 'task') {
+    const IDSplitted = summaryNode.item.whatToRender.id.split('>');
```
olemartinorg, line 43: Oh, this should probably be documented in the JsonSchema/generated types. 🙌
But I'm not sure it works entirely as expected. The way I read it, if you want to reference a component in another task, that component will be found no matter the pageId you're referencing.

src/layout/Summary2/SummaryComponent2/TaskSummary.tsx
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  return (
+    <div style={{ width: '100%' }}>
+      {pageKeys.map((page) => (
+        <div
+          style={{ marginBottom: '10px' }}
+          key={page}
+        >
+          <TaskSummaryAccordion pageKey={page}>
```
olemartinorg, line 67: Nice! But are we sure we want to group pages in accordions by default? For some apps that might make sense, but those that don't use the NavigationBar component they don't really have an incentive to add translations for the page keys. I'd make this optional and add an accordion config for those who want it.

src/layout/Summary2/SummaryComponent2/TaskSummary.tsx
```diff
@@ -0,0 +1,113 @@
...
+  return (
+    <Accordion
+      border
+      color={'neutral'}
+    >
+      <Accordion.Item
+        key={pageKey}
+        open={isOpen}
+      >
+        <Accordion.Header onHeaderClick={() => setIsOpen(!isOpen)}>{langAsString(pageKey)}</Accordion.Header>
```
olemartinorg, line 33: This could use <Lang /> instead of langAsString()

src/layout/Summary2/SummaryComponent2/ComponentSummary.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+  const childComponents =
+    componentNode.item.type === 'Group' &&
+    componentNode.item.childComponents.map((child) => (
+      <ComponentSummary
+        componentNode={child}
+        key={child.item.id}
+        summaryOverrides={summaryOverrides}
+      />
+    ));
```
olemartinorg, line 37: Groups are far from the only components with children nowadays. Instead of reaching directly into the node state, why not use node.children()?

olemartinorg (commented):
A few console.log() calls, commented-out code and unused locals, but I assume that will be cleaned up. 🙌

src/layout/Checkboxes/config.ts
```diff
@@ -25,7 +46,8 @@ export const Config = new CG.component({
         .setTitle('Alert on change')
         .setDescription('Boolean value indicating if the component should alert on uncheck'),
     ),
-  );
+  )
+  .addProperty(new CG.prop('summaryProps', CHECKBOX_SUMMARY_PROPS).onlyIn(Variant.Internal));
```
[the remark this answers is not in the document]
adamhaeger: See your point, need a little discussion to find a good solution :)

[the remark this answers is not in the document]
olemartinorg: Yep - these were my initial thoughts, I guess. :sweat_smile: I guess in hindsight (when reading the rest of the code) I think these could just be exported directly - but it doesn't make sense adding them as an internal property, as they never end up in `node.item` (which our types now say they do). It makes sense to only use these as an external type in the Summary2 component, but define them here in each separate component (for which their config belongs).

I've also mentioned earlier that I'd love to see some validation implemented for this (see #648), so that DevTools spits out errors/warnings if you configure overrides that are not possible for the target component.

src/layout/Dropdown/config.ts
```diff
@@ -1,7 +1,21 @@
...
+import { CG, Variant } from 'src/codegen/CG';
 import { ExprVal } from 'src/features/expressions/types';
 import { CompCategory } from 'src/layout/common';
 
+export const DROPDOWN_SUMMARY_PROPS = new CG.obj(
+  new CG.prop(
+    'hidden',
+    new CG.bool()
+      .optional()
+      .setTitle('Hidden')
```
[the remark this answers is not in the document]
adamhaeger: Moved to common :)

src/layout/Summary2/CommonSummaryComponents/MultipleValueSummary.tsx
adamhaeger: But wont the data be cleaned and processed and turned into a reliable comma separated list by the time it reaches a summary component? Also, if we do it the way you suggest, how will we know what raw option values are selected when we only get the display data as a string?

src/layout/Summary2/SummaryComponent2/TaskSummary.tsx
adamhaeger: The message from Edvin is that they should always render in accordion, but Ill ask him about this :)

src/layout/Summary2/config.tsx
adamhaeger: This is leftovers, we only use `whatToRender` :)

src/layout/Summary2/config.tsx
adamhaeger: Agree :)

src/layout/Summary2/config.tsx
adamhaeger: Agree :)

src/layout/Summary2/taskIdStore.tsx
```diff
@@ -0,0 +1,57 @@
+import React, { createContext, useContext } from 'react';
+
+import { create } from 'zustand';
+
+interface TaskState {
+  overriddenTaskId?: string;
+  overriddenDataModelId?: string;
```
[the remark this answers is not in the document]
adamhaeger: sure :)

src/layout/Summary2/SummaryComponent2/ComponentSummary.tsx
adamhaeger: Yes, this needs work, will look at it tomorrow :)

src/layout/Summary2/CommonSummaryComponents/MultipleValueSummary.tsx
olemartinorg: > But wont the data be cleaned and processed and turned into a reliable comma separated list by the time it reaches a summary component?

"Displaydata"/"displayValue" is a human-readable string, so when you have options like:

```
[
  { label: "Stavanger, Rogaland", value: 1103 },
  { label: "Sandnes, Rogaland", value: 1102 }
]
```

And you check both options in a Checkboxes component, the 'form data' (what is actually stored in the data model) will be `1103,1102`. But the "display data" value is much less strictly defined. It doesn't really have to follow any rules, but I would assume the display data here would be `Stavanger, Rogaland, Sandnes, Rogaland` (but yes, that's also hard for humans to read). I've seen app developers try to cram long strings with commas in them as option _values_, so I can all but guarantee there are apps with commas in option _labels_.

> Also, if we do it the way you suggest, how will we know what raw option values are selected when we only get the display data as a string?

What I'm saying is _do not get the display data as a string_ :wink: That hook is not meant for this use-case, so you need to make something more appropriate if you want to display each selected option as a list - `displayValue`.

src/layout/Summary2/SummaryComponent2/TaskSummary.tsx
Edavda: @olemartinorg Can apps that don't use the NavigationBar still have multiple pages, with their own respective titles?

The idea for having each page in it's own accordion is to mirror the structure of the form on the summary page, and to separate these pages more clearly with some visual element. I'm thinking that if an app doesn't show the navigation, it will still have the user going through a series of pages to fill out the form, and the accordions on the summary page will reflect this and make it a little easier to see what content they are reviewing.

olemartinorg: > Can apps that don't use the NavigationBar still have multiple pages[...]

Yes!

> [...] with their own respective titles?

Maybe. When adding a page, it doesn't have a natural title. The title of the _app_ shows up on all pages, but translating the page _key_ to something user-readable is optional (and I suspect it's rarely done).

> The idea for having each page in it's own accordion is to mirror the structure of the form on the summary page

Does that mean that pointing a `Summary2` to the current layout-set (aka "the form you're in now") also splits it up into one-accordion-per-page? I haven't seen any apps doing that using the old `Summary` component, and I would be hesitant to force apps into that pattern (rather have it be opt-in rather than opt-out).

Designsystemets recommendations are to [show one question per page](https://www.designsystemet.no/monstre/obligatoriske-og-valgfrie-felt#1-ett-sp%C3%B8rsm%C3%A5l-per-side), and if that turns into _one answer per accordion_, we're pushing app developers away from our own recommendations, as it would make sense to cram more things into fewer pages.

> I'm thinking that if an app doesn't show the navigation, it will still have the user going through a series of pages to fill out the form, and the accordions on the summary page will reflect this and make it a little easier to see what content they are reviewing.

My opinion is that a 'summary page' is , at the same time, also an 'overview page' to show you everything you've filled out before submitting, and hiding that content behind a clickable (assuming the accordions are closed by default?) just makes it more difficult to get that overview, IMO. It's always easier to _see the content without having to click anything_, and if we require one click per page to get that overview, you'll need just as many clicks as just browsing through the pages in a `NavigationBar`. So what's the point in having "one summary page" if that page consists of many smaller "accordion pages"?

src/layout/Summary2/SummaryComponent2/SummaryComponent2.tsx
adamhaeger: We have gone for a more intuitive way of configuring this, have a look :)

src/layout/Summary2/SummaryComponent2/TaskSummary.tsx
adamhaeger: Just an case detail: they are open by default

adamhaeger: I suppose the need for the accordions arose from apps that has tons of questions on their pages, which a lot of real world apps have. Case in point is the 1008 app from Finanstilsynet which has so much data and questions it would be insane to have one question per page. I really like the recommendation to have one question per page though and in that case I agree it would be strange to show pages in accordions.

So first of all the accordions are open by default. In the case of 1008 the accordions would allow you to easily go through the data on each page and open and close accordions as it suits you.

So how about we simpy make accordions opt-in? If you want accordions you would soon find that you need to add titles just like you would when adding a NavigationBar.

Edavda: I think having the accordions open by default will make sure its still easy to get an overview of the summary. 

But these are some valid points and I agree it sounds like a better idea to then have the accordion layout as an opt-in. It would definitely be more valuable for the larger and more content-heavy forms, but I think you're right about forcing an accordion setup mid-form is a bad idea. 

Good input!

src/layout/Address/AddressSummary/AddressSummary.tsx
```diff
@@ -0,0 +1,75 @@
...
+          componentNode={componentNode}
+        />
+
+        <SingleValueSummary
+          title={<Lang id={careOfTitle || 'address_component.care_of'} />}
+          displayData={careOf}
+          errors={errors}
+          componentNode={componentNode}
+          hideEditButton={true}
+        />
```
olemartinorg, line 39: This should probably also be conditionally rendered using `!componentNode.item.simplified && ...`

src/layout/Group/GroupSummary.tsx
```diff
@@ -0,0 +1,86 @@
...
+      >
+        {title}
+      </Heading>
+      <Paragraph
+        spacing
+        className={cn(classes.description)}
+      >
+        {description}
+      </Paragraph>
+      {renderChildComponents(componentNode, hierarchyLevel, summaryOverrides)}
```
olemartinorg, line 83: I would prefer making this a real React component instead of a plain function that returns JSX. :raised_hands:

src/layout/Input/InputSummary.tsx
```diff
@@ -4,19 +4,30 @@ import { Lang } from 'src/features/language/Lang';
...
-  summaryOverrides?: CompInputInternal['summaryProps'];
+  summaryOverrides?: InputSummaryOverrideProps;
 };
 
-export const InputSummary = ({ componentNode, displayData, summaryOverrides }: InputComponentSummaryProps) => {
+/*
+ * Hidden: Dersom hidden = true fjernes hele node fra hierarkiet.
+ * hvis vi da rendrer summary, vil fjernes også være fjernet fra summary
+ * Hvis vi da vil vise komponent kun på summary, er det vanskelig å få til.
+ *
```
olemartinorg, line 20: Dette er vel ikke riktig? Noder fjernes ikke fra hierarkiet selv om de er skjult. Du kan sjekke om noe er skjult via `node.isHidden()`, men noder som er skjult vil fortsatt hentes ut fra hierarkiet. Derfor er det ikke noe problem i seg selv å "vise noe som skal være skjult", og vi gjør allerede dette til en viss grad i f.eks. tabellen til repeterende grupper, hvor innhold (displayValue) vises i tabellen selv om komponenten det pekes på er skjult.

src/layout/Input/config.ts
```diff
@@ -167,5 +160,5 @@ export const Config = new CG.component({
           'Max length of the input field. Will add a counter to let the user know how many characters are left.',
         ),
     ),
-  )
-  .addProperty(new CG.prop('summaryProps', INPUT_SUMMARY_PROPS).onlyIn(Variant.Internal));
+  );
+//.addProperty(new CG.prop('summaryProps', INPUT_SUMMARY_PROPS).onlyIn(Variant.Internal));
```
olemartinorg, line 164: I prefer just deleting the code instead of commenting it out :innocent:

src/layout/Summary2/SummaryComponent2/ComponentSummary.tsx
```diff
@@ -11,54 +11,45 @@ import type { LayoutNode } from 'src/utils/layout/LayoutNode';
...
-      {childComponents}
     </Grid>
   );
 }
 
 export function ResolveComponent({ summaryProps, summaryOverrides }: ResolveComponentProps) {
-  const resolvedComponent = useNode(summaryProps.whatToRender.id);
+  if (!summaryProps.target?.id) {
+    throw new Error('Tried to render component without, this should not happen.');
+  }
```
olemartinorg, line 50: Without what? :wink: Also, isn't this a configuration error by the app developer? In such cases we should log to DevTools instead, via `window.logError()`.

src/layout/Summary2/SummaryComponent2/SummaryComponent2.tsx
```diff
@@ -12,52 +12,40 @@ export interface ISummaryComponent2 {
 }
 
 export function _SummaryComponent2({ summaryNode }: ISummaryComponent2) {
-  if (summaryNode.item.whatToRender.type === 'layoutSet') {
+  if (!summaryNode.item.target && !summaryNode.item.taskId) {
+    return <LayoutSetSummary />;
+  }
+
+  if (summaryNode.item.taskId) {
```
olemartinorg, line 19: Ohh, so you can point to one page or component in a different task? :star_struck: But can you point to just "the whole task" as well, aka all pages in that task? If we make app developers have to add multiple Summary2 components to point to different pages in the same (previous) task, we'll also force ourselves to generate state (like nodes) independently for each Summary2 component. When the hierarchy rewrite is merged, I suspect that will add lots of overhead and duplicated state in memory.

One alternative is to try to deduplicate `<FormProvider>` across multiple Summary2 components with the same task, but it will complicate things quite a lot if we have to store "state for a task" in an external store disconnected from the render tree..

Anyway - one more thing I thought when seeing this - shouldn't `taskId` also be a property of `target`? It's very related to the target, at least.

olemartinorg (approved):

src/layout/Summary2/SummaryComponent2/TaskSummary.tsx
adamhaeger: This has now been changed to be opt-in :)

src/layout/Summary2/SummaryComponent2/SummaryComponent2.tsx
adamhaeger: Moved the taskId to target :) I see your point with regards to multiple states. In light of our discussions about navigation vs tasks, this feature might become less important though, so I dont think we should put too much effort into it at this point, but we can mention in the docs that we advice against putting several summary components that render previous tasks on the same page for example :)

PR #2076, opened 2024-05-07
Several performace improvements

bjosttveit:
## Description

These performance improvements were identified by testing `ssb/ra0255-01` with 250 rows in a repeating group. Specifically, by adding an artificial pagination set-up so only 10/250 rows were visible. That way, performance issues that were not directly tied to rendering many components could be found more easily.

1. **Problem**: In `useDelayedMemoSelectorProto`, checking if a selector had changed was slow due to having to call `find` on the list of selectors, which would be very large for an app with hundreds of rows. This would also have to be called for every selector when rendering.
  **Solution**: Store previously accessed selectors in a `Map` instead of an array, to be able to find a selector in constant time.
2. In `useDelayedMemoSelectorProto`, when checking if a selector had changed, it would not reuse the previous value even if it had not changed due to a mistaken `!deepEqual`, causing unecessary rerenders.
3. **Problem**: `useTaskErrors` had to get validations for all nodes very often, which is slow for apps with many nodes, and also causing many rerenders in the `ErrorReport` due to being dependent on `useNodes` which changed a lot. 
  **Solution**: The only practical dependency to nodes was which nodes actually exist and are visible, since the actual validations and validation visibility is stored elsewhere. Therefore, a `useVisibleNodes` hook was introduced which only updates whenever nodes are added, removed, or changes their hidden state.
4. `useDataModelBindingsValidation` had a similar problem to `useTaskErrors`. It showed up during profiling as it was running almost continually. This was solved in a similar way by introducing a hook which only updates when nodes are added or removed, since `dataModelBindings` are static for any node.
5. **Problem**: The hierarchy was regenerated for each key-press, slowing things down significantly when the app had many nodes, regardless of whether the components were rendered or not. This was due to expression datasources changing on every render.
  **Solutions**:
    - Add `useCallback`s to `useAllOptionsSelector` as it was updated every render.
    - Wrap `useIsHiddenPage` in a `useCallback` as this caused `usePageNavigationConfig` to update on every render.
    - Move selector- and cacheKey-functions outside of `useDebouncedSelector` to prevent it from being created on every render, causing `useDelayedMemoSelectorFactory` to update on every render.
    - Same as above but for `useIsHiddenComponent`.
    - In `useDelayedMemoSelectorProto`, store the raw value of the selector as well as the value after post-processing. Use the raw values when checking if the selectors needs to be cleared and re-rendered. Previously, the post-processed value was compared with the raw output from the selector, causing it to clear for every render even if no selectors had changed.

**Note**: The optimizations for `useTaskErrors` and `useDataModelBindingsValidation` may have been made redundant by reducing the no. of times the hierarchy has to be generated, as these optimizations were made because the `useNodes`-hook was updated so often.

## Related Issue(s)

- In preparation for #581 

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels

### DIFF
--- a/src/core/contexts/zustandContext.tsx
+++ b/src/core/contexts/zustandContext.tsx
@@ -20,10 +20,7 @@ type DelayedSelectorFunc<T> = <U>(selector: Selector<T, U>, postProcessor?: (dat
 type DelayedSelectorFuncWithArg<Arg, U> = (lookup: Arg, postProcessor?: (data: unknown) => U) => U;
 type SelectorFuncLax<T> = <U>(selector: Selector<T, U>) => U | typeof ContextNotProvided;
 
-type DelayedSelectorState<T> = {
-  selector: Selector<T, any>;
-  prevValue: any;
-}[];
+type DelayedSelectorState<T> = Map<Selector<T, any>, { selector: Selector<T, any>; value: any }>;
 
 interface DelayedSelectorFactory<Param, RetVal, T> {
   selector: (lookup: Param) => Selector<T, RetVal>;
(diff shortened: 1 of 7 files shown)


olemartinorg (approved):
Fantastic, thank you! 🙌 🚀

PR #2085, opened 2024-05-10
Pagination for repeating groups

### LINKED ISSUE #581: Repeating groups: Pagination
### Description

In repeating groups, we'd like to add the option to limit the currently visible rows via pagination. This is only a visual effect, and does not change the content in the data model in any way. The design system components for pagination in a table will arrive soon via an external contribution.

### Blocked by
- #580 

### Related issue(s)
- #107
- #578 (Use the same configuration options for pagination as in this issue)

bjosttveit:
## Description

Adds basic support for pagination in repeating groups:
- Supports all modes (`edit.mode`)
- Pages containing validation errors are marked in red
- It will scale to support any device width

The configuration options currently available is the number of rows per page, and the text used on the next and previous buttons:
```json5
...
"textResourceBindings": {
  "pagination_next_button": "Fremover",
  "pagination_back_button": "Bakover"
},
"pagination": {
  "rowsPerPage": 10
},
...
```

bilde

### Other changes
- Rows containing errors now only get highlighted if one of the nodes that are shown in the edit container have errors. Not if only componets that are visible in the table have errors, as these will be highlighted already.
- Rows which were highlighted due to errors would loose the color on mouse hover, fixed.
- The `NodeNavigationHandler` for repeating groups would open the row for editing even though the field was shown in the table directly. This was due to cheking the `tableColumns` config using the `node.item.id` instead of the `node.item.baseComponentId` meaning that it would not find the config and therefore the check would always fail. This has been fixed.
- Made a small performance improvement on the `ErrorReport`. Turns out `JSON.stringify` (used for the key) for every validation error on every render was slowing things down. This showed up during profiling when there were many errors visible. Since validation objects do not have any unique fields, the `useGetUniqueKeyFromObject` was introduced as a fast way of getting a unique id based on an object reference. Since validations are reused if nothing changes, this will still prevent some rerenders while being much more efficient than `JSON.stringify` (even a completely random key for each render was faster than this).

## Related Issue(s)

- closes #581 

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels


src/layout/RepeatingGroup/RepeatingGroupContext.tsx
```diff
@@ -124,6 +128,87 @@ function produceStateFromNode(node: BaseLayoutNode<CompRepeatingGroupInternal>):
...
+      rowsPerPage: undefined;
+      rowsToDisplay: Row[];
+    };
+
+/**
+ * Produces the current pagination state if relevant
+ */
+function producePaginationState(
+  currentPage: number | undefined,
+  node: BaseLayoutNode<CompRepeatingGroupInternal>,
```
olemartinorg, line 152: Not really used that much, but in principle the component can override `BaseLayoutNode` and provide its own implementation. I do that in the `hierarchy-generator-rewrite` branch, but will probably revert that particular change down the road (as I don't seem to need it anyway), but I did have to go through lots of places and correct the code to point to LayoutNode instead of BaseLayoutNode.

```suggestion
  node: LayoutNode<'RepeatingGroup'>,
```

src/layout/RepeatingGroup/RepeatingGroupFocusContext.tsx
```diff
@@ -36,40 +36,48 @@ export function RepeatingGroupsFocusProvider({ children }: PropsWithChildren) {
...
       const tableColSetup =
-        (node.item.tableColumns && targetChild.item.id && node.item.tableColumns[targetChild.item.id]) || {};
+        (node.item.tableColumns && targetChildBaseComponentId && node.item.tableColumns[targetChildBaseComponentId]) ||
+        {};
 
       if (tableColSetup.editInTable || tableColSetup.showInExpandedEdit === false) {
         // No need to open rows or set editIndex for components that are rendered
         // in table (outside the edit container)
-        return false;
+        return true;
```
olemartinorg, line 75: I think the reason this returned false was that we can't guarantee the component is visible now, but if the component is rendered by GenericComponent somewhere on the page, that will return true. Returning true here would mean "stop looking for it, it's fine now", but there may be other navigation handlers that need to run for this to become visible (either now or at a later time with some new functionality to come). So stopping it early may be a disservice.

src/layout/RepeatingGroup/config.ts
```diff
@@ -190,6 +204,17 @@ export const Config = new CG.component({
...
     ),
   )
+  .addProperty(
+    new CG.prop(
+      'pagination',
+      new CG.obj(new CG.prop('rowsPerPage', new CG.int().setMin(1)))
+        .optional()
+        .setTitle('Pagination options')
+        .setDescription(
+          'Pagination options for the repeating group table. Not applicable when `edit.mode` is `showAll`',
```
olemartinorg, line 214: Looks like it now works even with `showAll`? Maybe this comment was written before then. :innocent:

src/utils/useGetKeyFromObject.ts
```diff
@@ -0,0 +1,29 @@
...
+ * The objects themselves do not have any unique identifiers,
+ * and most/all fields could potentially match in other objects.
+ * Using JSON.stringify(object) for each object in each render turns out to be very slow.
+ *
+ * This hook takes an object reference, which should be unique in the list,
+ * and assigns it a unique numeric id, scoped to the component where the hook is used.
+ * The use of WeakMap ensures that if the object is no longer referenced anywhere besides here,
+ * it can still be garbage collected.
+ */
+export function useGetUniqueKeyFromObject() {
```
olemartinorg, line 15: :star_struck:

olemartinorg (approved):
Looks good!! :godmode:

src/layout/RepeatingGroup/config.ts
bjosttveit: Ahh yes, woops 🙈

src/layout/RepeatingGroup/RepeatingGroupContext.tsx
bjosttveit: Yeah I pretty much copied from:
```typescript
function produceStateFromNode(node: BaseLayoutNode<CompRepeatingGroupInternal>): NodeState {
```

PR #2097, opened 2024-05-16
Fix cypress error caused by mixing up sync and async code

bjosttveit:
## Description

The cypress test for `components.ts` -> `is possible to upload and delete attachments ` keeps failing in GitHub. I was able to reproduce it locally by enabling network throttling. Cypress says the error is due to mixing up synchronous and asynchronous code, and this fixed it locally. I am not exactly sure how this custom cypress command works, so please correct me if this is wrong @olemartinorg 😅 

Pasted Graphic


olemartinorg (approved):
I have no idea either! 🤷 But if this is the solution, and it works, it's all good.

PR #2109, opened 2024-05-24
Use language in Nets

Magnusrm:
## Description

Giva current selected language as query parameter to payment endpoints in order to make use of the language functionality in backend.

Also solved an insconsistency with going directly to Nets where it would only work the first time, since the response for paymentInformation would be cached and it would seem like a payment existed even though it had been cleaned up in backend.

## Related Issue(s)

- closes #{issue number}

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels

### DIFF
--- a/src/features/payment/PaymentInformationProvider.tsx
+++ b/src/features/payment/PaymentInformationProvider.tsx
@@ -6,6 +6,7 @@ import { useAppQueries } from 'src/core/contexts/AppQueriesProvider';
 import { delayedContext } from 'src/core/contexts/delayedContext';
 import { createQueryContext } from 'src/core/contexts/queryContext';
 import { useLaxInstance } from 'src/features/instance/InstanceContext';
+import { useCurrentLanguage } from 'src/features/language/LanguageProvider';
 import { useIsPayment } from 'src/features/payment/utils';
 import type { QueryDefinition } from 'src/core/queries/usePrefetchQuery';
 import type { PaymentResponsePayload } from 'src/features/payment/types';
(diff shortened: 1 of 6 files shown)


src/features/payment/PaymentInformationProvider.tsx
```diff
@@ -16,10 +17,12 @@ export function usePaymentInformationQueryDef(
   instanceId?: string,
 ): QueryDefinition<PaymentResponsePayload> {
   const { fetchPaymentInformation } = useAppQueries();
+  const selectedLanguage = useCurrentLanguage();
   return {
     queryKey: ['fetchPaymentInfo'],
-    queryFn: instanceId ? () => fetchPaymentInformation(instanceId) : skipToken,
+    queryFn: instanceId ? () => fetchPaymentInformation(instanceId, selectedLanguage) : skipToken,
     enabled: enabled && !!instanceId,
+    gcTime: 0,
```
bjosttveit, line 25: Good catch!

bjosttveit (approved):
Looks good, is it not then necessary to clear the order-details-query-cache when the language changes? (in the PaymentDetails component)

Magnusrm:
@bjosttveit The `PaymentDetails` component uses the `orderDetailsProvider` and not the `PaymentInformationProvider` so it is sadly still needed in order to update orderlines dynamically.  

A solution we had earlier for the `PaymentDetails` component was to always refetch the orderlines after the datamodel saved data, regardless if the datafield was relevant or not. This allowed us to remove mapping, however it might be a bit much to allways refetch in regards to performance.

bjosttveit:
I just meant that if the order details now depends on the language, maybe also clear the cache when the language changes, in addition to when the mapping changes. If that makes any difference that is 😄

PR #2110, opened 2024-05-24
Feature/tabs

### LINKED ISSUE #1982: Tab komponent
## Beskrivelse
Støtte for [tab komponenten fra designsystemet](https://storybook.designsystemet.no/?path=/docs/komponenter-tabs--docs)

Link til skisser i Figma; https://www.figma.com/design/6dXNs3Ncbg9dJoK98qyZhV/Dashboard%2FSamhandlingstjeneste-i-Altinn?node-id=296-33960&m=dev 
Skjermbilde 2024-05-16 kl  14 21 46

## Ønsket løsning
En ny komponent i Altinn Studio og Apps som på hver tab kan inneholde andre komponenter.

## Akseptansekrav
- [ ] Hver tab må kunne inneholde child komponenter inkludert tab komponenten
- [ ] Støtte tastatur for navigering gjennom tilgjengelige tabs

## Gjennomføring
```[tasklist]
### Oppgaver
- [ ] Add a draft title or issue reference here
```

cammiida:
## Description

Created a new component, "Tabs", where you can configure each tab with the properties: id, title, icon and children. Children is a list of Generic components. Created summary for the component and  e2e tests, as well as updated existing tests.

## Related Issue(s)

- closes #1982 

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels


src/layout/LayoutComponent.tsx
```diff
@@ -83,6 +83,13 @@ export abstract class AnyComponent<Type extends CompTypes> {
    */
   abstract canRenderInAccordionGroup(): boolean;
 
+  /**
+   * Return false to allow this component to be rendered in a Tabs
```
olemartinorg, line 87:
```suggestion
   * Return true to allow this component to be rendered in a Tabs
```

src/layout/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,116 @@
...
+import type { PropsFromGenericComponent } from 'src/layout';
+
+export const Tabs = ({ node }: PropsFromGenericComponent<'Tabs'>) => {
+  const [activeTab, setActiveTab] = useState<string | undefined>(
+    node.item.defaultTab ?? node.item.tabsInternal.at(0)?.id,
+  );
+
+  useRegisterNodeNavigationHandler((targetNode) => {
+    for (const parent of targetNode.parents() ?? []) {
+      const tabsInternal = parent.item['tabsInternal'];
```
olemartinorg, line 17: I'd rather check that `parent.isType('Tabs') && parent.item.id === node.item.id`.

With this code you're indirectly checking that the parent is a `Tabs` component by relying on some internal state unique to the Tabs component right now - but who knows, in the future other components may also get the `tabsInternal` property (just like Grid rows are re-used in RepeatingGroup configuration). Also, you're not checking that the parent is _this_ Tabs component, just any of them. So if you have two Tabs components on the same page, I suspect this will cause tab-switching in the unrelated one.

src/layout/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,116 @@
...
+          />
+        ))}
+      </DesignsystemetTabs.List>
+      {tabs.map((tab) => (
+        <DesignsystemetTabs.Content
+          key={tab.id}
+          value={tab.id}
+          role='tabpanel'
+          style={{
+            backgroundColor: 'white',
```
olemartinorg, line 54: Do we have to have this? I'm curious how Tabs would look inside for example a Panel or a RepeatingGroup, which have other background colors than a plain top-level form component.

src/layout/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,116 @@
...
+  if (icon) {
+    const imgType = icon.split('.').at(-1);
+
+    if (!imgType) {
+      throw new Error('Image source is missing file type. Are you sure the image source is correct?');
+    }
+
+    if (imgType.toLowerCase() !== 'svg') {
+      throw new Error('Only SVG icons are supported');
+    }
```
olemartinorg, line 92: Any specific reason not to support `.png` and other image types?

src/layout/Tabs/config.ts
```diff
@@ -0,0 +1,61 @@
...
+    renderInTabs: false,
+    renderInCards: false,
+    renderInCardsMedia: false,
+  },
+})
+  .addTextResource(
+    new CG.trb({
+      name: 'ariaLabel',
+      title: 'Aria Label',
+      description: 'The aria label for the tabs',
```
olemartinorg, line 21: Would be nice to describe what an aria label is and how this is being used, instead of just repeating the title in the description with some more words. :pray: I.e. "Label read aloud by screen readers when \<...\>"

src/layout/Tabs/config.ts
```diff
@@ -0,0 +1,61 @@
...
+    }),
+  )
+  .addProperty(new CG.prop('size', new CG.enum('small', 'medium', 'large').optional()))
```
olemartinorg, line 24: In `.optional()` you can also specify what's the default setting if none are provided. :raised_hands:

src/layout/Tabs/config.ts
```diff
@@ -0,0 +1,61 @@
...
+  .addProperty(new CG.prop('defaultTab', new CG.str().optional()))
+  .addProperty(
+    new CG.prop(
+      'tabs',
+      new CG.arr(
+        new CG.obj(
+          new CG.prop('id', new CG.str()),
```
olemartinorg, line 31: Did you consider making `tabs` a key/value map instead, where the key is the tab ID? Or possibly making it an array without an `id` and using indexes instead? It would of course be nice to have some validation or error-handling if you happen to configure two tabs with the same ID, which would not be possible to do with any of the other solutions.

src/layout/Tabs/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+      </>
+    );
+  }
+
+  renderSummaryBoilerplate(): boolean {
+    return false;
+  }
+
+  shouldRenderInAutomaticPDF(node: BaseLayoutNode<CompTabsInternal, 'Tabs'>): boolean {
+    return !node.item.renderAsSummary;
```
olemartinorg, line 53: I believe the automatic PDF generator checks this for you already? :thinking: Did you test without it?

test/e2e/integration/frontend-test/pdf.ts
```diff
@@ -16,12 +16,13 @@ describe('PDF', () => {
     });
   });
 
-  it('should generate PDF for changename step', () => {
+  it.only('should generate PDF for changename step', () => {
```
olemartinorg, line 19: Whoops, better not commit these! :scream: There's a pre-commit husky script that should warn you when committing a test isolated using `it.only`, did it not work correctly for you or did you skip pre-commit checks? :thinking:

test/e2e/integration/frontend-test/validation.ts
```diff
@@ -120,22 +122,20 @@ describe('Validation', () => {
...
       .should('have.length', 4 + 2);
-
     const lastNameError = appFrontend.fieldValidation(appFrontend.changeOfName.newLastName);
     cy.get(lastNameError).should('exist').should('not.be.inViewport');
     cy.get(appFrontend.changeOfName.newLastName).should('not.be.focused');
 
     cy.get(appFrontend.errorReport)
       .get(`button:contains("${texts.requiredFieldLastName}")`)
-      .should('be.inViewport')
+      //.should('be.inViewport')
```
olemartinorg, line 138: Hmm..? :thinking: Did this cause the test to be flaky, or not work at all? The reason we test for this is that clicking on the submit button should scroll the page such that the submit button is in the exact same place on screen as before clicking it. Otherwise you'll click the button and the button potentially just disappears for you underneath a long list of validation errors.

src/layout/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,116 @@
...
+      {!!icon && (
+        <img
+          src={icon}
+          alt={translatedTitle ?? 'tab icon'}
+          style={{
+            width: '24px',
+          }}
+        />
+      )}
+      {translatedTitle}
```
olemartinorg, line 113: I'd prefer using the `<Lang />` component here instead of the stringy version. That makes it possible to support rich text in a tab title. :raised_hands:

olemartinorg (commented):
Good stuff!! :star_struck: Sorry for the long list of comments :face_with_peeking_eye: You've done a great job here, just a few (mostly minor) things I'd want to double-check before merging.

test/e2e/integration/frontend-test/pdf.ts
bjosttveit: I suspect it does not work correctly, I have never been stopped from making this mistake 😅

src/layout/Tabs/index.tsx
bjosttveit: I dont think this is checked in the auto-generation logic, but there is a default implementation of this method that does something similar, I see it is also used in Accordion, but I am not sure if this is inherited for these types of components or not 🤷

test/e2e/integration/frontend-test/pdf.ts
paal2707: ops, i did not skip it intensionally. Im not sure, havent seen any.

src/layout/Tabs/Tabs.tsx
paal2707: isType() finnes ikke på parent?

src/layout/Tabs/config.ts
paal2707: I have considered using indexes instead, but then you would not get to add validation or error handling. Could/would we add generated Ids for tabs to avoid similar id issues?

test/e2e/integration/frontend-test/validation.ts
paal2707: I think i commented it out to understand what was happening in the test, then i forgot to uncomment it back in.

src/layout/Tabs/Tabs.tsx
olemartinorg: Ah, `parent instanceof BaseLayoutNode && parent.isType('Tabs')` burde derimot funke

src/layout/Tabs/config.ts
olemartinorg: Hmm, not sure what you mean - in what way does validation or error handling not work when using indexes? I'm not aware of any such limitations (i.e. `Grid` is using arrays instead of objects with IDs).

> Could/would we add generated Ids[...]?

Studio can, and should, generate unique IDs for tabs, yes. However, app developers working on plain layout files cannot. We could make the `id` property optional and generate one if it's missing - but then I would argue that it's entirely useless functionality (when the index is already exactly that - a guaranteed unique identifier for each tabs in the array of tabs).

src/layout/Tabs/Tabs.tsx
```diff
@@ -104,13 +102,13 @@ function TabHeader({
       {!!icon && (
         <img
           src={icon}
-          alt={translatedTitle ?? 'tab icon'}
+          alt=''
```
olemartinorg, line 105: This would break WCAG requirements, I think..? :thinking: You can still use `langAsString()` here even if you use `<Lang />` below.

paal2707: No actually it wont. The designsystem is handling it. It takes the tab text an relates it to the icon.  You get a warning when you add alt text, saying that it is not necessary to have multiple text describing the image. But you should not remove the alt tag, so I added an empty one(It is the wcag way)

olemartinorg (approved):
Looks good! :rocket:

PR #2125, opened 2024-05-30
removes FooterProvider in favor of just useQuery

cammiida:
## Description

Since footerLayout is server data, it can just be kept in Tanstack 👌 I might be very wrong with regards to the stale time, tho, so happy to receive any feedback 😄 

## Related Issue(s)

- closes #{issue number}

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels


olemartinorg:
> Nice simplifaction and I think this approach might help with performance as well as context is known to rerender a lot of its children.

That only happens when the data in the context changes, which for the footer, never does.

> The only thing that might be a problem is data fetching between route transitions.

I believe the footer is static across all app tasks (or at least the footer layout is statically defined across them all), so no, I don't think that's a problem.

We (@cammiida, @bjosttveit and me) talked a bit about this on Slack last week, and while this looks good in principle the only drawback I can see with this is that it introduces a slightly different way of doing things than the providers we're currently using for pretty much everything else.

adamhaeger:
> > Nice simplifaction and I think this approach might help with performance as well as context is known to rerender a lot of its children.
> 
> That only happens when the data in the context changes, which for the footer, never does.
> 
> > The only thing that might be a problem is data fetching between route transitions.
> 
> I believe the footer is static across all app tasks (or at least the footer layout is statically defined across them all), so no, I don't think that's a problem.
> 
> We (@cammiida, @bjosttveit and me) talked a bit about this on Slack last week, and while this looks good in principle the only drawback I can see with this is that it introduces a slightly different way of doing things than the providers we're currently using for pretty much everything else.

So my performance comment was not aimed at this implementation, but the pattern in general. I think it might be a good direction to go in, and away from the provider pattern. But that probably warrants a longer discussion in the team :)

adamhaeger:
@cammiida @olemartinorg  so what do we want to do with this one and https://github.com/Altinn/app-frontend-react/pull/2124 ?

olemartinorg:
I think this one is fine - as mentioned before, `FooterProvider` is probably one of the few that didn't have to be blocking. If we go ahead doing this with the other simpler query providers, we'd have to make adjustments to `ReadyForPrint` as well (or do something more fundamental with the PDF generator first, as discussed in https://github.com/Altinn/app-template-dotnet/issues/214 and [here on Slack](https://digdir.slack.com/archives/C076T2293CM/p1717677412142179)).

As for #2124, that's riskier, and would need some additional work for me to be comfortable merging it, at least. After a cursory look, it was clear that PDF generation could run without that query being loaded, and error handling was missing a few places - and when errors were handled, they were handled incorrectly and would lead to rendering the `<DisplayError>` component _instead of_ the organization logo, which would look pretty bad.

bjosttveit (approved):
## 🙌

PR #2140, opened 2024-06-07
Alert on change for Dropdown and MultipleSelect

bjosttveit:
## Description

Adds `alertOnChange` property to `Dropdown` and `MultipleSelect`. These work exactly the same way as RadioButtons and Checkboxes respectively, meaning that it will show an alert when you "remove a value". For Radiobuttons and Dropdown this means that no alert is shown if the value was empty before, but does show when the previous value was not null. For checkboxes and multiple select it shows an alert when you uncheck or remove a selected value, this includes when you click the multiple select's delete all button.

The design system was upgraded for this to work, and the `LegacySelect` has been upgraded to the new `Combobox` component from the design system for `Dropdown`, `MutlipleSelect`, and `FileUploadWithTag` (and everywhere else).

Due to design system changes in this version I had to update some other components as well, most notably `Paragraph`. The design system `Paragraph` no longer lets you choose the tag to use, so I had to conditionally wrap it instead. When making these changes I noticed that our language function not only replaces the so called "root element" with a span (to support inline content in buttons etc.), it also replaces it even if there is no "root" but rather multiple block level elements, causing cases where I personally think the `p`-element should have been left in place to also be replaced with a span. I made a small change to this logic to not only check if it is at the top-level, but also check if it has any siblings. That way, if you have a say paragraph and a list as the content, the p-element will be left in place. I have not noticed this messing up the texts anywhere else, but I can remove it in case it does.

Another thing related to the Paragraph element, some of the unit tests did not actually do anything. This is because it did a `.forEach` with an async function which would finish immediately and no expect-calls would therefore happen before the function returned, which just meant they were always green. This has been fixed by changing it to a `for..of` loop instead of a `.forEach`.

## Related Issue(s)

- closes https://github.com/Altinn/app-frontend-react/issues/1233

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels


src/index.css
```diff
@@ -11,6 +11,11 @@
   color: var(--semantic-text-neutral-default);
 }
 
+/* Workaround for the design-system aggressively setting the font to 'Inter' */
+[class^='fds-'] {
+  font-family: var(--font-family);
+}
```
olemartinorg, line 17: :sweat_smile: :clap:

test/e2e/integration/frontend-test/all-process-steps.ts
```diff
@@ -70,10 +70,8 @@ function testConfirmationPage() {
 
   cy.snapshot('confirm');
 
-  // TODO: Enable this when it doesn't break custom receipt. It would be nice to make sure refreshing on a confirm step
-  // restores all state and we're not stuck in any way here.
-  // cy.reloadAndWait();
-  // cy.get(appFrontend.confirm.container).should('be.visible');
+  cy.reloadAndWait();
+  cy.get(appFrontend.confirm.container).should('be.visible');
```
olemartinorg, line 74: Oh, nice! :raised_hands: That should probably have been commented back in a good while ago.

olemartinorg (approved):
Great stuff! :godmode:

test/e2e/integration/frontend-test/all-process-steps.ts
bjosttveit: yeah probably for v4.0.0 😅

adamhaeger (approved):

PR #2159, opened 2024-06-13
DevTools: Click directly on components to show details

bjosttveit:
## Description

Convenient little function to more easily find components in dev tools by simply clicking on them, similar to how browser dev tools do it.

I also added the appropriate `data-componentid` and `data-componentbaseid` attributes where they were missing so that all components should be properly highlighted now.

https://github.com/Altinn/app-frontend-react/assets/47412359/72ae90c3-ea53-4299-9907-751d4e83ccfb

### DIFF
--- a/src/features/devtools/DevTools.module.css
+++ b/src/features/devtools/DevTools.module.css
@@ -61,6 +61,7 @@
 }
 
 .tabList {
+  overflow-y: hidden;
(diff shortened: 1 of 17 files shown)


olemartinorg (approved):
Exquisite! :pinched_fingers:

PR #2186, opened 2024-06-25
Fix dropdown crashing if option is missing label (attempt #2)

bjosttveit:
## Description

This is a new workaround replacing the solution from #2166. Instead of filtering out empty options which is arguably breaking (although its now possible to delete a selected option from Dropdown meaning that empty options should no longer be necessary), I now try to still support empty options without crashing the design system combobox. Since it throws an exception if the `displayValue` is null or empty, I instead give it a zero-width space as a fallback. This looks and acts completely as you would expect, and you are still able to delete the "empty" selected option with backspace. However, if the value is also an empty string, this will be treated as removing the value by json patch, and so it will not show as selected (just remove the existing value). The option list will look a bit wonky without any text (will shrink in height), so I added a `<wbr>` (Word Break Opportunity) element which acts just like a zero-width space.

This does indeed feel a little bit hacky, but its certainly less breaking than the previous attempt, and should actually work for all cases now.

Previously, you could use commas in option values that were used for single-select components (Dropdown and Radiobuttons), this was broken in #2140 and should now be restored.

A final addition is a small "verification" step in useGetOptions that checks if option values used with MultipleSelect or Checkboxes contains `,` and logs an error. This will not work as expected since we store the selected values in a comma-separated string. It also logs errors/warnings if the value is an empty string, as this will not work properly for multi-select components (error), and can look weird for single-select components depending on their use (warning).

## Related Issue(s)

- https://digdir.slack.com/archives/C077SDN5TT5/p1719308682879759?thread_ts=1719303938.809849&cid=C077SDN5TT5

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels


olemartinorg (approved):
Looks good! More compliant than last time, and I really like having warnings for these wonky options! :raised_hands:

PR #2196, opened 2024-06-28
Feature/number component

### LINKED ISSUE #2179: Number Component
Primitive base component for displaying numbers

Should support:
Label, data, icon, horizontal/vertical view

Config
```
{
  "id": "fornavn",
  "type": "Number",
  "direction": "horizontal",
  "format": "??"
  "textResourceBindings": {
    "title": "Fornavn"
  },
  "value": ["dataModel", "Person.Fornavn", "folkeregisteret_api"]
}
```
Image
Image

paal2707:
## Related Issue(s)

- closes #2179 

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels


src/layout/Number/NumberComponent.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+  if (icon) {
+    const imgType = icon.split('.').at(-1);
+
+    if (!imgType) {
+      throw new Error('Image source is missing file type. Are you sure the image source is correct?');
+    }
+    if (!['svg', 'png', 'jpg', 'jpeg', 'gif', 'bmp', 'tiff'].includes(imgType.toLowerCase())) {
+      throw new Error('Only images of the types: .svg, .png, .jpg, .jpeg, .gif, .bmp, .tiff, are supported');
+    }
```
olemartinorg, line 32: I'm still not excited about these types of limitations - why do we need them? We're making a developer platform, so if our app developers configures something that's not valid as an image, that's on them - not us. I would love for apps to be able to use AVIF, WebP or any other new and fancy image format that gets into browsers in the future, without us having to add things to a whitelist.

src/layout/Number/index.tsx
```diff
@@ -0,0 +1,13 @@
...
+import { NumberComponent } from 'src/layout/Number/NumberComponent';
+import type { PropsFromGenericComponent } from 'src/layout';
+
+export class Number extends NumberDef {
+  render = forwardRef<HTMLElement, PropsFromGenericComponent<'Number'>>(
+    function LayoutComponentNumberRender(props, _): JSX.Element | null {
+      return <NumberComponent {...props} />;
+    },
+  );
+}
```
olemartinorg, line 13: This should probably also implement the `DisplayData<'Number'>` interface, so that the numeric display data can be referenced in an expression. :raised_hands:

src/codegen/Common.ts
```diff
@@ -710,6 +710,102 @@ const common = {
     )
       .setTitle('Layout set')
       .setDescription('Settings regarding a specific layout-set'),
+  IFormatting: () =>
```
olemartinorg, line 713: If this is duplicated now, maybe the definition in the `Input` component should reference this instead?

src/layout/Number/config.ts
```diff
@@ -0,0 +1,27 @@
...
+    renderInCardsMedia: false,
+  },
+})
+  .addTextResource(
+    new CG.trb({
+      name: 'title',
+      title: 'Title',
+      description: 'The title of the value',
+    }),
+  )
```
olemartinorg, line 23: I think this should just set `rendersWithLabel: true` above, and then you can add `.extendTextResources(CG.common('TRBLabel'))` instead of a manual `title` text resource binding. That way this component renders with the built-in label (which also supports description and helpText), and in those cases where you need to skip rendering the label (such as in a panel listing), you can render it with forced 'direct rendering' to avoid the label:

```JSX
<GenericComponent
  node={someNumberNode}
  overrideDisplay={{
    directRender: true,
  }}
/>
```

PR #2239, opened 2024-07-12
Refactor application metadata

cammiida:
## Description

- [x] Refactors application metadata fetching to use static fetching function.
- [x] Preprocesses application metadata for consistency.
- [x] Mocks fetchApplicationMetadata for all tests.
- [x] Removes unused fields.

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels


test/e2e/integration/frontend-test/navigation.ts
```diff
@@ -62,8 +62,8 @@ describe('Navigation', () => {
   it('should not focus main-content when loading the page in the browser and there is no instance selector', () => {
     cy.intercept('**/applicationmetadata', (req) => {
       req.on('response', (res) => {
-        const body = res.body as IApplicationMetadata;
-        body.onEntry = undefined;
+        const body = res.body as IncomingApplicationMetadata;
+        // body.onEntry = undefined;
```
adamhaeger, line 66: Remove comment?

adamhaeger (approved):

src/features/applicationMetadata/ApplicationMetadataProvider.tsx
```diff
@@ -2,68 +2,80 @@ import React, { useEffect } from 'react';
...
+      return {
+        ...data,
+        isValidVersion:
+          !!data.altinnNugetVersion &&
+          isAtLeastVersion({
+            actualVersion: data.altinnNugetVersion,
+            minimumVersion: MINIMUM_APPLICATION_VERSION.build,
+          }),
+        onEntry,
+        isStatelessApp: isStatelessApp(onEntry.show),
```
olemartinorg, line 38: I'm not sure I'm fully on board here... Our 'is stateless app' thing has been easy to misinterpret for a long while now, since it doesn't mean "is this app stateless" but rather "are we stateless _now_". Since "being stateless" just means "we're currently showing a data task before we potentially proceed to initialize a new instance" it also means that when the user clicks on an `InstantiationButton` in a stateless task we immediately go from being stateless to no longer being stateless.

So, conceptually, is `isStatelessApp` a property of the "application metadata"? I would argue it's not, it relies on more than that - including the current URL/route (to see if we're in an instance or not). The way this used to work made a bit more sense to me, as it was a hook (even though it did not re-render your component when the route changed), but now I'm not quite sure the state will update when you go from stateless to stateful...? It does, at least, rely on when/how often this `select(data)` runs.

src/features/applicationMetadata/types.ts
```diff
@@ -6,24 +6,35 @@ type ILogoOptions = {
...
   autoDeleteOnProcessEnd: boolean;
   onEntry?: IOnEntry;
   features?: Partial<IBackendFeaturesState>;
   promptForParty?: 'always' | 'never';
   logo?: ILogoOptions;
 }
 
+export type ApplicationMetadata = {
+  isValidVersion: boolean;
+  dataTypes: IDataType[]; // TODO: next step
```
olemartinorg, line 25: What does this TODO refer to?

src/features/applicationMetadata/types.ts
```diff
@@ -6,24 +6,35 @@ type ILogoOptions = {
...
+  id: string;
+  org: string;
+  partyTypesAllowed: IPartyTypesAllowed; // FIXME: needs improvement
```
olemartinorg, line 28: Which improvements are needed for this one? I'd prefer elaboration or an issue. :pray:

src/test/renderWithProviders.tsx
```diff
@@ -122,7 +121,6 @@ export const makeMutationMocks = <T extends (name: keyof AppMutations) => any>(
 
 const defaultQueryMocks: AppQueries = {
   fetchLogo: async () => getLogoMock(),
-  fetchApplicationMetadata: async () => getApplicationMetadataMock(),
```
olemartinorg, line 125: Not quite sure I'm happy now that mocking the application metadata query now needs to be done differently than every other query.. :disappointed:

What's the reason this was needed?

olemartinorg (commented):
I'm skeptical of the change around `isStatelessApp` and the move away from our usual mocking strategy for the application metadata query. This also caused quite the merge conflict for #2244 - so I had to do a post-merge look over these changes to get a sense of what changed when I'm to integrate this.

src/features/applicationMetadata/types.ts
cammiida: Forgot to remove these. Will remove in a later PR 👌

src/test/renderWithProviders.tsx
cammiida: It is to move away from the tightly coupled way of setting up a component for testing. Every component should be tested in isolation, which basically is not possible with the current setup. Thus, we don't really have any unit tests, only integration and end-to-end.

Also, `QueryProvider` is not needed as it doesn't actually hold state, but static code, and therefore only adds complexity.  The consequences of the current setup is also that anything we refactor will result in massive changes, because everything is connected to everything. We need to move away from that.  

Without the `QueryProvider`, getting the function to use in `useQuery` also does not need to be a hook, which relives some of the pain you have expressed 😊

PR #1558, opened 2023-10-10
Alert on change - radioButtons and checkBoxes

lassopicasso:
## Description
Introduced a new prop, `alertOnChange`, which can be configured to trigger an alert when users attempt to select a different radioButton or uncheck a checkBox. 
This should also enable handling through expressions, as desired by SSB. However, I haven't had time to test this with expressions.. :

If any adjustments are necessary (_hopefully not too many for your sake_) , I hope that you guys can make the necessary refinements while I'm on vacation. ❤️ 

Arigatou gozaimasu!

## Related Issue(s)

- closes #1233

## Verification/QA

- Manual functionality testing
  - [x] I have tested these changes manually
  - [ ] Creator of the original issue (or service owner) has been contacted for manual testing (or will be contacted when released in alpha)
  - [ ] No testing done/necessary
- Automated tests
  - [ ] Unit test(s) have been added/updated
  - [x] Cypress E2E test(s) have been added/updated
  - [ ] No automatic tests are needed here (no functional changes/additions)
  - [ ] I want someone to help me make some tests
- UU/WCAG ([follow these guidelines](https://aksel.nav.no/god-praksis/artikler/utvikling?tema=universell-utforming) until we have our own)
  - [x] I have tested with a screen reader/keyboard navigation/automated wcag validator
  - [ ] No testing done/necessary (no DOM/visual changes)
  - [ ] I want someone to help me perform accessibility testing
- User documentation @ [altinn-studio-docs](https://github.com/Altinn/altinn-studio-docs)
  - [ ] Has been added/updated
  
  - [ ] No functionality has been changed/added, so no documentation is needed
  - [ ] I will do that later/have created an issue
  
- Support in Altinn Studio
  - [ ] Issue(s) created for support in Studio
  
  - [ ] This change/feature does not require any changes to Altinn Studio
- Sprint board
  - [x] The original issue (or this PR itself) has been added to the Team Apps project and to the current sprint board
  - [ ] I don't have permissions to do that, please help me out
- Labels
  - [x] I have added a `kind/*` label to this PR for proper release notes grouping
  - [ ] I don't have permissions to add labels, please help me out

PR #2244, opened 2024-07-22
Performance: Hierarchy generator rewrite

### LINKED ISSUE #1182: Performance: Make each component responsible for running expressions

### LINKED ISSUE #1872: Could not find component/node for signature
### Description of the bug

Devtools generates a warning for signature files (attachments), which is confusing.

```
Could not find matching component/node for attachment signature1/74f6abe1-2505-4a49-8f60-d87d9e1d65df (there may be a problem with the mapping of attachments to form data in a repeating group). Traversed 0 nodes with id signature1
```

Should probably have a special case to avoid that warning for signature data elements.

### Steps To Reproduce

Open an app with signature elements and se that the warning is shown.

### Additional Information

_No response_

### LINKED ISSUE #1938: File uploads in repeating group: Errors, warnings, content flashing in-and-out
### Description of the bug

Cypress waits until things stabilize, so this wasn't caught in our automatic tests

### Steps To Reproduce

https://github.com/Altinn/app-frontend-react/assets/700139/a2704fa1-c97e-459d-8371-f04a9534f95c

### Additional Information

- https://altinn.slack.com/archives/C02EJ9HKQA3/p1710493534112109

olemartinorg:
### The broad strokes

This change reduces the amount of re-rendering and thus increases performance when working in A3 forms. The main culprit for why performance was less than stellar in v4+ was the node hierarchy, as this was re-built on every data model change (++) at the point of data model debounce. As a consequence, every form component got a new LayoutNode reference as a prop, and had to re-render.

As the node hierarchy is our internal representation of the current interpreted layout state, and enables functionality in the expression engine, validation, etc, making any changes to how this worked internally in order to reduce re-rendering was quite an undertaking - as reflected by the large line-count in this PR.

So, onto some details:
- Where you could previosly get data directly off a `node` object (i.e. an object of the `BaseLayoutNode` class) via `node.item`, you now need to `useNodeItem(node)` or a `nodeItemSelector`. This is because the `node` object lifetime is now stable for as long as the node is present in the form (i.e. not removed by removing a row in a repeating group), but the state for that node may change throughout (when expression results change). You can, however, get identificators (like `node.id` or `node.baseId`) off the node without fetching these from `useNodeItem()` or a selector.
- The hierarchy generator itself is now implemented as a series of React components that render `null`, but stays in-place in the virtual DOM for as long as the current `<FormProvider>` is present. This way each `node` object has a corresponding `<NodeGenerator>` in VDOM, with state being updated through `useEffect()` and set in a zustand store that contains all node-related data.
- Each node, when generated, has multiple sub-components that collect, process and save state to the zustand store. This includes both generating the reference `BaseLayoutNode` object, resolving any expressions in `hidden`, resolving other expressions from the layout, and (in dev environments or when errors occur) validate the component against the layout schema, data model bindings against the data model, etc.
- You may want to add more functionality on a layout component that needs to track state or evaluate expressions, so for this we now can implement a `NodeDefPlugin`. Using these are as simple as calling `.addPlugin(new MyPlugin(...))` from the component `config.ts` file. The full list of plugins as of now:
  - `AlertOnCHangePlugin` adds a property (usually `alertOnChange` or `alertOnDelete`) to a component. This will automatically be resolved as an expression.
  - `ValidationPlugin` stores two new properties on the node inside the zustand store, namely `validations` and `validationVisibility`. This plugin is automatically added for every component in the categories `CompCategory.Form` and `CompCategory.Container`.
  - `OptionsPlugin` will add properties and make sure the component support options (aka code lists).
  - `NonRepeatingChildrenPlugin` will handle lists of child references and adopt these as children inside your component (i.e. for `Group`, `Accordion`, etc).
  - `RepeatingChildrenPlugin` will work for repeating children, for components that binds to an `Object[]` the data model and generates _rows_ containing multiple children in each _row from the data model_. This is similar to the more specialized `LikertRowsPlugin`, but in the latter case the children are not defined beforehand, and only one simulated `LikertItem` node will be added per row.
  - `GridRowsPlugin` will handle statically defined table-rows for the `Grid` component, but also works for `rowsBefore` and `rowsAfter` in the `RepeatingGroup` component.
  - `AttachmentsPlugin` will store references to attachments bound to your component
  - `CardsPlugin` and `TabsPlugin` are more specialized plugins for adopting children in the `Cards` and `Tabs` container components.
- If needed, a component itself can now also implement its own properties that should evaluate expressions via setting `functionality: { customExpressions: true }` in `config.ts`. This makes it possible for components like `Input` to handle expressions deep inside the `formatting` property without this possibly affecting any other components. As the referenced issue suggested, this now makes it possible for any component to implement _support for expressions in a property_ without having to set this globally in the expressions engine - and also without having to write a full `NodeDefPlugin`.

### Further details

More things of note when looking through the changes in this PR:

- Plugins: Some plugins needs more detailed control over the nodes zustand store, so for these I created `NodeDataPlugin`. Making such a plugin lets you add action function and selectors to NodesContext, for tasks that are more specific than just `SetNodeProp` and `useNodeItem()`/`NodesInternal.useNodeData()`.
- Changes to options (some slightly breaking!)
  - `useGetOptions()` have been split up into `useFetchOptions()` + `useSetOptions()`. Options are now only fetched in the hierarchy generator, not when rendering the node itself.
  - `preselectedOptionIndex` and stored attachment references are now better, as they happen even without the target component being visibly rendered. That means that you can navigate away from an uploader component without losing the attachment reference from the form data (whoops, that was a bug), but also that `preselectedOptionIndex` will be set in the form data before the target component is rendered. This is slightly breaking, as you could theoretically use `preselectedOptionIndex` to figure out when a component first rendered (i.e. via the backend), but that's not functionality we had advertised anyway.
- Code generator:
  - I've removed the concept of 'internal-only types' in the code generator. This was used to indicate a difference in the external configuration (i.e. a layout file) and our internal representation (i.e. what was in `node.item` after the node hierarchy generator ran). Instead we now only refer to external types, and our internal types (via ex. `CompInternal<'Dropdown'>`) now points to the _inferred types_ from the return value of the `evalExpressions()` function in the component definition class. This now also means that if you add a property in your component via `.addProperty(...)` in `config.ts` and mark it as possibly containing an expression (`new CG.expr(...)`), the internal type will no longer have you fooled into thinking that expression got magically detected by the expression engine and resolved. As before, if you want to support expressions in a property, you have to implement support for that.
  - There are lots of additions to `ComponentConfig.ts`. I've used the code generator to do quite a few things to make `CompInteral` type inferral work the way I wanted to, and I've moved around and cleaned up things in the code generator to hopefully let things make more sense when new components are created. This includes not requiring to implement validation for data model bindings (and displayValue) when there are none.
  - I've made more enhancements to the code generator when performance-related work required me to refactor some things. Among others, the label rendering config for a component is no longer just `true | false`, it's an enum describing the complex realies we already have around label rendering.
- Node generator
  - The node generator (in `_src/utils/layout`) is a collection of components that does the heavy lifting when generating node objects and setting node state. A short description of the building blocks:
  - `LayoutSetGenerator`: This takes an entire layout-set, generates the pages, which takes components and passes them onto `NodeGenerator`. Before it can generate components, it has to render once first to figure out which components in a layout file are top-level components, and which are claimed as children.
  - `NodeGenerator` takes a layout configuration and generates a node object, resolve expressions for that configuration, run layout validations (i.e. validate the configuration against the schema, validate data model bindings) and include children.
    - To do this, it will wrap generating code in components from `GeneratorStages` (more about that later).
    - It does not know which properties in the configuration should be resolved as expressions - that job is outsourced to the component `index.tsx` file (spesifically the `evalExpressions()` method).
    - It also does not know which children to render per-component. That comes from the `renderNodeGenerator()` method in the component `index.tsx` file, which is generated according to which plugins the component is configured with.
  - `GeneratorContext` is a context that inherits from contexts above it, and will maintain hierarchy state in the generator, such as counting how many levels deep a node is in the hierarchy. This also includes the 'recursive and direct mutators' which are used to mutate the layout configuration for child nodes (i.e. appending `-0` to the `id` for nodes in the first row of a repeating group).
  - `GeneratorErrorBoundary` wraps the generation of a node. If anything goes wrong and crash during node generation, we aim to log that to the node state `errors` list. This can prevent full `unknown error` crashes when something goes wrong in a component, and will display a red border around an error list should one component fail. When rendering the node via `GenericComponent`, it will be wrapped in a functionally similar `ComponentErrorBoundary` for runtime errors.
  - `GeneratorStages` is a core part, and is a mechanism to delay and order rendering and `useEffect` hooks in the generator. This is needed because (for example) we cannot run expressions for a component before we can be sure all other nodes have been added to the zustand store. This is because an expression can look up any other component, and the expression should throw an error if the target component is not to be found (and we don't want to throw errors during the initial generation). To make this all work together, the generation process is split into different stages (i.e. add nodes, mark as hidden, fetch options, evaluate expressions ++). These stages all run in the specified order. When new nodes are to be added later (i.e. when adding a row to a repeating group), the generator stages will mark a new 'generator run number', which will run all of these stages again, in the given order, for these new nodes. The zustand store will be marked as 'not ready' until the 'finished' stage has been reached (more about that later). 
  - `NodesInternal.useIsReady()` and the ready-state for the node store is worth noting. Because there are multiple states to set (i.e. adding a node, resolving expressions, adding validations), and we don't want to run all selectors after each set (especially traversal selectors), the ready-state will be set to 'not ready' when new node generation starts, and will automatically be set to 'ready' once it finished.
  - Requests and batching: Some states that are set would run many `setState()` calls in series, such as each call to add one node. When we call `setState()`, zustand has to re-run all selectors to figure out what needs to re-render, so when we have many calls in a short time-period it would cause lots of unneccesary effort. To mitigate this we batch up multiple requests (i.e. to set node properties, adding nodes) and call `setState()` once in the end instead. This will usually mean that each stage in `GeneratorStages` only calls `setState()` once or twice (instead of hundreds or thousands of times).
- Validation
  - Validations for nodes are now declarative, as we do not store (most) validations we're not going to show to the user, so there is no need to selectValidations() and filterValidations() after fetching them from the store. Usually there's an appropriate selector for what you want (i.e. 'just give me visible validations of type error for this node'). However, we do still store valdations for hidden nodes, etc.
  - As validations for a node is now stored _inside that node object in the zustand store_, there is no longer a need for the validation subsystem to try to sync between the node hierarchy and the validation subsystem. Thus there is no hiearchy of validation visibility, and no need for the validation subsystem to figure out which nodes had their data change recently.
- Node traversal: We used to do node traversal before as well, via things like `node.children().each(...)`, but now that the node objects are stable across state changes we need to wrap these calls in a hook that makes sure changes to the hierarchy also causes re-renders.
- Navigation / React Router changes
  - `AppRoutingContext` was added on top to reduce re-renders when components used URL state. For example, if your component had `useIsPdf()` in it, navigating to another page in a form would re-render it. This happened because `react-router` always causes re-renders when you use hooks that depend on the current URL, and even though things like `useIsPdf()` just check if the query parameter `?pdf=1` is set, it caused re-renders when anything changed.
  - `useNavigate()` from `AppRoutingContext` is now preferable to the native hook from `react-router`, as ours does not cause your component to re-render when the URL changes. 
  - `useNavigationParam()` will get you the navigation params just like the ones you could get from `useNavigatePage()` before. This includes common URL elements like part ID, instance GUID, task and pageKey.
  - `useQueryKey()` will get the value of a given query key, and will only cause re-renders when that part of the URL changes (not any part of it).
- `Summary2` / `PdfView2`:
  - Using `Summary2Props` instead of individual props passed to `renderSummary2()`, and made sure these functions behave more like react components (such as rendering out other components in multi-select variants).
  - I re-used PdfView2 when re-implementing legacy automatic PDF page generation. When doing so, and when testing with existing apps, I found and fixed multiple issues in Summary2 / PdfView2:
    - Overrides were passed per-component, not all overrides at the same time. This caused overrides to not survive groups / container components. The types were cast as if the full override list was passed, but it never was.
    - PdfView2 rendered an existing 'pdf page' just like a regular page (i.e. it would pluck out all components from that page and generate Summary2-like views for them). This is not what we wanted - a pdf page can be rendrered directly (it is already contains summary components).
    - Refactoring SummaryComponent (v1) to properly allow not sending a `summaryNode` property (as was done in `PdfView2`, but not expressed in types). This way it's clear that the code supports rendering a summary even without a Summary component node.
  - Multiple smaller fixes to issues I found when testing `Summary2` and `PdfView2` in multiple apps and comparing old/new generated PDFs side-by-side.
- `useDelayedSelector()` got a significant upgrade with bug-fixes. It now has two modes, `simple` and `innerSelector`:
  - `simple` gives you a function that takes any number of arguments. Your component will only re-render if any of those arguments change, or if the result (given the same arguments) change. It is no longer limited to only one argument.
  - `innerSelector` will give you a function with a hook-like signature. That is, `selector((something) => something.inside[node.id], [node.id])` will make your component re-render only when the thing you selected changes, or when anything in the dependency array (second argument) changes. Crucially, this will behave much like a hook but lets you break some hook rules - you can call it as many or as few times during (or after) a render as you want.
  - In order to make this efficient, I created the `ShallowArrayMap` class. This gives us a fast lookup for shallow arrays of anythings, such as answering "look up the value I stored with the key `[1, 2, 'str', objRef]`" even when the outer array memory reference changes.
- Component ID duplication in a layout-set is now a _really_ deprecated. Because of how node data is stored in the zustand-store, app-frontend would crash if you gave it a layout-set with two or more components with the same ID.
  - Existing apps in production will have workarounds in `quirks.ts`, so these should not be affected. However, errors will be logged to DevTools for these apps, urging their developers to fix their layouts.
  - New apps, or apps only in tt02 with duplicate IDs, will have their duplicated removed from the layout by a preprocessor, and an error will be logged to DevTools.
  - This affected some shared expression tests as well. Some of those layouts were broken, and I cleaned them up so they don't have duplicate IDs any longer.
- Other layout problems in existing production apps have gotten workarounds in `quirks.ts`. I purposely made these checks error-tolerant and very strict, so that they'll fail (and not apply a quirk) as soon as these layouts change. My hopes is that app developers will take notice of this the next time they make changes to their apps, and fix their layouts. This also includes some Group components that reference children that does not exist (which did not give you any errors before).
- Renamed all `*.d.ts` files to just `.ts`. We have `skipLibCheck: true` in our `tsconfig.json`, so none of these files got any type-checking when they were called `.d.ts`.
- Some providers were removed
  - `AttachmentsStoreProvider`, `AttachmentsProvider`: Attachments are now stored inside the nodes context, via the `AttachmentsPlugin`
  - `AllOptionsStoreProvider`, `AllOptionsProvider`: Options are now stored inside the nodes context as well, via the `OptionsPlugin`. As a step in the node generation, all options will be fetched before the node hiearchy can be marked ready.
  - `LayoutValidationProvider`: Layout validation is now just a part of `NodePropertiesValidation`, and errors are stored alongside any other errors per-node in the node store `errors` property.
- I removed pretty colors for expression errors. These errors were logged to `console.log()`, but errors meant for app developers should go (and now go) to the DevTools console instead.
- The `linkToComponent` expression function was slightly changed. It used to link to the same page as the current one, but add a query key indicating that the real node should be navigated to. The problem here is that it would not fail properly when the target node is not found. Now the link is created for the actual target page (where the component is located) and it will fail when the component is not found.
- `StrictMode` has been removed.. :cry: This had to be done because it would mess up node generation, as it always calls `setState()` twice, and renders all components twice. Since it also runs `useEffect()` unmount callbacks once, it could have the effect of removing a node from the node store as soon as it was added, or attempt to add a node that already existed. The new `GeneratorStages` plays a few tricks trying (but nothing dirty!) to get all the `setState()` calls to happen in the planned order, and duplicate renders via `StrictMode` would ruin that. Sadly it's not possibly to selective enable `StrictMode` for parts of the application. Luckily I also never saw StrictMode helping us find any weird bugs we introducted while we had it enabled, so I hope it's not a big loss.
- Fixes for loading + caching initial form data. The form data query would sometimes load stale data, but it was initially meant that `InitialFormData` should always fetch a fresh set of form data each time it was rendered anew. This would also mean that prefetching does not make sense for form data (as it would just fetch twice). At least one cypress test was found to crash (but after all assertions were successful) because of this, so I added a gloal `afterEach()` for all cypress tests that makes sure the app did not crash in the end (call `cy.allowFailureOnEnd()` if you want that to happen).
- Quite a few Cypress tests got flaky after this. Sometimes because they attempted to perform actions a bit too quickly (like clicking on things before `useEffects()` in the node generator had run successfully to set the needed state), but also because app-frontend is now fast enough that existing tests would fall apart (such as `tabbing.ts` which tried tabbing to the next component, but it now managed to do so before that component showed up at all).

## Related Issue(s)

- closes #1182 
- closes #1938
- closes #1872

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels


cammiida (commented):
Må ta review her litt i omganger, merker jeg. Altfor mye å ta på én gang 🙈

src/features/routing/AppRoutingContext.tsx
```diff
@@ -0,0 +1,126 @@
...
+  const queryKeys = useLocation().search ?? '';
+  const updateQueryKeys = useSelector((ctx) => ctx.updateQueryKeys);
+
+  useEffect(() => {
+    const map = Object.fromEntries(new URLSearchParams(queryKeys).entries());
+    updateQueryKeys(map);
+  }, [queryKeys, updateQueryKeys]);
+
+  return null;
+}
```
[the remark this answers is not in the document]
olemartinorg: Tja, nja, URLen er fin til å holde på state - men langt ifra like god som zustand til å hindre re-rendring.. :wink: Dette gjorde jeg fordi steder som leste ut query-keys (som f.eks. om man har satt `?pdf=1`) måtte re-rendre hver gang man byttet side - siden det også er en del av URLen. Så med dette kan man slippe en hel del unødvendig re-rendring på grunn av urelaterte endringer i URLen. Før jeg gjorde dette tok det merkbart lang tid å bytte side, for veldig mye måtte gjøres helt om igjen for hvert sidebytte.

src/core/ui/RenderStart.tsx
```diff
@@ -18,9 +20,24 @@ interface Props extends PropsWithChildren {
...
+  const navigationEffect = useNavigationEffect();
+
+  useEffect(() => {
+    if (!isLoading && navigationEffect) {
+      navigationEffect();
+    }
+  }, [isLoading, navigationEffect, pageKey]);
+
+  return null;
+}
```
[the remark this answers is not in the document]
olemartinorg: Denne flytta jeg vel bort fra `PresentationComponent` og inn hit istedenfor. Tanken med `RenderStart` var nemlig å ha slike ting der (ting som må kjøre innenfor i alle providere, uavhengig av hvor mange det er). Jeg har ikke tatt stilling til om det trengs, men dette er knyttet til faktisk funksjonalitet, så vi kan ikke bare fjerne det heller. Mener at dette en gang ble laget for å fikse en WCAG-feil, ved at vi setter fokus på riktig element etter side-navigering.

src/features/routing/AppRoutingContext.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+function queryKeysToString(qc: Context['queryKeys']): string {
+  const keys = Object.keys(qc);
+  if (keys.length === 0) {
+    return '';
+  }
+
+  const keysAndValues = keys.map((key) => `${key}=${encodeURIComponent(qc[key]!)}`);
+  return `?${keysAndValues.join('&')}`;
+}
```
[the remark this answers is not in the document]
olemartinorg: Nice, takker! Jeg måtte forsåvidt beholde funksjonen, for jeg vil fortsatt ha funksjonaliteten for å returnere tom streng dersom det ikke er noen keys. :blush:

src/codegen/ComponentConfig.ts
cammiida, line 1: Dette er veldig vanskelig å forstå og introduserer enda mer kompleksitet enn det vi allerede har. Det er heller ikke idiomatisk react. Hvorfor skal funksjonalitet på en komponent defineres gjennom CG når vi har kontroll på dem selv? Kan du heller se på måter å frikoble funksjonalitet fra å gå gjennom samme klasser og slikt? Siden vi styrer intern config selv trenger ikke den å gå gjennom samme generering som ekstern config fra TE, men kan bare kobles direkte på hver komponent.

Dette kan være en mer funksjonell approach der hver komponent kobler seg på det den trenger og bare får inn config fra TE og dataflyten er lett å se.

cammiida (commented):
Dette funker, men trenger det å være så komplisert? Jeg vet du har lagt ned mye jobb, og imponerende at du har fått det til å kjøre, men jeg tror dessverre dette vil skape mye teknisk gjeld og gjøre det vanskeligere for oss og nye folk som kommer inn i fremtiden.

Noen punkter:
- Måten komponenter blir implisitt konfigurert på gjennom CG er veldig kompleks. Vi mister masse støtte når vi definerer JSX i strenger og det er ekstremt vanskelig å følge flyten. Dessuten er konfigurasjonen av en komponent nå en kombinasjon av det TE kan konfigurere og interne kapabiliteter vi setter på komponenten. Det burde være et skille der. Når det gjelder intern config ser jeg ingen grunn til å ha det rett på komponentene. Det er mer idiomatisk React, tillater oss mer fleksibilitet og gjør det lettere å lese og forstå flyten.
- Som kommentert tidligere: et er veldig lite anbefalt å omgå vanlig web-funksjonalitet som du har gjort ved å feks hindre rerendring ved bytting av side. For meg er det et tegn på at man har tatt valg som ikke er optimale som man heller bør se på. Min filosofi: alltid prøv å jobbe etter hvordan webben funker, ikke ens egen javascript. Det er også filosofien til feks Remix / React Router.
- Å fjerne `StrictMode` er et major rødt flagg. Det sørger blant annet for at komponentene dine er pure + andre checks som catcher potensielle bugs. Om man “må” fjerne `StrictMode` mener jeg man har gjort et feil arkitekturvalg. 
- Som vi har diskutert tidligere: Tanstack inni Zustand inni Context er ikke er et pattern. Ingen av teknologiene er brukt til det fulle eller på den måten de er ment å brukes, noe som gjør det vanskelig og at man må hele tiden jobbe rundt dem. I tillegg til at dataflyten går på kryss og tvers og veldig vanskelig å følge eller debugge. Jeg har god erfaring med Tanstack og ser fortsatt ingen grunn til å ikke bare ha Tanstack, eventuelt en ellet to Zustand stores. Og kanskje et par Contexter, men da ikke på globalt nivå.

olemartinorg:
@cammiida

> Dette funker, men trenger det å være så komplisert? Jeg vet du har lagt ned mye jobb, og imponerende at du har fått det til å kjøre, men jeg tror dessverre dette vil skape mye teknisk gjeld og gjøre det vanskeligere for oss og nye folk som kommer inn i fremtiden.

Forslag for hvordan dette kan forenkles (men hvor vi ikke mister funksjonalitet eller ytelse) mottas selvfølgelig med takk! :pray: Det jeg har landa på, og som jeg oppretter PR med, er jo en løsning jeg har sett kan fungere, og som er rask (når den først har startet opp), men det kan såklart hende det kan komme både forenklinger, forbedringer, og fullstendinge omskrivinger i fremtiden. Dette er vel omtrent generasjon 3 av hierarki-generatoren, så jeg har nok tatt med meg mange erfaringer fra tidligere som nok gjør at denne virker mer komplisert enn den skulle trenge å være.

> Måten komponenter blir implisitt konfigurert på gjennom CG er veldig kompleks. Vi mister masse støtte når vi definerer JSX i strenger og det er ekstremt vanskelig å følge flyten.

Hva er det som blir implisitt konfigurert? Tenker på at komponenter i `Form`-kategorien automatisk får egenskaper som `readOnly` og `required`, eller tenker du på at når du legger til en plugin så genererer den kode for deg? Førstnevnte er ikke noe nytt her, og sistnevnte tenker jeg jo er hele poenget med å lage plugins. Nemlig at man konfigurerer det opp en gang (i `config.ts`) og så fører det til effekter hvor funksjonalitet blir lagt til i `config.def.generated.tsx` og `config.generated.ts`.

> Dessuten er konfigurasjonen av en komponent nå en kombinasjon av det TE kan konfigurere og interne kapabiliteter vi setter på komponenten. Det burde være et skille der. Når det gjelder intern config ser jeg ingen grunn til å ha det rett på komponentene. Det er mer idiomatisk React, tillater oss mer fleksibilitet og gjør det lettere å lese og forstå flyten.

Tja, matter of taste? :sweat_smile: Adam sa det godt; dette er som vårt interne rammeverk. Vi har hatt fullt frislipp og prop-drilling overalt tidligere i v3, og da syntes jeg det var vanskelig å forstå flyten - og spesielt vanskelig å oppdage og fikse problemer som oppstod fordi props ble overstyrt hit og dit, eller bare glemt. Å ha bare et sted hvor både interne og eksterne egenskaper konfigureres opp synes jeg er en styrke, ikke en svakhet.

> Som kommentert tidligere: et er veldig lite anbefalt å omgå vanlig web-funksjonalitet som du har gjort ved å feks hindre rerendring ved bytting av side. For meg er det et tegn på at man har tatt valg som ikke er optimale som man heller bør se på. Min filosofi: alltid prøv å jobbe etter hvordan webben funker, ikke ens egen javascript. Det er også filosofien til feks Remix / React Router.

Å si "hindre rerendring ved bytting av side" er jo en annen måte å si "unngå unødvendig re-rendring når man lytter på helt urelatert state i URLen" på. Mye av oppgaven her gikk ut på å forbedre ytelse, så jeg har gjort det som trengs for å få det til. I disse tilfellene bygget jeg over react-router og pakket inn vår state, og det løste et veldig konkret og veldig tydelig ytelsesproblem - som vi garantert hadde fått tilbakemelding om fra tjenesteeiere om vi ikke hadde fikset det. Jeg nekter å si "vi har ikke fått verktøyene til å løse dette problemet av react-router" og så prøve å leve med dårlig ytelse.

Og hva er "vanlig web-funksjonalitet" og "hvordan webben funker", egentlig? Hvis du tenker på at URL endres = alt må re-rendres, har vi jo brutt med det prinsippet med en gang vi laget en SPA. Legg også merke til at det finnes mange issues hos react-router hvor folk klager på at det er laget slik, for det fører med seg mye unødvendig re-rendring.

> Å fjerne StrictMode er et major rødt flagg. Det sørger blant annet for at komponentene dine er pure + andre checks som catcher potensielle bugs. Om man “må” fjerne StrictMode mener jeg man har gjort et feil arkitekturvalg.

Forsåvidt enig, men problemet her er egentlig også veldig enkelt. Når du har skrudd på StrictMode, kjører alle `useEffect()` to ganger. Så for node-hierarkiet blir det omtrent slik:

```javascript
  const [state, setState] = useState({
    nodes: {},
  });

  useEffect(() => {
    setState((prev) => {
      if (prev.nodes["abc123"]) {
        throw new Error("Node already exists, cannot overwrite");
      }

      return { ...prev, nodes: { ...prev.nodes, abc123: { id: "abc123" } } };
    });
  }, []);
````

Test ut [her i codesandbox](https://codesandbox.io/p/sandbox/xenodochial-euler-q65rms). Siden det å overskrive state for en node er et alvorlig problem jeg prøvde hardt å unngå at skulle skje, kræsjet dette også med dobbel-kjøringen til StrictMode. Tidligere var det forsåvidt også slik at en node ble fjernet fra staten ved unmount, og da var problemet enda verre - da ble noder fjernet før man begynte å skrive state til dem. Alt ble super-problematisk med StrictMode, egentlig. Meeen, jeg tenker nok det kan være verdt et forsøk å se på dette på nytt, og muligens få inn igjen StrictMode. Det vil jeg uansett tenke meg er lavpri - for nå handler det aller mest om å slutte med småting og dill, og faktisk få ut denne endringen. Attpåtil er det ikke et stort savn hos meg - etter at jeg skrudde på StrictMode har jeg aldri hørt noen som har funnet hittil ukjente feil som følge av den.

> Som vi har diskutert tidligere: Tanstack inni Zustand inni Context er ikke er et pattern. Ingen av teknologiene er brukt til det fulle eller på den måten de er ment å brukes, noe som gjør det vanskelig og at man må hele tiden jobbe rundt dem. I tillegg til at dataflyten går på kryss og tvers og veldig vanskelig å følge eller debugge. Jeg har god erfaring med Tanstack og ser fortsatt ingen grunn til å ikke bare ha Tanstack, eventuelt en ellet to Zustand stores. Og kanskje et par Contexter, men da ikke på globalt nivå.

Jeg jobber vel generelt ikke etter oppskriftsbok, så at dette ikke er dokumentert som et pattern noe sted bekymrer meg ikke så veldig. Hvem bestemmer hva som er et pattern, og da tilsynelatende hva som er lov og ikke lov å gjøre? :thinking: Hvis det er bloggposter og eksempler i dokumentasjon man går etter er jeg heller ikke så bekymret. Jeg har fokusert på å løse konkrete problemer her, og problemene har jeg løst, så om ingen andre har funnet opp dette hjulet før meg tar jeg det heller som et kompliment. Såvidt jeg vet er det tydelig dokumentert at Zustand er unopinionated, så jeg regner egentlig med at påstandende her er mest dine egne.

At du synes det er vanskelig å følge dataflyten, og debugge, er heller noe vi bør gjøre noe med. Men da tenker jeg vi må starte med å øke forståelse fremfor å kaste denne endringen ut vinduet. At debugging blir vanskelig etter denne endringen kan vi jo ikke si så mye om enda - denne endringen har jo foreløpig ikke blitt merget. Når ting laget her skal endres, la oss først bygge opp erfaringer og endre ting når man har konkrete forslag om hvordan.

File: src/features/expressions/validation.ts [hunk 4]
```diff
@@ -235,83 +216,81 @@ export function asExpression(
...
-  const pathStr = componentPath.join('.');
-  const cfg = getConfigFor(componentPath, config);
-  if (cfg) {
-    if (typeof input === 'object' && input !== null) {
-      const errText = `Invalid expression when parsing ${pathStr} for "${componentId}"`;
-      return asExpression(input, cfg, errText);
-    }
-
-    return input;
+function isScalar(val: unknown, type: ExprVal | undefined) {
```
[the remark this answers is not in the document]
olemartinorg: De gjør litt forskjellige ting, og er vel knyttet til konteksten av koden rundt. Derfor er vel heller ikke noen av dem eksportert. Jeg foretrekker å ha disse to separat fremfor å prøve å bygge en ny (eksportert) funksjon som skal gjøre to forskjellige ting.

src/features/expressions/validation.ts
```diff
@@ -235,83 +216,85 @@ export function asExpression(
...
+  return isScalar(obj, type) || isValidExpr(obj, errorText);
+}
 
-  for (const comp of layout) {
-    preProcessItem(comp, config, [], comp.id);
+function throwIfInvalid<EV extends ExprVal>(obj: unknown, type: EV, errorText?: string): void;
+// eslint-disable-next-line no-redeclare
+function throwIfInvalid(obj: unknown, type?: undefined, errorText?: string): void;
+// eslint-disable-next-line no-redeclare
+function throwIfInvalid(obj: unknown, type?: ExprVal, errorText?: string): void {
```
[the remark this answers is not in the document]
olemartinorg: Den blir brukt i et par tester også, ref kommentaren der hvor den blir eksportert (se `ExprValidation` under).

src/features/alertOnChange/AlertOnChangePlugin.tsx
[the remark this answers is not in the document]
olemartinorg: Æsj, nå forsvant konteksten av hvilken linje du kommenterte på, men jeg tolker det slik at du mente hele `AlertOnChangePlugin`. Den blir jo ikke bare brukt i kodegeneratoren, dette er funksjonalitet som har blitt brukt i flere ulike layout-komponenter. Og ja, den eksponerer også config til TE; en egenskap på komponenter som vanligvis heter `alertOnChange` eller `alertOnDelete`.

Vanligvis blir dette brukt slik at man setter `"alertOnChange": true` på f.eks. en `Checkboxes`-komponent, og da får man opp en advarsel om man prøver å fjerne en avkryssing man tidligere har gjort. Tjenesteeiere bruker gjerne dette dersom de har satt på logikk i backend som f.eks. oppretter en rad i en senere repeterende gruppe på bakgrunn av en checkbox - for da blir brukeren varslet om de går tilbake i skjemaet og fjerner en avkryssing som potensielt fører til sletting av data senere i skjemaet. Se #1233, #1558 og #2140.

src/features/attachments/AttachmentsStorePlugin.tsx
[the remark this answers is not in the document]
olemartinorg: Hmm, ja, det er jo en store-plugin, altså en plugin til `NodesContext`. Jeg syntes det gav mest mening å legge denne funksjonaliteten utenfor `NodesContext`, slik at jeg ikke blander inn attachments, validering osv inn i den samme suppa. Men de er jo del av det samme resultatet til slutt - selv om koden er plassert sammen med med andre vedleggsrelaterte ting i `src/features/attachments`.

Så, bare fordi det er laget støtte for en type plugins her, betyr det ikke at ordet 'plugin' er reservert i resten av prosjektet, tenker jeg. Vi har jo sånnsett samme utfordring med begrepet 'komponent', siden det både kan referere til 'react-komponent' men også 'layout-komponent'. Om du har forslag til et annet begrep å bruke istedenfor `NodeDefPlugin` og `NodeDataPlugin`, er det bare å hyle ut! :raised_hands:

src/codegen/ComponentConfig.ts
```diff
@@ -187,26 +263,215 @@ export class ComponentConfig extends GenerateComponentLike {
...
+      const extraMethodsFromPlugin = plugin.extraMethodsInDef();
+      additionalMethods.push(...extraMethodsFromPlugin);
+
+      const extraInEval = plugin.extraInEvalExpressions();
+      extraInEval && evalLines.push(extraInEval);
+    }
+
+    const childrenPlugins = this.plugins.filter((plugin) =>
+      isNodeDefChildrenPlugin(plugin),
+    ) as unknown as (NodeDefChildrenPlugin<any> & NodeDefPlugin<any>)[];
```
bjosttveit, line 388: I am curious if this could now work without a type assertion since we recently upgraded to TS 5.5, which has type inference on array.filter? 🤔

src/features/devtools/layoutValidation/useLayoutValidation.tsx
```diff
@@ -1,155 +1,21 @@
...
-  const allNodeIdsRef = useRef(allNodesIds);
-
-  if (allNodesIds === allNodeIdsRef.current || deepEqual(allNodesIds, allNodeIdsRef.current)) {
-    return nodesRef.current;
-  } else {
-    nodesRef.current = nodes;
-    allNodeIdsRef.current = allNodesIds;
-    return nodes;
-  }
-}
```
bjosttveit, line 130: Hvis jeg forstår riktig så er ikke sånne stygge optimaliseringer lenger nødvendig? Fordi man faktisk kan "lytte" til endringer i hierarkiet med en traversal-selector, som ikke påvirkes av alt annet? Sånn som i pagination logikken for å finne ut om en side inneholder valideringsfeil?

I dette tilfelle er vel kanskje hele layout-validerings-logikken flyttet inn i hierarkigenereringen?

olemartinorg: > Hvis jeg forstår riktig så er ikke sånne stygge optimaliseringer lenger nødvendig?

Korrekt!

> Fordi man faktisk kan "lytte" til endringer i hierarkiet med en traversal-selector, som ikke påvirkes av alt annet?

Ja, korrekt - det vil si, en `traversalSelector()` kjører bare på nytt når det er en endring i antall komponenter (altså noen har blitt lagt til/fjernet). Men en hvilken som helst annen selector vil potensielt kjøre ved hver eneste endring, ja.

> Sånn som i pagination logikken for å finne ut om en side inneholder valideringsfeil?

Du tenker på `useRowStructure()`? Ja, den ble også fjernet her.

> I dette tilfelle er vel kanskje hele layout-validerings-logikken flyttet inn i hierarkigenereringen?

Korrekt. Det meste ligger nå i `src/utils/layout/generator/validation`. Det inkluderer jo også generell feilhåndtering i generatoren, samt feilhåndtering ved rendring av komponenter, så nå vil en exception bare føre til at en gitt komponent tryner.

src/codegen/ComponentConfig.ts
olemartinorg: Yup, it sure did! :star_struck:

src/features/displayData/index.ts
```diff
@@ -1,16 +1,20 @@
...
-  attachments: IAttachments;
-  optionsSelector: ReturnType<typeof useAllOptionsSelector>;
+  attachmentsSelector: AttachmentsSelector;
+  optionsSelector: NodeOptionsSelector;
   langTools: IUseLanguage;
   currentLanguage: string;
   formDataSelector: FormDataSelector;
+  nodeFormDataSelector: NodeFormDataSelector;
+  nodeDataSelector: NodeDataSelector;
 }
```
bjosttveit, line 18: Veldig bra at "alt" nå er selektorer, liker veldig godt det patternet for app-frontend, siden det er så mange (nødvendige) avhengigheter imellom ting, og da slipper vi masse unødvendige oppdateringer 💪

src/features/formData/FormDataWrite.tsx
```diff
@@ -210,7 +215,13 @@ function FormDataEffects() {
...
   // event is used to warn the user when they try to navigate away from the page with unsaved changes.
   useEffect(() => {
-    document.body.setAttribute('data-unsaved-changes', hasUnsavedChanges.toString());
+    if (hasUnsavedChanges) {
+      document.body.setAttribute('data-unsaved-changes', 'true');
+    } else {
+      setTimeout(() => {
+        document.body.setAttribute('data-unsaved-changes', 'false');
+      }, 10);
+    }
```
bjosttveit, line 224: Could this `setTimeout` when setting unsaved changes to false be problematic?

What if it changes back to true before the timeout fires? Would it not then be set to false in the document body once the timer expires, even though its suppose to be true?

Maybe the timeout should be kept track of, so it can be cleared when setting to true?

src/features/formData/FormDataWrite.tsx
```diff
@@ -224,11 +235,11 @@ function FormDataEffects() {
...
   useEffect(
     () => () => {
-      if (hasUnsavedChangesRef.current) {
+      if (hasUnsavedChangesRef.current && !isSavingNow()) {
         performSave();
       }
     },
-    [hasUnsavedChangesRef, performSave],
+    [hasUnsavedChangesRef, isSavingNow, performSave],
   );
```
bjosttveit, line 243: Does this need to check `isSavingNow()`? Is this not already checked by the `hasUnsavedChangesRef`? Maybe I dont understand exactly what causes this to update, but this should only happen when unmounting the component?

src/features/formData/FormDataWrite.tsx
olemartinorg: Absolutely! :raised_hands: Changed that now (but haven't pushed yet, as tests are currently in disarray while I fix some stuff properly).

src/features/formData/FormDataWrite.tsx
olemartinorg: You're probably right. The problem I've had with this state is that its way too often out of date, and will cause multiple saves. Maybe I've resorted to check things twice, and it's quite possible I forgot to remove this again when it didn't turn out to be fully working. I think the best solution so far has been to make `useHasUnsavedChangesRef` into an anonymous class that actually checks if there are unsaved changes at the point when you read `ref.current`. But I guess that could just as easily be a function call now, so I'll look into this some more when tests are stable again. :pray:

src/components/label/Label.tsx
```diff
@@ -7,38 +7,38 @@ import cn from 'classnames';
...
+  const _item = useNodeItem(node);
+  const item = { ..._item, ...overrideItemProps };
+  const { id: nodeId, grid, textResourceBindings: _trb } = item;
+  const required = 'required' in item && item.required;
+  const readOnly = 'readOnly' in item && item.readOnly;
+  const labelSettings = 'labelSettings' in item ? item.labelSettings : undefined;
+
+  // These can be overridden by props, but are otherwise retrieved from the node item
+  const id = rest.id ?? nodeId;
+  const textResourceBindings = (rest.textResourceBindings ?? _trb) as ExprResolved<TRBLabel> | undefined;
```
cammiida, line 40: TS casting resulterer fort i bugs fordi `rest.textResourceBindings` eller `_trb` plutselig kan være noe helt annet enn `ExprResolved<TRBLabel> | undefined`, og da finner man ikke ut av det før runtime :/ 
Så `as` burde egentlig aldri brukes, med mindre man vet 100% hvilken type noe er (og har compile time checks), men at TS bare ikke har klart å skjønne hvilken type det er.
Supert om du får til å fjerne dem (nye altså), så blir det også lettere å refaktorere senere 😃

src/features/expressions/index.ts
```diff
@@ -44,150 +40,27 @@ export interface EvalExprOptions {
...
+export type SimpleEval<T extends ExprVal> = (
+  expr: ExprValToActualOrExpr<T> | undefined,
+  defaultValue: ExprValToActual<T>,
+  dataSources?: Partial<ExpressionDataSources>,
+) => ExprValToActual<T>;
 
 /**
  * Run/evaluate an expression. You have to provide your own context containing functions for looking up external
  * values. If you need a more concrete implementation:
  * @see evalExprInObj
```
cammiida, line 52: Referansen eksisterer ikke lenger? 😄

src/features/expressions/index.ts
cammiida, line 1: Veldig mye enklere å forstå expressions nå 🙌

src/features/expressions/index.ts
```diff
@@ -44,150 +40,27 @@ export interface EvalExprOptions {
...
+  _expr: Expression | ExprValToActual | undefined,
   node: LayoutNode | LayoutPage | NodeNotFoundWithoutContext,
-  dataSources: ContextDataSources,
+  dataSources: ExpressionDataSources,
   options?: EvalExprOptions,
 ) {
+  if (_expr === undefined || _expr === null || ['string', 'number', 'boolean'].includes(typeof _expr)) {
+    return _expr;
+  }
+  const expr = _expr as Expression;
```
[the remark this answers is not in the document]
olemartinorg: Takker! :raised_hands: Jeg endret litt på funksjonen her da. Gyldige uttrykk må ha `input.length >= 1` (det absolutt minste uttrykket man kan skrive er et funksjonskall uten argumenter).

src/components/label/Label.tsx
olemartinorg: Ja, jeg tenkte litt på det, men jeg gjorde ikke så mye med det (også for å ikke bare introdusere enda mer greier her i denne PRen). Slik det var før fikk man heller ikke noe feil om man sendte inn `textResourceBindings` til en komponent-type som ikke har dem definert (siden `LabelProps` gjorde `title`, `description` og `help` valgfri, men en komponent helt uten de tekstressursbindingene ville da også tilsynelatende fungert.

Jeg ser for meg at jeg heller ville laget noe ala `CompWithBehavior` her, og sagt at en label bare kan få inn node-objekter for komponent-typer som har blitt konfigurert til å tillate `TRBLabel`. Først da får man en TS-feil om man har glemt å konfigurere opp label-tekstressurser på en komponent man vil rendre label for.. :thinking:

Jeg ordner et issue på dette istedenfor! :raised_hands:

olemartinorg: https://github.com/Altinn/app-frontend-react/issues/2315

src/features/formData/FormDataWrite.tsx
olemartinorg: Looks like that went well! :raised_hands:

bjosttveit (approved):
Haven't found anything else in the code and I will not look through it again 😅, so it's a green light from me whenever you are ready.

🙏 Well done 🙏

src/components/label/Label.tsx
cammiida: Denne kommentaren var ment for generelt bruk av `as` som er gjort mange andre steder også. Supert om du får fikset det 😊

PR #2307, opened 2024-08-16
Fix payment information component not fetching latest data on first render

### LINKED ISSUE #2277: Payment Details component doesn't fetch data on initial render
### Description of the bug

If you edit some data in the data model that should have an effect on order details, and then navigate to another page where the payment details component is, the payment details components doesn't reload order details data.

If you add some data on page 1 that should affect the price, and then go to page 2, where the component is, it will show a sum of 0 and no order lines.

This does not happen if you also have the component in page 1, where you are editing the data, but that's because the loading of order details is happening on page 1, and is then cached for when it should be shown in page 2.

### Steps To Reproduce

Setup:
1. Have an app with payment setup. Suggestion: https://dev.altinn.studio/repos/ttd/payment-test.
2. Have one page where you can edit some data that should affect the order details being calculated in the IOrderDetailsCalculator implementation of the app. **Make sure there is no PaymentDetails component in this page, or else you won't be able to reproduce the bug.**
3. Have another page with a PaymentDetails component.

Steps:
1. Fill out some data in the first page that should affect the price.
2. Navigate to the page with the payment details component. Observe that the sum is still 0. No data has been loaded.

### Additional Information

_No response_

Magnusrm:
## Description

Because the order details are prefetched when the form loads we have to refetch the data when mounting the component in order to not use the stale cached data the was prefetched.

## Related Issue(s)

- closes #2277 

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels

### DIFF
--- a/src/layout/PaymentDetails/PaymentDetailsComponent.tsx
+++ b/src/layout/PaymentDetails/PaymentDetailsComponent.tsx
@@ -20,6 +20,11 @@ export function PaymentDetailsComponent({ node }: IPaymentDetailsProps) {
   const mappedValues = FD.useMapping(mapping);
   const prevMappedValues = useRef<Record<string, unknown>>(mappedValues);
 
+  // refetch data on mount by invalidating cache as the first fetch is done by the formPrefetcher
+  useEffect(() => {
+    refetchOrderDetails();
+  });
+
   useEffect(() => {
     if (!hasUnsavedChanges && mapping && !deepEqual(prevMappedValues.current, mappedValues)) {
       refetchOrderDetails();
--- a/src/layout/PaymentDetails/PaymentDetailsComponent.tsx
+++ b/src/layout/PaymentDetails/PaymentDetailsComponent.tsx
@@ -22,7 +22,6 @@ export function PaymentDetailsComponent({ node }: IPaymentDetailsProps) {
 
   // refetch data on mount by invalidating cache as the first fetch is done by the formPrefetcher
   useEffect(() => {
-    console.log('refetchOrderDetails');
     refetchOrderDetails();
   }, [refetchOrderDetails]);
 
--- a/src/layout/PaymentDetails/PaymentDetailsComponent.tsx
+++ b/src/layout/PaymentDetails/PaymentDetailsComponent.tsx
@@ -22,8 +22,9 @@ export function PaymentDetailsComponent({ node }: IPaymentDetailsProps) {
 
   // refetch data on mount by invalidating cache as the first fetch is done by the formPrefetcher
   useEffect(() => {
+    console.log('refetchOrderDetails');
     refetchOrderDetails();
-  });
+  }, [refetchOrderDetails]);
 
   useEffect(() => {
     if (!hasUnsavedChanges && mapping && !deepEqual(prevMappedValues.current, mappedValues)) {


src/layout/PaymentDetails/PaymentDetailsComponent.tsx
```diff
@@ -20,6 +20,11 @@ export function PaymentDetailsComponent({ node }: IPaymentDetailsProps) {
   const mappedValues = FD.useMapping(mapping);
   const prevMappedValues = useRef<Record<string, unknown>>(mappedValues);
 
+  // refetch data on mount by invalidating cache as the first fetch is done by the formPrefetcher
+  useEffect(() => {
+    refetchOrderDetails();
+  });
+
```
bjosttveit, line 27: Doesn't this just fire on every render since there is no dependency array?

Magnusrm: Yep, once every render i guess. Should i handle it differently?

bjosttveit: The comment says it only refetches on mount, in that case you would want an empty dependency array?

src/layout/PaymentDetails/PaymentDetailsComponent.tsx
```diff
@@ -20,6 +20,11 @@ export function PaymentDetailsComponent({ node }: IPaymentDetailsProps) {
   const mappedValues = FD.useMapping(mapping);
   const prevMappedValues = useRef<Record<string, unknown>>(mappedValues);
```
bjosttveit, line 21: The previous value is currently initialized to the current value, so the existing use effect will not fire on mount. Would it work if you just initialized `prevMappedValues` as `undefined` instead?

Magnusrm:
@bjosttveit Ble en del styr med typeguarding når jeg initialiserte prevMappedValues som undefined. La til dependency array, men får klager om t den bør inneheolde noe eller fjernes. La til refetchOrderDetails og tenker det går fint ettersom den trolig ikke vil endre seg med mindre man rendrer hele appen på nytt.

bjosttveit:
> La til refetchOrderDetails og tenker det går fint ettersom den trolig ikke vil endre seg med mindre man rendrer hele appen på nytt.

Ja det bør gå fint siden den bruker en useCallback.

>Ble en del styr med typeguarding når jeg initialiserte prevMappedValues som undefined

Type guarding hvordan da? Jeg fikk i hvertfall ikke noen type errors ved å endre til dette:

```diff
+ const prevMappedValues = useRef<Record<string, unknown> | undefined>(undefined);
- const prevMappedValues = useRef<Record<string, unknown>>(mappedValues);
```

Magnusrm:
Aha jeg misforstod, jeg endra bare til `prevMappedValues = undefined;`

olemartinorg:
> **Because the order details are prefetched** when the form loads we have to **refetch the data when mounting** the component in order to **not use the stale cached data** the was prefetched

I'm not entirely up to date, but mayyybe the fix here is to just remove prefetching? What's the point in prefetching if you're throwing that away be refetching? :thinking:

Magnusrm:
@olemartinorg 
The pre-fetching was added in order to make sure data was loaded before the pdf was created: https://github.com/Altinn/app-frontend-react/pull/2103

I'll make sure to create a regression test! 🫡

bjosttveit:
> but mayyybe the fix here is to just remove prefetching? What's the point in prefetching if you're throwing that away be refetching? 🤔

I am not convinced that prefetching is the real issue here. Would you not have the same problem if you navigate to the payment component (load the query once), navigate away and change something, then navigate back to payment. Now, the query needs to refetch, but it will not do it because it initially compares the current data with itself. I think this would happen regardless of prefetching 🤷

src/layout/PaymentDetails/PaymentDetailsComponent.tsx
```diff
@@ -20,6 +20,11 @@ export function PaymentDetailsComponent({ node }: IPaymentDetailsProps) {
...
 
+  // refetch data on mount by invalidating cache as the first fetch is done by the formPrefetcher
+  useEffect(() => {
+    refetchOrderDetails();
+  }, [refetchOrderDetails]);
+
```
bjosttveit, line 27: En ting jeg akkurat kom på, er jo at man ikke trenger å refetche med mindre `mapping !== undefined` 🤔

Magnusrm: Sant 👍

PR #2126, opened 2024-06-03
Fix/kan-ikke-velge-underenhet-2033

HauklandJ:
## Description

The issue was that `validParties` did not account for childParties.

Set valid parties based on the configuration in application metadata. 
The app metdata defines "partyTypesAllowed" - use this to filter out disallowed types.

## Related Issue(s)

- closes #2033

## Verification/QA
https://github.com/Altinn/app-frontend-react/assets/42615991/aabcc079-2e6e-4139-8faf-76edf495a117
- Manual functionality testing
  - [x] I have tested these changes manually
  - [ ] Creator of the original issue (or service owner) has been contacted for manual testing (or will be contacted when released in alpha)
- Automated tests
  - [x] Unit test(s) have been added/updated
  - [ ] Cypress E2E test(s) have been added/updated
  - [ ] No automatic tests are needed here (no functional changes/additions)
  - [ ] I want someone to help me make some tests
- UU/WCAG ([follow these guidelines](https://aksel.nav.no/god-praksis/artikler/utvikling?tema=universell-utforming) until we have our own)
  - [ ] I have tested with a screen reader/keyboard navigation/automated wcag validator
  - [x] No testing done/necessary (no DOM/visual changes)
  - [ ] I want someone to help me perform accessibility testing
- User documentation @ [altinn-studio-docs](https://github.com/Altinn/altinn-studio-docs)
  - [ ] Has been added/updated
  
  - [x] No functionality has been changed/added, so no documentation is needed
  - [ ] I will do that later/have created an issue
  
- Support in Altinn Studio
  - [ ] Issue(s) created for support in Studio
  
  - [x] This change/feature does not require any changes to Altinn Studio
- Sprint board
  - [x] The original issue (or this PR itself) has been added to the Team Apps project and to the current sprint board
  - [ ] I don't have permissions to do that, please help me out
- Labels
  - [x] I have added a `kind/*` label to this PR for proper release notes grouping
  - [ ] I don't have permissions to add labels, please help me out

PR #2309, opened 2024-08-19
fix: use allowedPartyTypes to filter valid parties

mikaelrss:
## Description

Our form at Digital gravferdsmelding has suddenly startet displaying this error message locally, and it seems like this started in version 4.4.0.

image

I think the issue was introduced in this [PR](https://github.com/Altinn/app-frontend-react/pull/2126/files/9242729e088a86bf477dd28050870b716c9c6f09#diff-9c7e421849a215d3749ef23201a265f6f0202ad2ed7a562c332715aa615f648e) but I am not quite sure if this is a bug in the app frontend or if our app is providing incorrect data.

The issue occurs because when the frontend is reducing valid parties it filters each party on a condition `partyTypeFilters[party.partyTypeName]` from line 38 in `partyProviderUtils`. The issue with this is that `partyTypeFilters` is an object that is contructed in this util function like this: 
```tsx
const partyTypeFilters = {
  [PartyType.Organisation]: partyTypesAllowed.organisation,
  [PartyType.SubUnit]: partyTypesAllowed.subUnit,
  [PartyType.Person]: partyTypesAllowed.person,
  [PartyType.BankruptcyEstate]: partyTypesAllowed.bankruptcyEstate,
};
```

Where `PartyType` looks like this: 
```tsx 
export enum PartyType {
  Person = 1,
  Organisation = 2,
  SelfIdentified = 3,
  SubUnit = 4,
  BankruptcyEstate = 5,
}
```

This, in my case, evaluates to a filter like this: 
```tsx
const partyTypeFilters = {
  2: false,
  4: false,
  1: true,
  5: false,
};
```
However, `party.partyTypeName` is not an int representing one of these values, but rather a string `"person"`, which means that line 38 will always evaluate to `false` for every party, even though that party should be valid. 

So I am wondering if the `partyTypeName` should actually be the string `"person"`, and there is a bug in the code, or are we doing something incorrectly in our form that makes the `/parties`-endpoint return incorrect data?

EDIT: I see that the cypress tests are all failing, and that the types for `IParty` suggest that `partyTypeName` should in fact be `PartyTypes` (enum integer value). Can someone confirm if this is correct? :sweat_smile: 

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels

### DIFF
--- a/src/components/altinnParty.test.tsx
+++ b/src/components/altinnParty.test.tsx
@@ -7,6 +7,7 @@ import { userEvent } from '@testing-library/user-event';
 import { getPartyMock } from 'src/__mocks__/getPartyMock';
 import { AltinnParty } from 'src/components/altinnParty';
 import { renderWithInstanceAndLayout } from 'src/test/renderWithProviders';
+import { PartyType } from 'src/types/shared';
 import type { IAltinnPartyProps } from 'src/components/altinnParty';
 
 const user = userEvent.setup();
@@ -88,7 +89,7 @@ describe('altinnParty', () => {
         party: {
           ...getPartyMock(),
           orgNumber: '1000000',
-          partyTypeName: 2,
+          partyTypeName: PartyType.Organisation,
         },
       });
       expect(screen.getByTestId('org-icon')).toBeVisible();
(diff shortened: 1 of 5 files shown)


HauklandJ:
Closing this PR as further investigation shows the issue is not with the frontend, but rather not enforcing enum serialization policy in the backend, see [729](https://github.com/Altinn/app-lib-dotnet/issues/729).

PR #2324, opened 2024-08-21
Feat: External API

### LINKED ISSUE #2207: Støtte for externt api i expressions

cammiida:
## Description
Can be easiest to read this PR commit for commit. 

Adds support for referring to an external api in expressions.
Fetches all external api ids from application metadata and fetches the corresponding data from the relative endpoints on the backend.

## Related Issue(s)

- closes #2207 

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels


src/utils/layout/useExpressionDataSources.ts
```diff
@@ -37,6 +73,9 @@ export function useExpressionDataSources(): ExpressionDataSources {
   const nodeTraversal = useNodeTraversalSelectorLax();
   const transposeSelector = useDataModelBindingTranspose();
 
+  const externalApiIds = useApplicationMetadata().externalApiIds ?? [];
+  const externalApis = useExternalApis(externalApiIds);
```
olemartinorg, line 77: Hmm, this would mean that everyone using expression data sources (everywhere expressions are resolved) will  have to re-render whenever external API data sources change. Right now, it doesn't seem like cache invalidation and re-fetching is implemented though, so this is fine for now. :raised_hands:

I vaguely recall us discussing how that could be done, and what it would affect, but maybe we concluded to delay that until later? :thinking: Correct me if I'm wrong!

src/features/applicationMetadata/types.ts
```diff
@@ -12,26 +12,20 @@ export interface IncomingApplicationMetadata {
   org: string;
   partyTypesAllowed: IPartyTypesAllowed;
   dataTypes: IDataType[];
-  altinnNugetVersion?: string;
   autoDeleteOnProcessEnd: boolean;
-  onEntry?: IOnEntry;
   features?: Partial<IBackendFeaturesState>;
   promptForParty?: 'always' | 'never';
+  externalApiIds?: string[];
```
olemartinorg, line 18: Expanding the `applicationmetadata` file is a bit tricky, since so many different packages use this file, and we've seen before that we quickly end up in a situation where one package (i.e. Studio) removes unknown properties and re-deploys the file without them.

But what we _did_ do is adding custom stuff just between app-frontend-react and app-lib-dotnet, such as the `altinnNugetVersion` property. So when the backend reads the file from disk, it will add in bits and bobs we need, including that version number, feature flags enabled for the app, etc. So maybe that's the best way forwards? I.e. get the backend to generate a list of allowed external apis from what's implemented, and return those together with the rest of the metadata. Saves us a request as well. :raised_hands:

src/features/expressions/ExprContext.ts
```diff
@@ -1,148 +0,0 @@
...
-export interface PrettyErrorsOptions {
-  config?: ExprConfig;
-  introText?: string;
-}
-
-/**
- * The expression context object is passed around when executing/evaluating an expression, and is
- * a toolbox for expressions to resolve lookups in data sources, getting the current node, etc.
- */
-export class ExprContext {
```
olemartinorg, line 51: Good job cleaning this up! :star_struck: :clap:

src/features/expressions/expression-functions.ts
```diff
@@ -0,0 +1,605 @@
...
+        path && externalApiData && typeof externalApiData === 'object'
+          ? dot.pick(path, externalApiData)
+          : externalApiData;
+
+      if (!res) {
+        return ''; // Print error?
+      }
+
+      if (typeof res === 'object') {
+        return JSON.stringify(res);
```
olemartinorg, line 307: I would prefer returning null in that case. We want to add list and object support in the future (#1179), and it's best if app developers start getting used to this returning a JSON string they could try to substring their way out of, which we'll later swap out with real object/arrays... :scream: Currently, `dataModel` lookups also returns null when an object or array is found.

src/features/expressions/shared-tests/functions/externalApi/should-return-empty-string-on-undefined-result.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "Should return empty string on undefined result",
+  "expression": ["externalApi", "testId", "results[0].id"],
+  "expects": "",
```
olemartinorg, line 4: I prefer null in this case as well, which is more consistent with the rest of the expression language

src/features/expressions/shared-tests/functions/externalApi/should-pick-data-with-nested-path.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "Should pick data with nested path",
+  "expression": ["externalApi", "testId", "results[0].id"],
+  "expects": "1",
+  "externalApis": { "testId": { "data": { "results": [{ "id": "1" }] } } }
```
olemartinorg, line 5: Any reason this is packed inside `data` and `results`? :thinking: Can there be more information outside `data`? If it's a direct mapping to the results from `useQuery()`, we can't really tie this all to a frontend-only tech, as these expressions should run on the backend as well.

Also, from reading this and other tests, it kindof looks like all these looks have to start with `results...`? I would think less generically-sounding test data would resolve my misunderstanding, i.e. if the data were something like `{ "Cars": [{ "Brand": "BMW" }, { "Brand": "Lada" }] }`

src/features/expressions/shared-tests/functions/externalApi/should-return-empty-string-when-path-does-not-exist.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "Should return empty string on undefined result",
+  "expression": ["externalApi", "testId", "non.existent.path"],
+  "expects": "",
```
olemartinorg, line 4: Same here, this should also return null

olemartinorg (approved):
Great work!! :rocket: :clap:

src/features/applicationMetadata/types.ts
cammiida: Yes, that would be much better than having the application developers define it 👌

src/features/expressions/expression-functions.ts
cammiida: Makes sense, and probably results in less maintenance too.

src/utils/layout/useExpressionDataSources.ts
cammiida: Good point, I had totally forgotten to add `staleTime`. The data will only change when the cache entry is stale or invalidated, so I setting `staleTime` for this query to 10 min or something would ensure the data to be _relatively_ fresh whilst not having to update very often. But we probably have to have to dig a bit deeper soon to find out sane defaults, or maybe if we want the user to be able to set them 🤷‍♀️ 

What'chu think?

olemartinorg: The way it is now, that would mean that we'd have a once-every-10min full re-run of all expressions (in apps that use external APIs). Also, that might just make things harder to debug when we don't have any other mechanisms that behave like that. (I.e. "this is not right, let me just write a message about it to....wait, it fixed itself?? I must have been seeing things!")

We've used the `mapping` property (and the now more modern and capable `queryParameters`) for these things in components before, as a mechanism to invalidate caches and re-fetch options. I think we should introduce something later when (we know the use-cases better? or maybe that's just me). And at that point, it would be natural to use a (delayed) selector instead of the full query results - but, you know, that would also mean that we'd have to jam the query result into a zustand store.. :face_with_peeking_eye:

PR #2326, opened 2024-08-21
Add regression test for PaymentDetails component

Magnusrm:
## Description

Adding a cypress test make sure orderlines are fetched and displayed for PaymentDetails component.

## Related Issue(s)

- closes #{issue number}

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels

### DIFF
--- a/test/e2e/integration/payment-test/payment.ts
+++ b/test/e2e/integration/payment-test/payment.ts
@@ -77,4 +77,17 @@ describe('Payment', () => {
       });
     });
   });
+
+  describe('PaymentInformation component', () => {
+    it('should display the payment information', () => {
+      // hiding the payment details component on the page where the orderlines are created in order to test the
+      // payment information component fetching order details on first render
+      cy.interceptLayout('05GoodsAndServices', (component) => {
+        if (component.id === 'paymentDetails' && component.type === 'PaymentDetails') {
+          component.hidden = true;
+        }
+      });
+      cy.findByRole('row', { name: /1 - test 1 1000 NOK/ }).should('exist');
+    });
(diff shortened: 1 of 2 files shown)


test/e2e/integration/payment-test/payment.ts
```diff
@@ -77,4 +77,17 @@ describe('Payment', () => {
       });
     });
   });
+
+  describe('PaymentInformation component', () => {
+    it('should display the payment information', () => {
+      // hiding the payment details component on the page where the orderlines are created in order to test the
+      // payment information component fetching order details on first render
+      cy.interceptLayout('05GoodsAndServices', (component) => {
```
adamhaeger, line 85: Hmm why do you need to hide this? Arent we then messing with the layout? This state would never happen for the user right?

olemartinorg: Usually when I test things like this, I make `hidden` an expression instead. When it's just `true` the component will always be hidden, and you're just simulating an app where the component never shows up. What's more interesting to test is, I would assume, that if the expression result changes and the component now _should show up_, the query runs and the freshest data is fetched.

Magnusrm: I am not testing on the component i am hiding in this case. There exists another PaymentDetails component on the last page in the data task, where it is being rendered after the order details have changed after the initial pre-fetch. 

I am hiding this component because it exists on the same page where the order-details are being created and it has a hook that performs a re-fetch when the relevant datamodel fields change.

PR #2283, opened 2024-08-13
Add support for showing geometries in Map component

bjosttveit:
## Description

**Pre-release version**: `v4.8.0-geometrymap.preview2`

This PR adds support for showing geometries in either `GeoJSON` or `WKT` format in the Map component.

Geometries are read from the data model, by pointing to an array in `dataModelBindings` like this:
```json5
...
"dataModelBindings": {
  "geometries": "MapData.Geometries"
},
...
```

The array needs to be defined in the data model as follows, having the properties `data`, and `label` as strings:

```json5
...
"type": "array",
"items": {
  "type": "object",
  "properties": {
    "data": { "type": "string" },
    "label": { "type": "string" }
  }
}
...
```

The `data` field can contain either `WKT` or stringified `GeoJSON`. The component tries to parse as `GeoJSON` by default, but can be changed by setting the following property on the Map component:
```json5
...
"geometryType": "WKT",
...
```

The component will center and zoom the map to fit the geometries. This will override any `centerLocation` and `zoom` set on the component.

### Example configuration

```json5
{
  "id": "map",
  "type": "Map",
  "textResourceBindings": {
    "title": "Områder",
  },
  "dataModelBindings": {
    "geometries": "MapData.Geometries"
  },
  "geometryType": "WKT",
},
```

bilde

Note that the tooltips with the label are always visible, this is because the geometries become almost invisible when zooming out far enough.

### Other changes
- Moved map implementation from `altinn-design-system` to this repo
- Made `simpleBinding` (and setting a marker) optional
- Made marker draggable
- Map is no longer faded in PDF, only in regular summary
- Attribution (copyright notice) is styled better for PDF
- Set minimum zoom and max bounds so it looks less buggy when moving around
- Added print media emulation and snapshots for PDF tests, so we can actually catch mistakes with the PDF view. The map marker was at one point missing from the PDF because it had `role=button`, and we hide buttons by default in PDF, things like this should now be caught by percy. Also setting the viewport to simulate A4 paper size before taking the snapshots but after #readyForPrint is visible. This way we simulate how the PDF-generator works as well.
  - **Edit**: It seems like percy does not actually take a screenshot but rather a snapshot of the dom and renders that in its own service, so the print emulation is not actually active in the snapshots.
- Added cypress tests for old and new functionality of the map, and also added map to existing PDF tests.

## Related Issue(s)

- closes https://github.com/Altinn/altinn-studio/issues/12984

## Verification/QA

- Manual functionality testing
[...]

PR #2446, opened 2024-09-13
Direct Percy snapshot method to capture PDF with emulated print media

bjosttveit:
## Description

In #2283, snapshots were added to PDF testing in order to catch visual bugs in PDF. Additionally, print media emulation was added to PDF testing so that css using `@media print` will be applied. This is sometimes used to hide elements, and so it then became possible to test that elements that should be hidden actually are hidden in PDF.

However, the combination of print media emulation and snapshots was not working in this version. This is due to how Percy takes snapshots. Percy does not take screenshots of the browser window directly, but rather, it stores a snapshot of the DOM, uploads this to its own servers, puts this DOM snapshot into its own browser, and takes screenshots at different viewport sizes there. Because of this, we loose the context of our own browser which uses print media emulation.

To fix this we need to take a screenshot directly in the browser running cypress, and somehow upload this image to Percy. This is not something that Percy supports. So instead, I take a screenshot, convert it to a base64 data url, and change the whole document to only contain an image tag with the dataurl as the `src`. Percy will happily upload this DOM-snapshot and take its own screenshot of my screenshot.

This may seem like a hacky solution, but percy actually has a CLI-tool that allows you to upload a set of existing screenshot images, and would you guess how it does this? By putting the image into an image tag and uploading the HTML: https://github.com/percy/cli/blob/master/packages/cli-upload/src/utils.js 🤦 My solution is based on the Percy CLI-tool, so I am pretty sure this is the only proper way to achieve this, and it seems to work perfectly well. Though, I wish I could simply upload an image directly...

Here is an [example](https://percy.io/700b0900/altinn-app-frontend-react/builds/36373331/changed/1988767608?browser=chrome&browser_ids=59&device_name=&group_snapshots_by=similar_diff&os=&other-browser=&subcategories=unreviewed%2Cchanges_requested%2Capproved&utm_source=github_status_public&viewLayout=overlay&viewMode=new&width=794&widths=794%2C1280) of what the PDF snapshots look like now. You can tell that print styling is applied due to it now having a white background instead of a gray background.

## Related Issue(s)

- ~~closes #{issue number}~~

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels

### DIFF
--- a/test/e2e/support/custom.ts
+++ b/test/e2e/support/custom.ts
@@ -10,6 +10,7 @@ import { breakpoints } from 'src/hooks/useDeviceWidths';
 import { getInstanceIdRegExp } from 'src/utils/instanceIdRegExp';
 import type { LayoutContextValue } from 'src/features/form/layout/LayoutsContext';
 import JQueryWithSelector = Cypress.JQueryWithSelector;
+
 import type { BackendValidationIssue } from 'src/features/validation';
 import type { ILayoutFile } from 'src/layout/common.generated';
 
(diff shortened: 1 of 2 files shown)


Magnusrm (approved):
Nice, super useful! 👑

PR #2453, opened 2024-09-13
Fixing missing nodes in repeating group when UUID has changed

### LINKED ISSUE #2421: Nested group components are not showing when deleting a previously added repeating group index. Stopped working in v.4.8
### Description of the bug

When adding two repeating group indexes (includes nested groups) and you delete the first added repeating group index, the nested group information in the second repeating group index will not show, unless you do a hard refresh. 

**What it looks like in the second repeating group index in the nested group after deleting the first repeating group index:**
image
**When adding a new index to the nested group:**
image
**After hard refresh:**
image

**Note** that this is only happening when deleting an earlier repeating group index. E.g. if you have index 1 and 2 and delete index 1. 
I tried different versions of frontend and noticed that this works fine in v.4.7.0, and stops working when using v.4.8 and above/including preview versions. 

### Steps To Reproduce

1. Add a repeating group with one or more nested repeating groups.
2. Add two indexes of the repeating group with information in the nested groups
3. Delete the first index.

### Additional Information

_No response_

olemartinorg:
## Description

Late in the node generator rewrite, I looked at nodes in repeating groups, which have their IDs suffixed to include the row-index they're in (so `name-0`, `name-1` etc). I had, up until that point, included the row UUID in the data the node generator was working with, but when the results of that experiment somewhat broke, I removed the UUID part. That means, `name-0` points to the first row, and if you change the order of these rows (on the backend), `name-0` will still just point to the first row - but the data in that row now changed. Row UUIDs were then primarily only used for patch generation, to compare two different rows to figure out if something _changed_ or if the entire row was _replaced_.

Anyway, I didn't fully remove the UUIDs from the node generator, so in some cases you could end up with a changing UUID (which could happen because of a deleted previous row, as in the bug report, and also via a full data model changeup via a CustomButton, as in this new Cypress test I added). That would call a useEffect() cleanup function that would remove the row from the parent node, but it would never be added back again.

This fix removes the UUID from the node generator entirely, and just updates it in a separate component, with a separate set of requests.

It also includes minor fixes for other things:
- If you used a `component` lookup in a `hidden` expression on a page, that expression could end up running before all the components were added to the store, breaking the expression. Solved by wrapping it in a `GeneratorCondition`
- Refactoring the effects in `useGetOptions` to make them all work in the generator and not have them rely on rendering the target component.
- Making `useEffectRemoveStaleValues` a bit slower, to prevent it from deleting stuff from nodes that may soon become hidden. I did this as the new test case I was writing in Cypress broke when you clicked a `CustomButton` to reset the form data back to the original state it was in. In that case, if the set of components that were hidden would change as a result of the `CustomButton` click, you might end up deleting form data from the new model you just received from the backend because app-frontend had _not yet figured out_ that the component performing the deletion was actually hidden.
- Removing node traversal in `useTableNodes()`
- Returning `undefined` when the node state is not yet ready to know if a component is hidden
- In `ttd/frontend-test` a pattern emerged where we'd add an initially hidden page that could be unlocked by checking it in a `Checkboxes` component on the first page. This was useful to add more test-case 'arenas' without affecting other tests. This PR adds a shortcut to reach such pages via `cy.gotoHiddenPage()`. 

## Related Issue(s)

- fixes #2421

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels
  
…case' pattern

### DIFF
--- a/src/features/options/useGetOptions.ts
+++ b/src/features/options/useGetOptions.ts
@@ -1,5 +1,8 @@
 import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 
+import deepEqual from 'fast-deep-equal';
+
+import { DEFAULT_DEBOUNCE_TIMEOUT } from 'src/features/formData/types';
 import { useDataModelBindings } from 'src/features/formData/useDataModelBindings';
 import { useCurrentLanguage } from 'src/features/language/LanguageProvider';
 import { useLanguage } from 'src/features/language/useLanguage';
@@ -67,24 +70,9 @@ interface EffectProps {
   setValue: (values: string[]) => void;
   isNodeHidden: boolean | undefined;
   isNodesReady: boolean;
+  dataModelBindings: IDataModelBindingsOptionsSimple | IDataModelBindingsSimple | undefined;
+  valueType: OptionsValueType;
 }
-
-const getLabelsForActiveOptions = (selectedOptions: string[], allOptions: IOptionInternal[]): string[] =>
-  allOptions.filter((option) => selectedOptions.includes(option.value)).map((option) => option.label);
-
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-const usePrevious = (value: any) => {
-  const ref = useRef();
-  useEffect(() => {
-    ref.current = value;
-  });
-  return ref.current;
-};
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-const useHasChanged = (val: any) => {
-  const prevVal = usePrevious(val);
-  return prevVal !== val;
-};
 
 const defaultOptions: IOptionInternal[] = [];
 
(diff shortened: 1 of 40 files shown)


src/features/options/useGetOptions.ts
```diff
@@ -190,42 +156,124 @@ function usePreselectedOptionIndex(props: EffectProps) {
...
+      if (!options || !isReady()) {
+        return;
+      }
+      const itemsToRemove = unsafeSelectedValues.filter((v) => !options?.find((option) => option.value === v));
+      if (itemsToRemove.length > 0) {
+        setValues(unsafeSelectedValues.filter((v) => !itemsToRemove.includes(v)));
+      }
+    }, 200);
+  }, [getAwaiting, readyAsRef, propsAsRef, ready]);
+}
```
bjosttveit, line 202: I am not sure if this is a problem or not, but whenever I see setTimeouts in a use effect that are not cleared, I start thinking about what would happen if the dependencies were to change in rapid succession. I dont know how likely this is to happen though, as I dont know exactly what would cause `ready` to change.

- In the first case, would it force rerender many times? Is that necessary? or could it clear the existing timer and just postpone the rerender instead?
- If it ends up in the second timeout, I guess it would not likely call setValues several times since it is using a ref, but is it necessary to try to remove stale values several times? It only depends on the current options and selected values 🤔

olemartinorg: Great points! I should have cleaned up after my timeouts here. Adding that now, re-running the tests, and hoping for the best! :crossed_fingers:

olemartinorg: Looks good now! Marking this as resolved. :muscle:

bjosttveit (approved):
## 🙌

Issue #2148, opened 2024-06-10
Main form must validate all sub form data before submission


HauklandJ:
## Update
The below description and diagram is possibly outdated. Read comments and investigate as required before tackling this task.

## Blocked by
- https://github.com/Altinn/app-frontend-react/issues/1484

## Description
Part of the [Underskjema Epic](https://github.com/Altinn/app-frontend-react/issues/2137)

A user should NOT be able to submit a form if one or more sub forms are invalid. Two possible solutions:

1. Validate form and sub forms on `process next`
2. Validate sub forms on `sub form close` and check validation state of sub forms + validate form on `process next`

![validation2 drawio](https://github.com/Altinn/app-frontend-react/assets/42615991/593c14dd-1bb6-4eff-a6e7-f821a96e2596)

Option 1 is straight-forward, but may lead to performance issues in the case of many sub forms.

Option 2 is more complex, but would be a better user experience in the case of many sub forms. It would include computing and persisting a hash of the validated sub form on `sub form close` and then computing hashes of the sub forms on `form submit` and verifying against the persisted sub form hashes

bjosttveit:
Just throwing some thoughts on this:

The way validation in frontend works currently, is that we validate everything initially, and then incrementally as data in the form changes. We dont actually validate explicitly before submitting, we just check the current validation state which should be up to date at that point. However, the backend will validate everything when we call process next. I am assuming this will be the case for all sub forms in the task as well?

I am not exactly sure what the best model for sub forms would be. What if you refresh the page before submitting? Then you would not have any state from having been inside of the sub form and closing it, so there probably needs to be some mechanism for validating all sub forms from the "outside"? But whether it should be done on submit or in a similar way to the existing validations I do not know.

HauklandJ:
> Just throwing out some thoughts on this:

> The way validation in frontend works currently, is that we validate everything initially, and then incrementally as data in the form changes. We dont actually validate explicitly before submitting, we just check the current validation state which should be up to date at that point. However, the backend will validate everything when we call process next. I am assuming this will be the case for all sub forms in the task as well?
> 
We toying with the possibility of validating via backend when a sub form is closed with the "Add/Legg til" button. This would include a hash of the sub form content. Whether this is a performance boost that is needed is something we value input on, given the extra complexity introduced.
[...]

PR #2479, opened 2024-09-23
Validate subform data in main form

bjosttveit:
## Description

This adds validation of all sub-forms in the main form. To accomplish this in a relatively straight forward manner, we rely heavily on backend validation instead of trying to run frontend-validations as this would be a much bigger task. The main changes are the following:

- Changed how we store datamodel-validations in the `validationContext` to store per data element instead of per data type. This allows us to store all backend validations for all subforms when calling `/validate`. This requires some small changes to how we then lookup validations to store them in nodes, as well as how we store the other datamodel validations like schema validation.
- Added a component validation on the subform component that checks if there are any validation errors on each of its data elements. A special validation type was added for this purpose `SubformValidation`, which includes information about which dataElementIds contains errors, this is used to highlight the rows where there are errors.
- When clicking the `Edit` button on a subform row which is highlighted for containing errors we add a query parameter `?validate=true` when navigating to the subform. This causes the visibility of errors inside the subform to be set to `All` after loading.
- When exiting a subform, the initial validation (`/validate`) query is invalidated so that we fetch validations again so we can get an updated state in the main form. Maybe it could be possible to somehow copy the current state over to the main form, but that seemed complicated.
- I also call `/validate` again when determining whether navigation should be blocked or not if the page contains a subform component, to avoid potentially stale data from blocking navigation. This may be unecessary, and would only really apply if you edit it in another tab, which will already cause problems with patch preconditions in the main form, so this may not be needed?

There are also some changes with routing and using url parameters. These should reduce re-rendering a bit, but they also exposed some issues with the `AppRoutingContext` which were difficult to get around. So there are some changes to how we determine if we need to navigate to the "first" page whatever that is currently also because of this.

## Related Issue(s)

- closes https://github.com/Altinn/app-frontend-react/issues/2148

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels


danielskovli (approved):
All of this is perhaps a bit over my head, but looks good. Lets merge 🚀

src/features/validation/callbacks/onFormSubmitValidation.ts
```diff
@@ -53,14 +58,19 @@ export function useOnFormSubmitValidation() {
...
+        return false;
+      }
 
-    await validating();
-    return callback();
-  }, [callback, validation]);
+      await validating();
+      // TODO(Subform): Figure out a better way to wait for validations to have propagated to node data
+      includeSubform && (await waitForAnimationFrames(10)); // Ugh
+      return callback(includeSubform);
```
olemartinorg, line 72: :cry:

I've been having similar issues lately, so I sympathize.

olemartinorg (approved):
:rocket:

PR #2511, opened 2024-10-01
Stale options bugfix (and more performance improvements)

### LINKED ISSUE #2429: Multiple select options in a repeating group index disappear when removing an previously added repeating group index in v.4.8
### Description of the bug

Related bug issue (two different use cases): https://github.com/Altinn/app-frontend-react/issues/2421

**Before:**
image
**What the repeating group looks like:** 
image
**After deleting class number 20:** 
Screenshot 1 (outside)
image
Screenshot 2 (inside)
image

### Steps To Reproduce

1. Add a repeating group that have multiple Select options
2. Add some repeating group indexes
3. Delete one of the first repeating group indexes that you have previously added
4. Compare v.4.7 with version 4.8 and above

### Additional Information

_No response_

### LINKED ISSUE #2484: Form becomes slow when having multiple elements in big repeating group, and dynamic code lists fail to load
### Description of the bug

When filling out a big repeating group, which can contain up to two inner repeating groups (separate), that has a lot of dynamically fetched code lists, the form starts to get noticably slower, and soon enough the dynamic code lists in new Group-elements will get stuck in an eternal "loading spin" animation. This is solved by refreshing the page, but it's not ideal and can be confusing for a user. 

I checked the network tab in Chrome and it doesn't seem to be as simple as the code lists constantly loading in again and again, as the call to each seems to only happen once for each page no matter how many people you add. 

The source code is in this app: https://altinn.studio/repos/brg/rrh-innrapportering 
URL in TT02: https://brg.apps.tt02.altinn.no/brg/rrh-innrapportering/

(This app is somewhat difficult to get running locally, but if necessary I can be contacted and I will tell you how to do it)

### Steps To Reproduce

1. Open the app. 
2. Login with foedselsnummer: 01812649198
3. Select organization ALLSLAGS ORIENTAL TIGER AS (311780352)
4. Select that the organization "Har reelle rettighetshavere". 
5. On the next page, add a lot of Reelle rettighetshavere (can be done either by foedselsnummer/last name combination, or by selecting "No" on the first question and manually filling out date of birth, name, and similar). Always choose the lowest interval on the question on how much control so you can add a lot of them. 
6. Observe that after a while (sometimes as low as 3, sometimes you need 10) that a lot of the choices (such as intervals) start spinning and never finishing. 

### Additional Information

_No response_

olemartinorg:
## Description

The core of the referenced issue was way too difficult to find, but everything comes down to the fact that the `mapping` used to fetch options utilized the _debounced_ data model, while the node generator (and the repeating group itself for that matter) uses the _current_ data model. So when deleting a row near the top in a repeating group it would take a while for the debouncing to happen, and for that row to be removed from the debounced data as well. I struggled with getting the 'remove stale options' effect to wait to do its job until that time, but when I figured out the root cause of the problem the fix is relatively simple - deleting (and adding) rows now performs the same action on both models (the current and debounced), which also causes this data change to be immediately saved to the backend.

While debugging this, I found and fixed multiple performance-related oddities and unnecessary re-renders, or other jankiness. Notable changes, in chronological order from the beginning:

- I renamed all the `PerformWork` components to something unique. When you use the react profiler these would otherwise be hard to tell apart.
- I moved the options-related effects (removing stale data, setting preselectedOptionIndex, saving downstream parameters and storing the label/displayValue) into separate components that will generally not even be rendered unless options-components are ready for it. I'm hoping this makes this functionality easier to read and reason about.
- I added a key to all `GeneratorRunProvider` components inside the `NodeGenerator`. Not because it's strictly needed, but because it makes it much easier to debug and read the tree in the react devtools.
- When rows are removed from a repeating group, the nodes context is marked as 'not ready' a bit earlier, and the logging (which can be set up in `debug.ts`) now shows you more info about why the state was marked as not ready.
- All the components in the node hierarchy that sets data in a commit-queue doesn't really need to do that inside a `useEffect()`, because adding something to the commit-queue does not set state directly - it just schedules state to be set later. I rewrote all of these to add commits during rendering instead, and using a bit more logic to check if data should be set or not. When doing this I discovered that some of that state-setting could overwrite existing state unintentionally, such as overwriting `updating: true` in attachments or clearing visibility for attachment validations. With the later fixes to this logic, it's now much clearer when we overwrite state and what we keep.
- I removed some complexity that we don't need in the hierarchy generator, such as having its own wrapped and tracked variant of `useEffect` and `useConditionalEffect`. From now on, conditional rendering should work as the only mechanism we need to delay these state changes until parents, etc, has been added.
- The GeneratorStages had its own store, but didn't really need to. I moved the store into NodesContext to avoid having to pass things between them (such as the `markAsReady` function).
- Some parts of the earlier GeneratorStages behaved like refs (were mutated) instead of the more common immutable object pattern used in zustand stores. Instead of putting these in zustand, I moved them to a separate `useRef()` bootstrapped in `GeneratorContext`.
- A bug that was discovered in the `Subform` branch was fixed there by moving `commitTimeout` (in `GeneratorStages.tsx`) to a ref instead of a global variable. I changed this here as well (which will cause a merge conflict), by moving that ref to one that is global for the entire `NodesContext`, not local to the hook.
- All the commit-queue stuff was moved into its own file (`CommitQueue.tsx`) and out from `GeneratorStages.tsx` (this caused quite a bit of the line-count, I would think)
- Avoiding re-rendering in GenericComponent by only fetching what is needed there. It would routinely re-render a bit much every time something changed in (for example) a RepeatingGroup item.
- Using `requestIdleCallback` when available to mark the NodesContext as ready, instead of an over-active `setInterval`. This reduces the load required to check if we can mark the state as ready.
- In a RepeatingGroup we'd set the row UUID in the state for each row, but when deleting an earlier row that state could be stale/outdated for a while. I opted to remove the value instead (set it to undefined) when rows are removed so we'd notice sooner that the UUID was not the correct one. This is then updated later by `MaintainRowUuid` so that we can know when the `item` state is up to date with the data model.
- In `useNodeItem()` I used to just return the latest state when components re-ran their selectors in the periods where the NodesContext is not ready. In this change I added a ref with the previous value and let the selector run on that while the store is not ready - that way lots of re-renders for visible components are delayed until state has stabilized. I also added a check to make sure useNodeItem() is not called inside the generator itself, which we've seen before could cause unexpected results (this hook simplifies the world for you and lets you think the item is always set, but in the generator it isn't).
-  Rewriting `deepValidationsForNode` to not use node traversal, as it would cause unneccessary re-renders when deleting a repeating group row.
- Every component (and plugin) can now check if their state is ready, and the readiness-checker is guaranteed not to mark the state as ready until that's done.
- I noticed that `string[]` in the data model could cause patch changes even when the prev/next values were essentially the same. Comparing with `deepEqual` fixed that.
- When uploading attachments to repeating groups, I saw gigantic re-renders because data in `InstanceContext` was altered. As a quick stop-gap solution, I rewrote it to use a zustand store instead.
- Same goes with `ProcessContext.tsx`. Now I only see large re-renders when uploading attachments because of changes in `ProcessWrapper`, but I suspect that will be fixed in upcoming changes in `Subform`.
- Asking the backend to include row UUIDs when performing a regular single-data-model PATCH request. I supect this was left out by mistake when adding support for multiple data models.
- In repeating groups, when earlier rows are deleted (or all rows sorted), the UUIDs in the store would not match the data model for a while, and when that happened the edit/delete buttons would fail to work. Fixing this by disabling the buttons while the row is not 'fresh' according to the data model.

Test changes:
- Rewrote some tests that used fake timers, but broke (presumably because of `requestIdleCallback`).
- Various minor updates to cypress tests, typically to make them wait until things show up before performing the next action
- I implemented a mechanism to simulate form data saving in unit tests, along with faking `ProcessDataWrite` (untested as of now). One test failed because adding/removing rows are now debounced immediately, and then it would get stuck waiting for saving - it would previously finish before saving occurred.
- At least one test failed because `useDeviceWidths()` did not catch a resize after rendering, before useEffect() was called

## Related Issue(s)

- closes #2429
- closes #2484

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels


olemartinorg:
Percy snapshot shows some issues with `Map`, along with a disappearing nested group in summary. I'll look into it.

src/test/renderWithProviders.tsx
```diff
@@ -387,13 +393,52 @@ export function setupFakeApp({ queries, mutations }: SetupFakeAppProps = {}) {
...
+    });
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  (mutationMocks as any).doPostStatelessFormData = jest.fn().mockImplementation(async (url: string, data: unknown) => {
+    const afterProcessing = typeof mockBackend === 'function' ? mockBackend(data, url) : data;
+    models[url] = afterProcessing;
+    return afterProcessing;
+  });
+}
+
```
bjosttveit, line 433: This looks very useful! Is there a reason multi-patch does not have a mock implementation? I guess we dont have any unit tests with multiple data models yet

test/e2e/integration/frontend-test/attachments-in-group.ts
```diff
@@ -24,7 +24,7 @@ describe('Repeating group attachments', () => {
   // Allows you to toggle reload() tests off. These have the effects of testing mapAttachments() to make sure state is
   // the same even after reloading the instance data, but you can toggle them off to save some time when developing
   // the test.
-  const runReloadTests = true;
+  const runReloadTests = false;
```
bjosttveit, line 27: Should this still be off?

bjosttveit (approved):

src/test/renderWithProviders.tsx
olemartinorg: Yep, that's about it. Shouldn't be any issues adding support for it, but as it stands I didn't bother looking into the nitty gritty.. :face_with_peeking_eye:

test/e2e/integration/frontend-test/attachments-in-group.ts
olemartinorg: Whoops, thanks! :pray:

PR #2550, opened 2024-10-09
Purge azure front door cache on deploy

bjosttveit:
## Description

Purge Azure Front Door cache on release.
- Purging `index.json` should make pre-releases show up in localtest immediately
- For actual releases also purging `/major/` and `/major.minor/` so that e.g. `/4/` or `/4.8/` points to the new version immediately


tjololo (approved):
Looks good to me

PR #2567, opened 2024-10-10
MUI Grid Cleanup

cammiida:
## Description

Remove use of MUI Grid component and replace it with our own Flex component, as a part of the way to get rid of the dependency to material-ui.

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels


paal2707:
Har du tatt i bruk designsystemet sine tokens her? 
https://www.designsystemet.no/grunnleggende/designelementer/storrelser-og-avstander

src/app-components/Flex/Flex.module.css
```diff
@@ -0,0 +1,342 @@
...
+  flex-basis: 91.66%;
+}
+.col-xs-12 {
+  flex-grow: 0;
+  max-width: 100%;
+  flex-basis: 100%;
+}
+
+/* Small (sm breakpoint) */
+@media (min-width: 600px) {
```
olemartinorg, line 153: These breakpoints don't match what we've been using before (see the exported `breakpoints` from `useDeviceWidths.ts`). If you took the default values from MUI, remember that we're overriding the theme (see `altinnApTheme.tsx`, which also don't match the breakpoints in `useDeviceWidths.ts` :confused:)

src/app-components/Flex/Flex.tsx
```diff
@@ -0,0 +1,90 @@
...
+          display: 'block',
+          ...style,
+        };
+
+    return (
+      <Tag
+        id={id}
+        ref={ref}
+        className={cn(
+          'altinn-flex',
```
olemartinorg, line 71: If this is for testing, I would prefer using a `data-testid` property instead

src/app-components/Number/DisplayNumber.tsx
```diff
@@ -30,7 +37,12 @@ export const Number = ({ value, formatting, iconUrl, iconAltText, labelId, curre
           alt={iconAltText}
         />
       )}
-      <span aria-labelledby={labelId}>{displayData}</span>
+      <span
+        className='number'
```
olemartinorg, line 41: Same thing here as for `Date`

src/app-components/Text/DisplayText.tsx
```diff
@@ -18,6 +18,11 @@ export const Text = ({ value, iconUrl, iconAltText, labelId }: TextProps) => (
         alt={iconAltText}
       />
     )}
-    <span aria-labelledby={labelId}>{value}</span>
+    <span
+      className='text'
```
olemartinorg, line 22: Same thing here as for `Date`

src/components/presentation/Presentation.tsx
```diff
@@ -36,6 +35,10 @@ export const PresentationComponent = ({ header, type, children, renderNavBar = t
   const userParty = useProfile()?.party;
   const { expandedWidth } = useUiConfigContext();
 
+  const { showProgress } = usePageSettings();
+  const isProgressEnabled = type !== ProcessTaskType.Archived && showProgress;
+  const showProgressBar = isProgressEnabled && showProgress;
```
olemartinorg, line 40: I faintly remember there being a reason for moving these out to their own component (quite possibly re-renders caused by the page settings update causing weird bugs). If there's no good reason for moving it inline into the very centrally important `Presentation` component, I would prefer to keep the separation of concerns that was.

src/features/validation/ComponentValidations.tsx
```diff
@@ -50,42 +50,45 @@ export function ComponentValidations({ validations, node: _node }: Props) {
   const info = validationsOfSeverity(filteredValidations, 'info');
   const success = validationsOfSeverity(filteredValidations, 'success');
 
+  if (!node || !validations?.length) {
+    // eslint-disable-next-line react/jsx-no-useless-fragment
+    return <></>;
```
olemartinorg, line 55: Why is the useless fragment needed? :sweat_smile:
```suggestion
    return null;
```

src/features/validation/ComponentValidations.tsx
```diff
@@ -50,42 +50,45 @@ export function ComponentValidations({ validations, node: _node }: Props) {
   const info = validationsOfSeverity(filteredValidations, 'info');
   const success = validationsOfSeverity(filteredValidations, 'success');
 
+  if (!node || !validations?.length) {
```
olemartinorg, line 53:
```suggestion
  if (!node || !filteredValidations?.length) {
```

src/layout/Accordion/Accordion.module.css
```diff
@@ -7,3 +7,7 @@
     break-inside: avoid;
   }
 }
+
+.fullWidth {
+  width: 100%;
+}
```
olemartinorg, line 13: Was this ever used? :thinking:

src/layout/Group/GroupSummary.tsx
```diff
@@ -90,16 +90,18 @@ export const GroupSummary = ({ componentNode, hierarchyLevel = 0, summaryOverrid
...
-        size={isNestedGroup ? 'xsmall' : 'small'}
-        level={headingLevel}
-      >
-        <Lang
-          id={summaryTitle ?? title}
-          node={componentNode}
-        />
-      </Heading>
-      <Grid
+      {(summaryTitle || title) && (
```
olemartinorg, line 93: Nice! :raised_hands: I just found the same problem when working on the `Option` component. :partying_face:

src/layout/Summary/SummaryComponent.tsx
```diff
@@ -100,21 +100,23 @@ export const SummaryComponent = React.forwardRef(function SummaryComponent(
...
-      lg={displayGrid?.lg || false}
-      xl={displayGrid?.xl || false}
+      item
+      size={{
+        xs: displayGrid?.xs ?? 12,
+        sm: displayGrid?.sm,
+        md: displayGrid?.md,
+        lg: displayGrid?.lg,
+        xl: displayGrid?.xl,
+      }}
```
olemartinorg, line 112: Doesn't these do the same thing? :thinking: 
```suggestion
      size={displayGrid}
```

olemartinorg (approved):
Solid piece of work! :partying_face: Great job! :clap: :rocket: The only thing I'm hesitant about is the breakpoints (see relevant comment), but otherwise this looks very good.

src/app-components/Flex/Flex.module.css
paal2707: I had the correct breakpoints at first, but i got a diff from percy, and i changed it all back. I will look deeper into it! :)

src/layout/Summary/SummaryComponent.tsx
paal2707: No, the difference is that the fallback for xs on 12 is not applied.

src/app-components/Date/DisplayDate.tsx
```diff
@@ -30,7 +30,12 @@ export const Date = ({ value, format, iconUrl, iconAltText, labelId }: DateProps
           alt={iconAltText}
         />
       )}
-      <span aria-labelledby={labelId}>{displayData}</span>
+      <span
+        className='date'
```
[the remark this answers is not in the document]
paal2707: This is something i tested with. Forgot to remove.

src/layout/Summary/SummaryComponent.tsx
olemartinorg: So if you don't specify a size for `xs`, what does it do when you shrink it down? :thinking: I would imagine 12 is a good default.

paal2707: If you dont specify size for xs you dont get fluid sizing. This is the default behavior of MUI. it would still be 12 as a default but not fluid sizing. We could rewrite this to be default fluid 12, but it would require a rewrite css of many of our components. The priority in this PR is just to replicate what we already had. Many possibilities to improve our grid system now that we own it ourselves.

src/app-components/Flex/Flex.module.css
paal2707: I see now that we need the current breakpoints as it is, because this corresponds with the MUI Grid component. They are equal to the overriding theme values(`altinnApTheme.tsx` ). I think that we should consider the changing the values of the spacing/breakpoints/etc to correspond with the designsystemet as a next iteration on the Flex component. Now the focus is to get rid of MUI without breaking changes.

src/layout/Summary/SummaryComponent.tsx
olemartinorg: I see! :raised_hands:

src/app-components/Flex/Flex.module.css
olemartinorg: Yep - I suggest creating an issue for this with `quality/debt` and `kind/breaking-change` that we can tackle when we're planning v5. We should consolidate all the breakpoints.

Issue #1850, opened 2024-02-21
List component accessibility issues


Febakke:
### Description of the bug

There are multiple issues with our list component
1. Users cant navigate the radiobuttons using the arrow-keys
2. No focus visible on radio buttons
3. Radiobuttons are not grouped. This will confuse users as screen readers will say radiobutton 1 of 1.
4. "rader per side" is not connected to the label. 

- [ ] Consider recreating this with designsystem components.

### Steps To Reproduce

Go to an app with a list and run accessiblity tests

Febakke:
- [ ] Add screenshots showing of the bugs @Febakke 
- [ ] Consider fixing excisting List or build new using designsystem components

olavflar:
Related to; #2373

Magnusrm:
@adamhaeger I tested issue #6 after it was updated to DS, and the issue is still there. I think we should retest all the issues and cross off ones that are fixed, and fix the ones that are still there.

PR #2585, opened 2024-10-16
Fix bugs in List component

bjosttveit:
## Description

- Removes dataModelBinding validation that asserts that you can only use bindings from table headers. This check was broken when initially added, but fixed in #2108. It turns out that several apps in prod used dataModelBindings for mapping columns that were not shown in tableHeaders, so fixing this dataModelBinding validation was a breaking change.
- Removing `filteredRows` from `ListComponent`. This would filter out values that should not be shown in mobile view if `tableHeadersMobile` was set. The only problem is that the filteredRows object was also used for setting formData, so if you scaled down your browser window, some fields may no longer get saved when you click the row. Added unit tests for this.
- Table was using the raw data-list row when mapping cells in the table body. This does not preserve the column ordering specified in table-headers, and if not all columns in the data-list were used in table headers it would add more columns to the table body than the header has (see image below). Now using tableHeaders as the source of truth for what goes where.

image

## Related Issue(s)

- https://digdir.slack.com/archives/C0760NPT2BE/p1729059789947969

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels

### DIFF
--- a/src/layout/List/ListComponent.test.tsx
+++ b/src/layout/List/ListComponent.test.tsx
@@ -205,6 +205,9 @@ describe('ListComponent', () => {
     const user = userEvent.setup({ delay: null });
     const { formDataMethods } = await render({ component: { tableHeadersMobile: ['Name', 'FlagLink'] } });
 
+    // Make sure test is not broken by changing mobile-view implementation
+    expect(useDeviceWidths.useIsMobile).toHaveBeenCalled();
+
     // There should be one radio for each country, but none of them should be checked
     await waitFor(() => expect(screen.getAllByRole('radio')).toHaveLength(6));
     expect(screen.queryByRole('radio', { checked: true })).not.toBeInTheDocument();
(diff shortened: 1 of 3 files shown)


Magnusrm:
Code looks good to me! 👍 
Found a couple of issues through accessability testing, though they were probably there earlier as well.

One is that the selected value of the list is read as the stringified json object of the row, which is not that user friendly:
image

The other is just an empty header reported by WAVE:
image

This could probably be solved by using a visually hidden text that describes the action; "Velg/Select" or something.

bjosttveit:
> though they were probably there earlier as well.

Indeed, pretty sure its been like this since day one of the `List`-component

Magnusrm:
> Does this not have a wcag issue?

I only found this: #1850
But it does not mention the issues i found.

We could update that issue with the ones i found i guess 😊

cammiida (approved):
LGTM! Bra catcha 🚀

Issue #2571, opened 2024-10-11
New Table component


adamhaeger:
We have seen the need for a simple table component that displays data in a table.

This component should be used as the table view for other components that needs to display a table.

Acceptance criterea:

- [ ] A table component that has no dependencies to rest of the app.
- [ ] Receives data and column configuration.
- [ ] Is WCAG friendly.

Link to [design](https://www.figma.com/design/d4767trHl7AcKN4XjgDdEX/Table?node-id=0-1)

PR #2593, opened 2024-10-17
Feature/table

adamhaeger:
## Description

- Adds Table component
- Introduces component libary
- Includes ADR describing the propsed change

## Related Issue(s)

- closes [#{issue number}](https://github.com/Altinn/app-frontend-react/issues/2571)

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels


adr/001-component-library.md
```diff
@@ -0,0 +1,60 @@
...
+- Allows us to really quickly get started with a component library
+- Bad, because it does not fulfill B3. If we simply use a new folder, it will be up to developers to enforce the rules of the UI components, like the avoiding dependencies to the app.
+
+### A2
+
+- Good, because this alternative adheres to B1, B2 and B3.
+- This way our libs would live separately to the app, and it would be obvious that it is a lib.
+- The con is that it takes more setup.
+
+### A2
```
cammiida, line 55:
```suggestion
### A3
```

cammiida (commented):
Dette blir bra! 🚀

adr/001-component-library.md
```diff
@@ -0,0 +1,60 @@
...
+
+Today our UI components are tightly coupled to the app in which they are rendered.
+
+This leads to several issues:
+
+- Makes it hard to do component testing outside a fully rendered app.
+- Makes refactoring the app framework a lot harder.
+- Leads to unclear interfaces between UI components and the app framework.
+- Makes developing UI components complex without deep understanding of the application.
+
```
martinothamar, line 21:
```suggestion
- Enables sharing of pure components (docs, Studio, ???)
```
?

HauklandJ:
As ADRs are introduced here, it would be convenient to add the template.md file to the adr folder :raised_hands:

adr/001-component-library.md
```diff
@@ -0,0 +1,60 @@
...
+## Problem context
+
+Today our UI components are tightly coupled to the app in which they are rendered.
+
+This leads to several issues:
+
+- Makes it hard to do component testing outside a fully rendered app.
+- Makes refactoring the app framework a lot harder.
+- Leads to unclear interfaces between UI components and the app framework.
```
martinothamar, line 19: Would be nice to have linked examples of code/components for some of these points, I'm sure you've discussed and looked at tons of examples here to inform this assertion but for myself presently and future readers it would be helpful to know the context here 😄

adr/001-component-library.md
```diff
@@ -0,0 +1,60 @@
...
+- Makes developing UI components complex without deep understanding of the application.
+
+## Decision drivers
+
+A list of decision drivers. These are points which can differ in importance. If a point is "nice to have" rather than
+"need to have", then prefix the description.
+
+- B1: UI components should only receive data to display, notify the app when data is changed, and notify validity of user input.
```
bjosttveit, line 27:
```suggestion
- B1: UI components should only receive data to display, and notify the app when data is changed.
```
I suggest that we remove the validation of user input part, as it makes little sense to leave this to UI components in our context

adr/001-component-library.md
adamhaeger: Agree, done!

src/layout/Table/TableComponent.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+  const data = formData.simpleBinding as IDataModelReference[];
+  if (!data) {
+    return null;
+  }
+
+  if (data.length < 1) {
+    return null;
+  }
+  return (
```
paal2707, line 38: linebreak here?

src/layout/Table/TableComponent.tsx
```diff
@@ -0,0 +1,109 @@
...
+  if (item.enableDelete) {
+    actionButtons.push({
+      onClick: (idx) => {
+        removeFromList({
+          startAtIndex: idx,
+          reference: {
+            dataType: item.dataModelBindings.simpleBinding.dataType,
+            field: item.dataModelBindings.simpleBinding.field,
+          },
+          callback: (_) => true,
```
paal2707, line 82: What does this do? `(_) => true` ? What is the reason for having the underscore?

paal2707 (approved):
This looks really good. Looking forward to having this on the rest of the component as well. Approved with some questions/comments.

src/layout/Table/TableComponent.tsx
adamhaeger: In typescript the _ inidcates that the variable is unused so it doesnt give you a unused variable error :)

src/app-components/table/Table.tsx
```diff
@@ -0,0 +1,189 @@
...
+  actionButtons?: TableActionButton[];
+  /** Accessible header value for action buttons for screenreaders */
+  actionButtonHeader?: React.ReactNode;
+  /** Displays table in mobile mode */
+  mobile?: boolean;
+  size?: 'sm' | 'md' | 'lg';
+  zebra?: boolean;
+}
+
+function formatIfDate(value: unknown): string {
```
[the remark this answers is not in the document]
adamhaeger: Absolutely :) Will update when Date is moved to component lib :)

src/features/formData/FormDataWriteStateMachine.tsx
```diff
@@ -399,13 +399,17 @@ function makeActions(
           window.logError(`Tried to write to readOnly dataType "${reference.dataType}"`);
           return;
         }
+
+        console.log('statedataModels', state.dataModels);
+
```
Magnusrm, line 404:
```suggestion
```

cammiida (commented):
Vil ikke approve pga ekstern datamodell, men ser meget bra ut 💯

src/layout/Table/TableComponent.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+  const data = formData.simpleBinding as IDataModelReference[];
+
+  if (!data) {
+    return null;
+  }
+
+  if (data.length < 1) {
+    return null;
+  }
```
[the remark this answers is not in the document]
adamhaeger: Skulle gjerne, men det går ikke: Operator < cannot be applied to types boolean and number

src/layout/Table/config.ts
```diff
@@ -0,0 +1,62 @@
...
+    renderInTabs: true,
+  },
+  functionality: {
+    customExpressions: false,
+  },
+})
+  .extends(CG.common('LabeledComponentProps'))
+  .extendTextResources(CG.common('TRBLabel'))
+  .addProperty(new CG.prop('title', new CG.str()))
+  .addDataModelBinding(CG.common('IDataModelBindingsSimple'))
```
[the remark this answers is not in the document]
adamhaeger: Vil merge den først siden da kan DIBK ta den i bruk, hva er galt med å legge til støtte for det senere?

src/layout/Table/TableComponent.tsx
```diff
@@ -0,0 +1,115 @@
...
+  const item = useNodeItem(node);
+  const { formData } = useDataModelBindings(item.dataModelBindings, 1, 'raw');
+  const removeFromList = FD.useRemoveFromListCallback();
+  const { title, description, help } = item.textResourceBindings ?? {};
+  const { langAsString } = useLanguage();
+  const { elementAsString } = useLanguage();
+  const accessibleTitle = elementAsString(title);
+  const isMobile = useIsMobile();
+
+  const data = formData.simpleBinding as IDataModelReference[];
```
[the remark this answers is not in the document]
adamhaeger: Done :)

adr/001-component-library.md
```diff
@@ -0,0 +1,65 @@
...
+- Enables sharing of pure components (docs, Studio, storybook)
+
+## Decision drivers
+
+A list of decision drivers. These are points which can differ in importance. If a point is "nice to have" rather than
+"need to have", then prefix the description.
+
+- B1: UI components should only receive data to display, and notify the app when data is changed.
+- B2: UI components should live in a separate folder from the app itself, and have no dependencies to the app.
+- B3: UI components should live in a lib separately to the src folder to have stricter control of dependencies.
```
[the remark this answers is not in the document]
olemartinorg: Ah, maybe it's my misunderstanding of ADRs. This is just documenting everything you talked about, even though the final solution fell short of supporting B3? Still, I didn't immediately see the connection to the problem context.

src/app-components/readme.md
```diff
@@ -0,0 +1,11 @@
...
+
+Components in this folder should be 'dumb', meaning the should have no knowledge about the containing app.
+
+They should implement three things:
+
+1. Receive and display data in the form of props.
+2. Report data changes in a callback.
+3. Receive and display the validity of the data.
+
+These components will form the basis of our future component library, and will enable refactoring of the frontend app.
```
olemartinorg, line 11: Not sure that's how it's meant to be used, but might as well link to the ADR from here as well. :pray:

src/app-components/table/Table.tsx
```diff
@@ -0,0 +1,161 @@
...
+                  data-header-title={col.header}
+                >
+                  <ul>
+                    {cellValues.map(formatIfDate).map((value, idx) => (
+                      <li key={idx}>{value}</li>
+                    ))}
+                  </ul>
+                </Table.Cell>
+              );
+            })}
```
olemartinorg, line 136: Might be a matter of taste, but I would prefer to extract large chunks like this into one or more separate components.

src/layout/Table/TableComponent.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+  if (!Array.isArray(data)) {
+    return null;
+  }
+
+  if (data.length < 1) {
+    return null;
+  }
+  return (
+    <AppTable<IDataModelReference>
```
olemartinorg, line 41: This is surely not displaying a table for a `IDataModelReference`? That's config, and not meant to be exposed to the user in any way. Since you `useDataModelBindings(..., 'raw')` above, the data is `FDValue` (or whatever that type is called), which is practically almost the same as `any`.

Another way of putting it; maybe this doesn't need to use generics at all, or at least provide a default value when the generic value is not used for type-safety at all (like is the case here).

src/layout/Table/TableComponent.tsx
```diff
@@ -0,0 +1,116 @@
...
+import type { LayoutNode } from 'src/utils/layout/LayoutNode';
+
+type TableComponentProps = PropsFromGenericComponent<'Table'>;
+
+type TableSummaryProps = {
+  componentNode: LayoutNode<'Table'>;
+};
+
+export function TableSummary({ componentNode }: TableSummaryProps) {
+  const item = useNodeItem(componentNode);
```
olemartinorg, line 25: You're only using `item.dataModelBindings`, `item.textResourceBindings` and `item.columnConfig` here, so it might be more performant to just select those individually.

src/layout/Table/TableComponent.tsx
```diff
@@ -0,0 +1,116 @@
...
+            title={<Lang id={title} />}
+            description={description && <Lang id={description} />}
+            helpText={help ? { text: <Lang id={help} />, accessibleTitle } : undefined}
+          />
+        )
+      }
+      data={data}
+      columns={item.columnConfig.map((config) => ({
+        ...config,
+        header: langAsString(config.header),
```
olemartinorg, line 109: Does it have to be a string, or could you pass `<Lang id={config.header} />` here instead?

src/layout/Table/TableComponent.tsx
```diff
@@ -0,0 +1,116 @@
...
+        removeFromList({
+          startAtIndex: idx,
+          reference: {
+            dataType: item.dataModelBindings.simpleBinding.dataType,
+            field: item.dataModelBindings.simpleBinding.field,
+          },
+          callback: (_) => true,
+        });
+      },
+      buttonText: langAsString('general.delete'),
```
olemartinorg, line 87: This should probably also accept a `<Lang id='general.delete' />`.

src/layout/Table/config.ts
olemartinorg, line 22: This should absolutely be a `group` ~~or `list`~~ binding, not a `simpleBinding` (the latter is used for strings, numbers, simple values - not groups/arrays).

EDIT: Not `list` either, when I thought more about it. In short, we've historically used:
- `simpleBinding` for simple leaf-values like strings, numbers, dates, booleans. This name is just a historical artifact and should be phased out IMO (we should use keys like `value` or `text` in components like `Input`, not `simpleBinding`). 
- `list` points to a an array containing simple values (i.e. `string[]`, `number[]`, `boolean[]`)
- `group` points to an array of objects (`object[]`)

src/layout/Table/config.ts
```diff
@@ -0,0 +1,62 @@
...
+      new CG.bool().setTitle('Enable delete').setDescription('If true, will allow user to delete row').optional(),
+    ),
+  )
+  .addProperty(
+    new CG.prop('size', new CG.enum('sm', 'md', 'lg').setTitle('Size').setDescription('Size of table.').optional()),
+  )
+  .addProperty(
+    new CG.prop(
+      'excludedChildren',
+      new CG.arr(new CG.enum('action', 'md', 'lg'))
```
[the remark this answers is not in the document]
olemartinorg: Also, if these are only excluded from the summary, maybe the prop name should indicate that? I.e. `excludedFromSummary`?

src/layout/Table/typeguards.ts
```diff
@@ -0,0 +1,16 @@
...
+    'dataType' in obj &&
+    'field' in obj &&
+    typeof (obj as { dataType: unknown }).dataType === 'string' &&
+    typeof (obj as { field: unknown }).field === 'string'
+  );
+}
+
+export function isIDataModelReferenceArray(obj: unknown): obj is IDataModelReference[] {
+  return Array.isArray(obj) && obj.every(isIDataModelReference);
+}
```
olemartinorg, line 16: What's this for? Arrays of data model references would be an entirely new concept unknown to me at least.

olemartinorg (commented):
I'm a bit skeptical of naming this component `Table`. While it currently has support for repeating structures, it is not clear how we should expand this configuration to include support for non-repeating structures (like the current `Grid` and `List` components). If we don't land on something now, we risk having to integrating these concepts with a sub-optimal configuration later on, because the `Table` component name has already been reserved at that point. I noticed that [my config suggestion](https://digdir.slack.com/archives/C076GU2QHCG/p1726064345921509) received few comments last round, and I don't see any traces of inspiration from there in this PR.

src/layout/Table/config.ts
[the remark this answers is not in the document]
adamhaeger: good catch, this should certainly not be here, just a copy paste error :)

adr/001-component-library.md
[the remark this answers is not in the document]
adamhaeger: So, the ADR conculdes that we should not introduce yarn workspaces or NX at this point, that would be separate ADR if we decied to investigate further. So that would as you say be pulled into a separate ADR.

But youre right, it could make sense that this ADR asked "should we introduce a component library?". However this was really already agreed upon in frontend architecture forum, so this ADR basically formulates _how_ what was already agreed upon should be implemented.

src/app-components/readme.md
adamhaeger: This is more a general guideline for the components in the folder. My preference would be to have this lib completely separate from the src folder so that importing the wrong way is not possible, but as we have not decided to do this yet, this readme serves as a litt reminder :)

src/layout/Table/TableComponent.tsx
adamhaeger: Yeah I think a key valye type of some sort is better :)

src/layout/Table/config.ts
adamhaeger: Ill have a look at group binding :) Didnt know simpleBinding was going away, but that makes sense :)

src/layout/Table/typeguards.ts
```diff
@@ -0,0 +1,16 @@
...
+export function isIDataModelReference(obj: unknown): obj is IDataModelReference {
+  return (
+    typeof obj === 'object' &&
+    obj !== null &&
+    'dataType' in obj &&
+    'field' in obj &&
+    typeof (obj as { dataType: unknown }).dataType === 'string' &&
+    typeof (obj as { field: unknown }).field === 'string'
+  );
+}
```
bjosttveit, line 12: There is already a typeguard for this in `utils/databindings.ts`: `isDataModelReference()`

src/layout/Table/config.ts
olemartinorg: It's probably not going away any time soon.. :cry: But I consider it left over from times past, back when it was difficult to introduce new data model binding keys, so it seemed natural to just re-use the same keys for everything. Now it's very easy to introduce new bindings, so there's no need to use `simpleBinding` for everything anymore.

adamhaeger:
> I'm a bit skeptical of naming this component `Table`. While it currently has support for repeating structures, it is not clear how we should expand this configuration to include support for non-repeating structures (like the current `Grid` and `List` components). If we don't land on something now, we risk having to integrating these concepts with a sub-optimal configuration later on, because the `Table` component name has already been reserved at that point. I noticed that [my config suggestion](https://digdir.slack.com/archives/C076GU2QHCG/p1726064345921509) received few comments last round, and I don't see any traces of inspiration from there in this PR.

I had not seen your comment there, Ill have a look :)

src/layout/Table/TableComponent.tsx
olemartinorg: Yep - but make sure the component also handles nested arrays, objects, null, etc without failing. :raised_hands:

src/layout/Table/config.ts
adamhaeger: Is the "group" name special or could it be anything like `tableData`?

src/layout/Table/typeguards.ts
adamhaeger: Removed :)

src/layout/Table/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { TableDef } from 'src/layout/Table/config.def.generated';
+import { TableComponent, TableSummary } from 'src/layout/Table/TableComponent';
+import type { LayoutValidationCtx } from 'src/features/devtools/layoutValidation/types';
+import type { PropsFromGenericComponent } from 'src/layout';
+import type { SummaryRendererProps } from 'src/layout/LayoutComponent';
+import type { Summary2Props } from 'src/layout/Summary2/SummaryComponent2/types';
+
+export class Table extends TableDef {
+  validateDataModelBindings(_: LayoutValidationCtx<'Table'>): string[] {
+    return [];
```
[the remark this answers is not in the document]
adamhaeger: Good one, added :)

src/layout/Table/config.ts
olemartinorg: Not really special, no - `tableData` could work as well. But I always try to go for deduplicating names (so `simpleBinding` -> `simple`, as all keys in the `dataModelBindings` are bindings), so in this case I would just call it `table` (as all all keys in `dataModelBindings` point to data).

adr/001-component-library.md
```diff
@@ -0,0 +1,65 @@
...
+- Makes developing UI components complex without deep understanding of the application.
+- Enables sharing of pure components (docs, Studio, storybook)
+
+## Decision drivers
+
+A list of decision drivers. These are points which can differ in importance. If a point is "nice to have" rather than
+"need to have", then prefix the description.
+
+- B1: UI components should only receive data to display, and notify the app when data is changed.
+- B2: UI components should live in a separate folder from the app itself, and have no dependencies to the app.
```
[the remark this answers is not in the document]
adamhaeger: This is like saying its confusing to use designsystemet as a view library.

The benefits of separating view components from the app logic includes but is not limited to:

**1. Code Reusability:** Pure view components are reusable across different parts of the app, as they are only concerned with how the UI looks, not how it behaves. This makes it easy to use the same component in multiple places without duplicating logic.

This is a more important point than reusing components with Studio or Storybook, but it also gives us that possibility.

**2. Simplified Testing:** With view components separated, you can test them independently without worrying about application logic. This isolation makes unit tests more focused and reliable, as you only need to check if a component renders correctly for given inputs (props) and not the entire app flow.

This is a huge issue in our app currently. This will ensure we can both move more testing to components which will be faster now and scale better in the future.

**3. Easier Maintenance:** Separation makes it easier to locate and fix bugs since you know exactly where the UI ends and the logic begins. This modularity also makes refactoring or redesigning the UI simpler, as you’re only changing visual components without touching business logic.

As we are reusing things like lists, tables, inputs etc, we only update in one spot as well.

**4. Improved Readability:** Clean separation between views and logic helps other developers (or your future self) understand the structure of the app more quickly. 

They can identify what each part does at a glance, as there is a clear boundary between the UI layer and the functionality layer.

**5. Scalability:** As applications grow, so does their complexity. A separation between UI and logic allows for better scalability by organizing code into distinct concerns. This allows teams to work on specific parts of the application more independently, making collaboration smoother.

In the future it could make sense to have one team focused on input components, and another focused on the app logic.

**Easier State Management:** When the view components are separate, managing state becomes clearer. Logic components handle state, business rules, and data fetching, while view components remain “dumb” and only render the provided data. This results in a more predictable data flow.

src/app-components/table/Table.tsx
```diff
@@ -0,0 +1,161 @@
...
+              const cellValues = col.accessors
+                .filter((accessor) => !!pick(accessor, rowData))
+                .map((accessor) => pick(accessor, rowData));
+              if (cellValues.every((value) => value == null)) {
+                return (
+                  <Table.Cell
+                    key={colIndex}
+                    data-header-title={col.header}
+                  >
+                    -
```
olemartinorg, line 97: Why do we need to show a hyphen for empty cells? No other table components do that as I'm aware of. If it's a matter of taste, it should probably be configurable.

PR #2594, opened 2024-10-17
Add workflow for quickly rolling back to an earlier release

bjosttveit:
## Description

With big changes coming up in the next release, I thought it would be a good idea to have the ability to quickly roll back to an earlier release in case we introduce bugs that affect a lot of apps/users, without having to create a new release.

This workflow works by manually dispatching from the `Actions`-page where you specify the tag of the version you want to roll back to. The workflow will checkout the `altinn-cdn`-repo where we store our builds, and copy the selected version over into the `major` (e.g. `/4/`) and `major.minor` (e.g. `/4.8/`) folders, sync this to azure, and purge the cache.

This can be used if shit really hits the fan as it should be much faster than fixing the bug and building a new version. When the bug(s) are eventually fixed, releasing a new version should still overwrite this temporary roll back.

### DIFF
--- a/.github/scripts/revert.sh
+++ b/.github/scripts/revert.sh
@@ -57,6 +57,7 @@ while [[ $# -gt 0 ]]; do
   esac
 done
 
+ROOT=$(pwd)
 TARGET=toolkits/altinn-app-frontend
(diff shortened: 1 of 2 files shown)


tjololo (approved):
LGTM, easy to test and verify actual behavior by removing `--commit` and `--azure-sync-cdn` from the workflow execution.

PR #2680, opened 2024-11-04
Refactor Input component to App Components

cammiida:
## Related Issue(s)

- closes #{issue number}

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels


src/app-components/Input/Input.tsx
```diff
@@ -0,0 +1,63 @@
...
+  if (props.textOnly) {
+    const { value, id, className } = props;
+    if (value === null || (typeof value === 'string' && value.length === 0)) {
+      return null;
+    }
+
+    return (
+      <Paragraph
+        id={id}
+        size='small'
```
HauklandJ, line 48: Why do we hardcode size for paragraph?

HauklandJ (approved):
LGTM

src/app-components/Input/Input.tsx
cammiida: Good question! Probs not 😅

src/app-components/Input/Input.module.css
```diff
@@ -0,0 +1,19 @@
+input[type='search'] {
+  background-image: url('data:image/svg+xml,%3Csvg%20xmlns%3D%22http%3A%2F%2Fwww.w3.org%2F2000%2Fsvg%22%20width%3D%221.5em%22%20height%3D%221.5em%22%20fill%3D%22none%22%20viewBox%3D%220%200%2024%2024%22%20focusable%3D%22false%22%20role%3D%22img%22%3E%3Cpath%20fill%3D%22currentColor%22%20fill-rule%3D%22evenodd%22%20d%3D%22M10.5%203.25a7.25%207.25%200%201%200%204.569%2012.88l5.411%205.41a.75.75%200%201%200%201.06-1.06l-5.41-5.411A7.25%207.25%200%200%200%2010.5%203.25M4.75%2010.5a5.75%205.75%200%201%201%2011.5%200%205.75%205.75%200%200%201-11.5%200%22%20clip-rule%3D%22evenodd%22%3E%3C%2Fpath%3E%3C%2Fsvg%3E');
```
olemartinorg, line 2: I would think it's better to import [the icon from aksel](https://aksel.nav.no/ikoner/MagnifyingGlass) instead of having a copy here.

src/app-components/Input/Input.tsx
```diff
@@ -0,0 +1,63 @@
...
+  | 'placeholder'
+  | 'inputMode'
+  | 'style'
+>;
+
+export function Input(props: InputProps) {
+  const { size = 'sm', ...rest } = props;
+
+  if (props.textOnly) {
+    const { value, id, className } = props;
```
olemartinorg, line 40: It might be a bit misleading to just throw away all the other props here. I'm not sure where we should 'draw a line in the sand' when it comes to these pure components, but I would think this should be kept closer to the calling code rather than be a 'gotcha' inside here for future callers to stumble over.

olemartinorg (approved):
Looks good! :raised_hands:

src/app-components/Input/Input.tsx
cammiida: Not sure I agree with that. If we're following the same logic for everything, props like `hidden` or `disabled` would be exactly the same, yet it is a feature of that component to be able to be hidden or disabled based on other variables that might change. I don't see how `textOnly` is different 🤷‍♀️  
When it comes to the configuration of the layout component, it was already there and we have to ensure that that works as before.

src/app-components/Input/Input.module.css
cammiida: I agree it probably should be an import somewhere, but couldn't find a way to use the React component in a nice way without if feeling somewhat 'hacky'. Please tell me (or create a PR or smth) if you do find a way tho! 🙏

olemartinorg:
@cammiida I now get this warning in the console:
```React does not recognize the `textOnly` prop on a DOM element```

Seems like that prop is passed on to the underlying input element?

PR #2690, opened 2024-11-05
Simplifying node hierarchy: Removing `addChild`/`removeChild`

### LINKED ISSUE #2662: Duplicate components showing up in fringe cases
### Description of the bug

@bjosttveit ran the `all-process-steps.ts` against the latest `app-lib-dotnet` main branch and found that some components in repeating groups were rendered twice. This should obviously not happen.

### Steps To Reproduce

1. Run the test as described above
2. Find that it fails consistently in the `group` task

### Additional Information

_No response_

olemartinorg:
## Description

A part of the node generator/plugin functionality is to add and remove `LayoutNode` objects to node data/state. This functionality is removed here, and replaced by a simpler approach; just storing the IDs of target nodes in the store, and then using hooks like `useNode()` if objects are needed.

This removes some code and complexity, reduces run-time and memory allocation as a result, and prevents problems like #2662 from happening again. It also moves us in a slightly simpler direction, where the node generator can hopefully be written out of existence in the future.

Bullet points detailing the changes:
- Most of the code changes here are to work around `children`-like states now having strings (node IDs) instead of LayoutNode objects. Sometimes this includes splitting up components so that we can use hooks like `useNode()`. Often it also changes code from calling `GenericComponent` to instead call `GenericComponentById`.
- I changed the return type for `useNode()` to possibly return `undefined` in all cases. Previously it would seem like you'd always get a `LayoutNode` if you gave this thing a string, but that's not guaranteed.
- I split off some functionality in `useNavigatePage()`, as it caused re-rendering which lead me to the test failure that showed that `useNode()` should have indicated that it could return `undefined` all along.
- Child claim metadata is removed, as it was only used when passed to `addChild()` and `removeChild()`.
- Adding a concept of `idMutators` which is passed to plugins. Mutatators like this are used to mutate child component IDs (turning `currentName` into `currentName-0`).
- The Likert component did not correctly maintain row UUIDs. This is usually not a problem, as the use-case for that component calls for mostly static rows in the data model, but it could break if used dynamically. Added a `MaintainReferences` component to maintain this.
- `LikertRowSummary` called a hook conditionally in cases where the node object was undefined, breaking rule of hooks.
- I brought back the `isFresh` functionality to `RepeatingGroupTableRow`, as removing it in #2715 caused the broken test there.
- Repeating groups gets a hacky `internal` state in `item`. This is used to keep track of which child-ids should be added to the component after a row is added. This state cannot be stored in the `RepeatingChildrenPlugin` class instance, as that is shared among all `RepeatingGroup` components.
- Removing the `setRowUuids` queue, as it does the same thing as `setRowExtras`.
- Preventing `MarkPageHidden` from re-rendering when the current page changes (by introducing `useIsCurrentView()` and making sure it doesn't set state that was already set).

## Related Issue(s)

- #2683
- closes #2662

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels


olemartinorg:
Some snapshots in Percy are worrying - I'l look into that first before requesting review.

bjosttveit (approved):
LTGM!

PR #2699, opened 2024-11-07
Tiny benchmarking harness

martinothamar:
## Description

A benchmarking harness using `tinybench` as a suggestion to start tracking perf numbers. Using benchmarking provides us a paper trail - when we discover slow code that affect real usecases, we can provide benchmarks to show improvements in isolation. Upon future changes we can verify change in perf characteristics (for a variety of inputs) such as 
  * Throughput
  * Latency 
  * Std deviation (I'm not sure if it's possible to see the whole distribution, e.g. to rule out bimodal distributions)

Memory allocations is missing, I couldn't find a way to correctly correlate heap changes to benchmark runs, as there might be GCs occurring in the middle of runs.

The included benchmark is a microbenchmark comparing some implementations of `splitDashedKey` that did pop up in some profiling (although this is not an issue anymore, so it serves mostly as an example).

image

## Related Issue(s)

- N/A

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,8 @@
     "tsc": "yarn gen && tsc && tsc --project test/tsconfig.json",
     "tsc:watch": "yarn gen && tsc --watch",
     "tsc:watch:cypress": "yarn gen && tsc --watch --project test/tsconfig.json",
-    "lint": "yarn gen && eslint ."
+    "lint": "yarn gen && eslint .",
+    "bench": "tsx --expose_gc --tsconfig benchmarks/tsconfig.json benchmarks/${0}"
   },
   "devDependencies": {
     "@babel/core": "7.26.0",
@@ -116,10 +117,12 @@
     "source-map-loader": "5.0.0",
     "style-loader": "4.0.0",
     "terser-webpack-plugin": "5.3.10",
+    "tinybench": "^3.0.3",
     "ts-jest": "29.2.5",
     "ts-loader": "9.5.1",
     "ts-node": "^10.9.1",
     "tsconfig-paths": "^4.2.0",
+    "tsx": "^4.19.2",
     "typescript-plugin-css-modules": "^5.1.0",
     "use-immer": "^0.10.0",
     "utility-types": "3.11.0",
--- a/benchmarks/splitDashedKey.ts
+++ b/benchmarks/splitDashedKey.ts
@@ -1,0 +1,31 @@
+/* eslint-disable no-console */
+import { Bench } from 'tinybench';
+
+import { splitDashedKey, splitDashedKeyIterative, splitDashedKeyRegex } from 'src/utils/splitDashedKey';
+
+const inputs: string[] = ['mycomponent-0-1', 'f81d4fae-7dec-11d0-a765-00a0c91e6bf6-100-200'];
+
+const bench = new Bench({
+  name: 'splitDashedKey benchmark',
+  time: 100,
+  setup: (_task, mode) => {
+    // Run the garbage collector before warmup at each cycle
+    if (mode === 'warmup' && typeof globalThis.gc === 'function') {
+      globalThis.gc();
+    }
+  },
+});
+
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -112,10 +113,12 @@
     "source-map-loader": "5.0.0",
     "style-loader": "4.0.0",
     "terser-webpack-plugin": "5.3.10",
+    "tinybench": "^3.0.3",
     "ts-jest": "29.2.5",
     "ts-loader": "9.5.1",
     "ts-node": "^10.9.1",
     "tsconfig-paths": "^4.2.0",
+    "tsx": "^4.19.2",
```
olemartinorg, line 121: Nice! :star_struck: We already have `ts-node` as a typescript runner, but [I looked into it](https://github.com/privatenumber/ts-runtime-comparison) and `tsx` looks a lot better/more promising. I'll look into replacing our few usages of `ts-node` with `tsx` after this.

package.json
```diff
@@ -112,10 +113,12 @@
     "source-map-loader": "5.0.0",
     "style-loader": "4.0.0",
     "terser-webpack-plugin": "5.3.10",
+    "tinybench": "^3.0.3",
```
olemartinorg, line 116: I looked into it, and while `tinybench` seems to work well enough for the example use-case here, it doesn't seem there's any straight-forward ways to benchmark react components (including re-rendering) with it. There are libraries to do that, some of which seems to spin up headless chrome to run their benchmark in.

olemartinorg (approved):
Looks good! :raised_hands: Have you perhaps found a way to run and present these benchmarks directly on github? I'd love to have an automatic runner that tells us immediately if the PR is detrimental to performance (compared to a recent baseline from main).

package.json
martinothamar: oo didn't even notice, used `tsx` because it was part of the example code of tinybench

package.json
martinothamar: In principle I can't think of any reason why a React test renderer wouldn't run under NodeJS in this context when it is no problem in a jest testrunner? If it's practically much harder for some reason I'm not opposed to other tools at all, and I think we can even have multiple tools for different use-cases (though I have previously often just stuck to the same benchmarking harness for all level of benchmarks previously). But I think we should start to define a process around this to figure out

* What should we require benchmarks for
* How should we track the numbers over time

adamhaeger:
Hmm I dont really see how tinybench is useful for real world performance testing of a webapp. Seems more focused on low level testing of functions.

Id say some of the metrics we should focus on are:

1. Component Render Times
   - Initial Render Time
   - Re-render Times
   - Wasted Renders
2. React Reconciliation (Diffing Algorithm) Efficiency
3. State and Prop Update Times
4. Time to First Paint (TTFP) and First Contentful Paint (FCP)
5. Memory Usage
6. Time to Interactive (TTI)
7. Largest Contentful Paint (LCP)
8. Layout Shift (Cumulative Layout Shift - CLS)
9. Network Request Time for Data Fetching
10. JavaScript Bundle Size and Code Splitting

### Tools to Measure These Metrics
- React DevTools Profiler
- Chrome DevTools Performance Tab
- Web Vitals
- Lighthouse

martinothamar:
> Seems more focused on low level testing of functions.

Agreed, and I think we need both. We need high level testing that uses metrics focused on UX, but during the optimization process of our code we also need benchmarking that verifies that we actually optimize. 

The tools you mention are good for the explorative/analysis phase where you are running one-offs and high-level before/after. Benchmarks are good for iterating on optimizations when you've chosed your place of optimization. Proper measurement of optimization requires a lot of runs/executions so that the distribution of latency can be inspected. Higher level tests where you measure metrics to capture UX quality like the ones you mention can't be run a ton of times while also having a fast feedback loop

PR #2700, opened 2024-11-07
Associate test and staging envs in pipeline

martinothamar:
## Description

Matching PR for https://github.com/Altinn/app-lib-dotnet/pull/881

## Related Issue(s)

- N/A

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,7 +7,12 @@ jobs:
   build:
     name: Build and deploy
     runs-on: ubuntu-latest
-    environment: prod
+    environment: |-
+      ${{
+         contains(github.event.release.tag_name, 'preview') && 'test'
+      || contains(github.event.release.tag_name, 'rc')      && 'staging'
+      ||                                                       'prod'
+      }}
     permissions:
       id-token: write
     steps:
--- a/.github/workflows/revert.yml
+++ b/.github/workflows/revert.yml
@@ -12,7 +12,12 @@ jobs:
   revert:
     name: Roll back and deploy
     runs-on: ubuntu-latest
-    environment: prod
+    environment: |-
+      ${{
+         contains(github.event.release.tag_name, 'preview') && 'test'
+      || contains(github.event.release.tag_name, 'rc')      && 'staging'
+      ||                                                       'prod'
+      }}
     permissions:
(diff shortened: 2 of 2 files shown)


olemartinorg (commented):
I'm not sure this leads to any practical difference for app-frontend. When we release, we build using `webpack.config.production.js`, which has its mode hard-coded to production, and doesn't seem to care about environment variables at all. And if it did, what should be different for a 'staging' build? Included source-maps, non-minimized code, or something else?

martinothamar:
This is not about the build quality at all, but how we track release cadence and other things in the form of DORA metrics. The Swarmia tool we (maybe) will be using to track DORA metrics will track releases to test, staging and prod across all Altinn products, so we agreed that the mapping from preview and RC releases to these respective environments would be the most fitting

olemartinorg:
I see! :thinking: It might not map exactly to our release procedures, then. I see that lately we've been naming releases `rc`, and `preview`, but there's no naming conventions set in stone here. If possible, varying on the 'pre-release' checkbox in github might be a better approach. In other words, I can't think of any practical difference between our `rc` and `preview` releases other than their names.

martinothamar:
That's been the process thus far, but that is likely going to change. We are working on an ADR for amongst other things the release process, and part of that is trying to communicate practical differences between RC, preview and stable. We have discussed something along these lines:
* Production - stable releases, these should be bugfree
* RC - these are candidates for a stable release, we consider the feature work for the minor release done and don't expect to make any breaking changes
* Preview - work in progress, anything can happen

The reason we wanted RC in the first place is to have a more stable/complete release which could possibly motivate more apps to adopt these versions, while still leaving us some room to stabilize and sanity-check before releasing fully.

olemartinorg (approved):

bjosttveit:
I did not see this before merge, but I think this could break the release workflow for preview releases. I added the `prod` environment to the release workflow in https://github.com/Altinn/app-frontend-react/pull/2550 becuse the secrets needed for the `Azure login` step was added to the environment secrets for `prod`. I don't have access to check if secrets have been added for other environments but I would assume it has not

PR #2703, opened 2024-11-07
Performance: static componentConfigs

bjosttveit:
## Description

Profiling when loading `krt/krt-3010a-1` revealed that a lot of places use the generated `getComponentConfigs()` function, which showed up a lot in certain windows of time like these ~3 seconds here:

image

This is a very large structure of objects and classes which allocates a lot of memory and only leads to lots of GC. I don't see any reason this has to be recreated and discarded every time its needed as my understanding is that the classes are static (in practice) and don't hold any instance state?

## Related Issue(s)

- closes #{issue number}

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels


olemartinorg (approved):
As long as this builds, and don't cause any import order problems, it's all good! :raised_hands: I suspect it might become a bit more brittle in the future with a change like this, but only when import order changes causes a component definition to become undefined. Those are hard to debug, and is why I first made this change in #2244.

bjosttveit:
What about this instead?
image

olemartinorg:
Yep, that would work. And it wouldn't lead to the import order problem.

PR #2715, opened 2024-11-13
Safer way to select older data while re-generating parts of the node tree

olemartinorg:
## Description

In the main branch now, `useNodeItem()` will remember the last item it selected and give you that in case the node context readiness turns to 'not ready'. The intention here was to not re-render and to not give the component the new results before everything had settled (as the node hierarchy generator consists of multiple components that each could re-render and update their own part of the larger state tree).

This approach was OK, but still a bit simple and naive. With this PR I take the concept further by making a copy of `nodeData` when the node context turns 'not ready', and then select from that state instead. This also means that new components can be rendered while the nodes context is not ready (and they will select the older state from last time the context was ready).

In addition to this, I found and fixed a few issues that caused problems here and in #2690:
- Two contexts in `RepeatingGroup` had usages of `useNodeItemRef()`. The whole point is to select some state but not re-render if that state changes (this is mostly used to check rows and row state when opening/closing a row for editing, among other things). Instead of using a ref, which will always re-run the selector to keep the ref up to date, I made getter-functions that will select the state immediately when called instead. This fixed a problem with some state that was already outdated because the ref-selector did not get a chance to run yet, and it's much more efficient.
- I removed the `isFresh`  part from `RepeatingGroup` because it's no longer needed now, it won't be re-rendered until it is fresh again.
- When NodesContext is reset because of a layout change (i.e. when changing the layout in a cypress test, when using devtools to change it, or when proceeding to the next task), we now avoid running `removeNode` requests and updating the state. These are all added to the queue on unmount, and if they are committed _after_ the state has reset you could end up in a state where the newly added nodes (in the recently reset state) were removed. This also has the benefit of not even updating the state to remove all nodes after a `process/next`.
- When debugging I found that `GeneratorStages.useIsGenerating()` was broken. It used to work, but because the stages data was moved into NodesContext, it now only checks if the NodesContext exists - which it does. There was already another hook for this that worked (`GeneratorInternal.useIsInsideGenerator()`).
- A bit better error reporting in `useNodeTraversal()`

This was all extracted out from #2690 to make that PR focus on what it's all about; removing `addChild()` + `removeChild()`.

## Related Issue(s)

- #2690

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels
  
…onger the same as when *.generated.ts was generated (where there is no minifying)


olemartinorg:
I'm unable to reproduce the failing test, so I'm leaning towards calling it a fluke. :v: If it actually turns flaky from this change, I'll look into that later on.

src/utils/layout/NodesContext.tsx
```diff
@@ -1179,16 +1277,20 @@ export const NodesInternal = {
...
+  },
+  useLaxNodeDataSelector: () => {
+    const insideGenerator = GeneratorInternal.useIsInsideGenerator();
+    return Store.useLaxDelayedSelector({
       mode: 'innerSelector',
-      makeArgs: (state) => [((node) => selectNodeData(node, state)) satisfies NodePicker],
-    }),
+      makeArgs: (state) => [((node) => selectNodeData(node, state, insideGenerator)) satisfies NodePicker],
+    });
+  },
```
bjosttveit, line 1293: My first thought was that deps was missing, but maybe this does not need deps, as I suppose you cannot "move" in or out of the generator

bjosttveit (approved):

src/utils/layout/NodesContext.tsx
olemartinorg: True! :sweat_smile: Also, the `makeArgs` function can't take deps. The delayed selector will re-generate if the hook re-renders.

PR #2724, opened 2024-11-19
Faster save processing, reworking `useNodeValidation()`

olemartinorg:
## Description

When form data is saved, we get new validations from the backend (when the app has implemented backend validation, at least). Previously, we saved the last processed validations pretty much everywhere (luckily only as a reference) so that we could make sure all `StoreValidationsInNodeWorker` components had done their job and stored new validations if needed. This caused lots of state updates on all nodes for every save operation, and worsened performance.

Now, the `useNodeValidation()` hook will instead subscribe to the validation store and will only re-run validation when that changes. It will then save which backend validations have been processed into the `Registry`, so that we can `useWaitForNodesToValidate()` to wait until they have processed their validations (or, at least, `useNodeValidation()`  has run to completion).

This also means that `NodesContext` no longer has to store the latest validations, and also means that it can now only be 'ready' or 'not ready' (no longer 'processing the last save'). The node generator will now only become 'not ready' after a save if new components are introduced (typically if a row has been added).

In addition to this, a few minor changes. Either because they were tiny, or stuff I had to do because tests failed:
- Getting rid of the `renderPresentation` flag in `<Loader />`, and figuring out if the Presentation component has rendered via a Context. I've seen glimpses of double-PresentationComponent rendering lately, and this should fix it.
- Optimizing some routes. There's no need for a different route to `<FormFirstPage />` as `<Form />` will also detect that case and redirect you to the first page.
- Fixing the `navigate()` call in the Subform component that broke on me. It somehow started to remove the main form page from the URL when navigating to a relative path (`.../Task_1/mainPage` navigating to the relative `subFormComponentId/subFormDataElementId` turned into `.../Task_1/subFormComponentId/subFormDataElementId`).

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels
  
…here. Making sure to not render the presentation stuff inside itself.


bjosttveit (approved):
Very nice!

PR #2729, opened 2024-11-19
BugFix: change key for geojson so it updates on change

### LINKED ISSUE #2727: Geometry in map component not updating when changed
### Description of the bug

The map-component renderes geometries given to it by the datamodel. The component expects an array of objects `{ data: string; label?: string; }`.
When objects are added to the array, the map component renderes them properly. On the other hand, if you change the data/polygon part of a geometry that has previously been rendered on the map, the map moves to the new geometry location, but the geometry itself remains unchanged. Since the map moves to the new placement, that means the geometry bounds are recalculated, showing that the datamodel has been updated correctly. It seems like this happens as long as the data/gemoetry object has the same index and label in the datamodel array as it had on previous renders/updates. Reversing the list in the app before storing it in the datamodel so the index of the object has been changed since last update makes the geometry behave correctly/as expected.

The problem is propably with the key of the geojson components in the map component. 
```ts
{geometries?.map(({ data, label }, i) => (
        <GeoJSON
          key={`${i}-${label}`}
```

### What happened
Geometry is placed in datamodel and rendered in map. When the geometry is changed in the datamodel, the geometry remains in place on the map, instead of moving where it should.

### What should have happened
Map should have moved to the new placement of the geometry, and the geometry object should have moved and adapted to the changed geometry data.

### Steps To Reproduce

1. Make an app with a map and geometries that can be changed by the user.
2. Place geometries in the datamodel/map
3. Watch them render on the map
4. Change/update the data prop/field of one of the geometries
5. The map should move the view to the updated placement of the geometry, while the geometry stays in place.

### Additional Information

Placed geometry on map, and rendered correctly
Image

Changed the data of a geometry, but kept same label and placement in the array. The map moves but geometry does not.
Image

Expected behavior is that the geometry updates to new place:
Image

MartinRoberg:
## Description

Using altinnrowid from datamodel instead of index+label as key for geojson-components to make sure the geometries update when changed.
Added altinnrowid to geometry and rawgeometry type, and updated utils.ts for map to take altinnrowid into geometry from rawgeometry.

## Related Issue(s)

- closes #2727 

## Verification/QA

  - [x] Creator of the original issue (or service owner) has been contacted for manual testing (or will be contacted when released in alpha)
- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels
  - [x] I don't have permissions to add labels, please help me out

### DIFF
--- a/src/layout/Map/Map.tsx
+++ b/src/layout/Map/Map.tsx
@@ -137,9 +137,9 @@ export function Map({
           subdomains={layer.subdomains ? layer.subdomains : []}
         />
       ))}
-      {geometries?.map(({ data, label }, i) => (
+      {geometries?.map(({ altinnRowId, data, label }) => (
         <GeoJSON
-          key={`${i}-${label}`}
+          key={altinnRowId}
           data={data}
           interactive={false}
         >
--- a/src/layout/Map/types.ts
+++ b/src/layout/Map/types.ts
@@ -1,11 +1,13 @@
 import type { GeoJSON } from 'geojson';
 
 export type RawGeometry = {
+  altinnRowId: string;
   data: string;
   label?: string;
 };
 
 export type Geometry = {
+  altinnRowId: string;
   data: GeoJSON;
   label?: string;
 };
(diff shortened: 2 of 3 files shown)


bjosttveit (approved):
### Good catch! 🙌 🙌

PR #2747, opened 2024-11-22
Pass tracecontext from PDF generator cookies back to app

martinothamar:
## Description
When rendering PDF, get W3C `traceparent` propagated by the app and pass it back to the app when making subsequent requests.

Backend PR: https://github.com/Altinn/app-lib-dotnet/pull/924

Tested locally using Application Insights both with  OTel and the classic App Insights SDK

I don't know what the best place for this code is, but at least it needs to be pretty far upstream so that the interceptor is added before any downstream requests are made.

## Related Issue(s)

- N/A

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels

### DIFF
--- a/test/e2e/integration/frontend-test/pdf.ts
+++ b/test/e2e/integration/frontend-test/pdf.ts
@@ -28,11 +28,12 @@ describe('PDF', () => {
       'altinn-telemetry-tracestate': tracestateValue,
     };
 
-    cy.intercept(`*/api/**`).as('apiRequests');
+    cy.intercept(`*/instances/**`).as('apiRequests');
     cy.goto('message', { cookies });
 
     cy.testPdf(false, () => {
       cy.get<Interception<unknown, unknown>>('@apiRequests').then(({ request }) => {
+        cy.log('Request intercepted:', request.url, JSON.stringify(request.headers, null, 2));
         expect(request.headers, 'request headers').to.include({
           'X-Altinn-IsPdf': 'true',
           traceparent: traceparentValue,
--- a/test/e2e/integration/frontend-test/pdf.ts
+++ b/test/e2e/integration/frontend-test/pdf.ts
@@ -25,10 +25,7 @@ describe('PDF', () => {
       'altinn-telemetry-traceparent': traceparentValue,
       'altinn-telemetry-tracestate': tracestateValue,
     };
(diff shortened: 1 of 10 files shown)


bjosttveit (approved):
Looks good to me. Although since this appears to require no hooks, I suppose it does not need to lie in the render function of index? Not completely sure what the best place is either but i guess the closest thing we have to an established pattern for things like this would be something similar to `src/features/baseurlinjection` which is just imported directly in the index file

martinothamar:
Extracted it into a separate function in a file in `src/feature` (but still use the same evt handler)

bjosttveit:
> Extracted it into a separate function in a file in `src/feature` (but still use the same evt handler)

That looks great, much cleaner

olemartinorg:
@martinothamar
> Are there any good ways to unit-test this stuff? 
I would probably have made a Cypress test for this. Loading an existing instance with `?pdf=1`, intercept all requests and assert they contain the header we want.

test/e2e/integration/frontend-test/pdf.ts
```diff
@@ -18,6 +18,28 @@ describe('PDF', () => {
...
+      'altinn-telemetry-tracestate': tracestate,
+    };
+    for (const [key, value] of Object.entries(cookies)) {
+      cy.setCookie(key, value);
+    }
+    cy.goto('message');
+
+    cy.intercept({ method: 'GET', url: '**/applicationmetadata' }).as('appMetadata');
+
+    cy.testPdf('message', () => {
```
bjosttveit, line 35: The first argument here is the name of the (percy) snapshot to take (which will get `(PDF)` appended to it). There is already a PDF-test with this snapshot name, but in this case its probably not necessary to take a snapshot at all so you can set it to `false` instead.

bjosttveit:
Have you tried running cypress locally? 😅 
`npx cypress open -e environment=tt02`
if you set `it.only(...)` on the test and click the `pdf` test in the cypress runner you can get a little faster feedback than testing in GitHub

martinothamar:
Yeah the normal runner doesn't work for me, probably because I'm in WSL. Managed to run it in Docker but I'm not getting the same error so I think I can't run this locally atm

PR #2754, opened 2024-11-25
Add support for new attachment API and improve UX for error messages

### LINKED ISSUE #1992: Give user more time to read validation issues on file upload
### Description

This error message disappears too quickly. User has to "catch" this with the pointer before it disapears. This is not very user-friendly. Would be better with a alert type where user has to click Ok button, or something like that. 

image

### Additional Information

Currently the ValidationIssues that prevented upload of a file is shown in a toast that disappears after a few seconds. This is likely fine for most users when the message is only "This is a file of the wrong type", but some service owners upload csv/excel/xml/json files where there might be lots of issues with the file that might need manual fixing before uploading again. They need to be able to read all the information before it is gone.

Typical suggestions might be to use

Dismissible full screen modal
Show the errors in a dismissible box inside the upload component (auto dismiss on next upload?)
It would also be possible to show single issue and issue list differently
Tagging as accessibility because using "Toast" for error messages is problematic as slow users might not have time to read even really short messages.

### LINKED ISSUE #2012: FileUpload - Attachment box gets highlighted with red indicating error for allowed content type
### Description of the bug

This was discovered by one of the testers, when opening the app in tt02. 
When uploading a .zip folder **by dragging the file** to the component, the box turns red even though the content type is allowed. As seen in the video, the zip is actually uploaded. However, the red box causes confusion for those who associate it to not being able to upload.

https://github.com/Altinn/app-frontend-react/assets/61122289/61dc5db3-6fa4-4ce6-863f-7725d5d195d0

### Steps To Reproduce

I have not been able to reproduce the error myself. There might be something connected to different mime's of zip, but i'm not sure. 

### Additional Information

in applicationmetadata:
```json
   {
      "id": "3-1-7-FileUpload",
      "allowedContentTypes": [
        "application/pdf",
        "application/vnd.ms-excel",
        "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet",
        "application/msword",
        "application/vnd.openxmlformats-officedocument.wordprocessingml.document",
        "application/rtf",
        "image/jpg",
        "image/jpeg",
        "text/plain",
        "application/vnd.openxmlformats-officedocument.presentationml.presentation",
        "application/vnd.ms-powerpoint",
        "application/zip",
        "application/x-zip",
        "application/x-compressed",
        "application/x-zip-compressed",
        "application/vnd.oasis.opendocument.presentation",
        "application/vnd.oasis.opendocument.spreadsheet",
        "application/vnd.oasis.opendocument.text"
      ],
      "taskId": "Task_1",
      "maxSize": 200,
      "maxCount": 25,
      "minCount": 0,
      "grouping": "3.1.7-FileUpload.grouping-title",
      "enablePdfCreation": true,
      "enableFileScan": true,
      "validationErrorOnPendingFileScan": false,
      "enabledFileAnalysers": [],
      "enabledFileValidators": []
    },
```

### LINKED ISSUE #2596: Use new endpoint for POST when saving attachments
### Description

In https://github.com/Altinn/app-lib-dotnet/issues/850, we add a new endpoint for POST, so that service owners can write code that modifies instance data (mutate form data or add/remove elements) when attachments gets uploaded.

Frontend needs to use the new endpoint and update its internal state when the backend version is new enough (likely 8.5.0 (not preview), we will see)

### Additional Information

_No response_

bjosttveit:
## Description

Will now use the new attachment API if the backend version is up to date. Since the new API can make changes to form data, validations, and instance data we have to upload attachments in series instead of in parallel (if multiple are selected at the same time) to prevent race conditions. However, it will gather all of the changes and update the state only once at the end. This could really use some cypress tests to verify that form data etc. is updated correctly, but its a lot of set up, and I cannot spend that much more time on this right now.

Errors with uploading a file is no longer showed in a toster (only if you drag more files than allowed, you still get a toaster saying you selected too many files), but is shown below the fileupload component, the user needs to close the alert for it to disappear. Attachments which fail to upload are stored in the node-state instead of the react-component's state so that they don't disappear when switching pages. It will however disappear on refresh since nothing gets stored in the backend.

This, together with https://github.com/Altinn/app-lib-dotnet/issues/850 deals with some of the issues in https://github.com/Altinn/app-frontend-react/issues/2706. Specifically, enabling updating the form data / validations and the app frontend getting these changes as well as improved UX for error messages that caused the upload to be rejected.

This is what it looks like now. Long filenames are truncated in the middle.

image

If there are multiple issues with an attachment, it will be shown as a list. If there are more than 3 issues you have to expand to show them all so that it does not immediately take up a lot of space if the list is very long.

image

## Related Issue(s)

- closes #2596 
- closes #1992 
- closes #2012

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels


olemartinorg:
@bjosttveit Some questions before I go into reviewing the actual code:

> However, it will gather all of the changes and update the state only once at the end.

Will it handle failures, then? What about timeouts, i.e. you internet speed drops to zero (or you upload a gigantic attachment). Will you risk losing out on that state update for previously uploaded attachments if you (for example) navigate to another page in the meantime?

> Long filenames are truncated in the middle.

Will you still be able to see the full filename on hover?

> If there are more than 3 issues you have to expand to show them all so that it does not immediately take up a lot of space if the list is very long.

Nice! Should this be a button instead of a link, maybe? :thinking:

bjosttveit:
> Will it handle failures, then? What about timeouts, i.e. you internet speed drops to zero (or you upload a gigantic attachment). 

The upload itself should not throw any uncaught exceptions. The promise for uploading each file has a `.then` which adds a success result and updates the intermediate state, and a `.catch` which adds a failure result (and does not update the intermediate state). So in theory any number of attachments failing to upload for any reason should not affect successful uploads. 

> Will you risk losing out on that state update for previously uploaded attachments if you (for example) navigate to another page in the meantime?

What do you mean by navigate to another page? Navigating within the app should not affect things. The FileUpload component simply calls the (async) function without awaiting it or expecting a return value. All of the state updates are handled by the attachmentplugin (store) and propagate back to the component throught hooks (as it was before).

Navigating away (like refreshing) will maybe cancel pending requests? But you would also get the updated form data for what has already completed when loading I guess.

> Will you still be able to see the full filename on hover?

Yes

> Nice! Should this be a button instead of a link, maybe? 🤔

It is a button 😄  (variant tertiary)

src/layout/FileUpload/DropZone/DropzoneComponent.tsx
```diff
@@ -88,10 +84,9 @@ export function DropzoneComponent({
             : undefined
         }
       >
-        {({ getRootProps, getInputProps, isDragActive, isDragReject }) => {
+        {({ getRootProps, getInputProps, isDragActive }) => {
```
olemartinorg, line 87: Does this mean you fixed #2012? :partying_face:

olemartinorg (approved):
Good stuff! :muscle:

src/layout/FileUpload/DropZone/DropzoneComponent.tsx
bjosttveit: That was not the intention but it probably did 😅 I was just confused that the box turned red (and stayed red) after trying to upload something the dropzone rejected

PR #2829, opened 2024-12-18
Fixing failing `component` lookups when deeply linked

olemartinorg:
## Description

Multiple apps had problems with `["component", ...]` expressions after the performance-related fixed in the release on the 27th of november. This problem was caused by the new multiple-delayed-selectors functionality, specifically the code that tried to reduce the amount of re-renders triggered when state changes. Unit-testing uncovers that React _marks the component for re-rendering_, so we don't have to worry about multiple re-renders being triggered after all. Removing that code fixes the problem.

Also:
- Adding a unit-test to test simple functionality for multiple delayed selectors
- Adding a cypress test more closely resembling the actual use-case that triggered this problem. 10 components are chained visible via `["component", ...]` expressions, and without this fix the third one failed to show up (now all of them do).
- Changing useIsDev() from a hook-like signature to just a plain function (with a cache when not unit-testing it). This way I could call it from inside the delayed selectors.
- Developer exprience: Logging to the console when a delayed selector is called multiple times, but returns a different result. This is usually what happens if you forget to pass your dependencies to the deps-array (second argument).

## Related Issue(s)

- https://altinn.slack.com/archives/C02EJ9HKQA3/p1733256844754139

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels

### DIFF
--- a/src/hooks/delayedSelectors.ts
+++ b/src/hooks/delayedSelectors.ts
@@ -280,11 +280,9 @@ class SingleDelayedSelectorController<C extends DSConfig> extends BaseDelayedSel
 }
 
 class MultiDelayedSelector<C extends DSConfig> extends BaseDelayedSelector<C> {
-  private changeCount = 0;
-  private lastSelectChangeCount = 0;
-  private onChange: (lastSelectChangeCount: number) => void;
+  private readonly onChange: () => void;
 
-  constructor(props: DSProps<C>, onChange: (lastSelectChangeCount: number) => void) {
+  constructor(props: DSProps<C>, onChange: () => void) {
     super(props);
     this.onChange = onChange;
   }
@@ -293,16 +291,8 @@ class MultiDelayedSelector<C extends DSConfig> extends BaseDelayedSelector<C> {
     return this.selectorFunc;
   }
 
-  public setChangeCount(count: number) {
-    this.changeCount = count;
-  }
-
-  protected onCallSelector(): void {
-    this.lastSelectChangeCount = this.changeCount;
-  }
-
   protected onUpdateSelector(): void {
-    this.onChange(this.lastSelectChangeCount);
+    this.onChange();
   }
 }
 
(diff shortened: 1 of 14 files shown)


src/utils/isDev.ts
```diff
@@ -1,26 +1,30 @@
...
+
 /**
  * Indicates whether the application is running in a development environment.
  * This can be either through LocalTest, altinn studio preview or TT02.
  */
-export function useIsDev(): boolean {
-  const isLocalOrStaging = useIsLocalOrStaging();
-  const isStudioPreview = useIsStudioPreview();
-  return isLocalOrStaging || isStudioPreview;
+export function isDev(): boolean {
```
bjosttveit, line 10: Nice, I have thought about doing this several times but never gotten around to it 😆

src/hooks/delayedSelectors.ts
```diff
@@ -28,7 +29,10 @@ type ModeFromConf<C extends DSConfig> = C extends DSConfig<any, infer M> ? M : n
  * will not be able to be used as a cache key.
  */
 export function useDelayedSelector<C extends DSConfig>(props: DSProps<C>): DSReturn<C> {
-  const state = useRef(new SingleDelayedSelectorController(props));
+  const state = useRef<SingleDelayedSelectorController<C>>();
+  if (!state.current) {
+    state.current = new SingleDelayedSelectorController(props);
+  }
```
bjosttveit, line 35: What difference does this make? 🤔

bjosttveit (approved):
Very nice, great tests 🥇

src/hooks/delayedSelectors.ts
bjosttveit: Oh, I never thought about that, but of course it does, its just a function parameter after all 🤦

PR #2859, opened 2025-01-02
Fix future issues with processing layout-sets

bjosttveit:
## Description

These are some fixes that for things that are not actually problems right now, but could become problems in the near future.

In `feat/navigation` I added some ui settings to the `layout-sets` file that needed some post-processing. It turned out that adding a `process`-function that changed the results somewhat (creates a new object) would cause problems when the query cache became stale after 10 minutes. It would trigger it to re-render with the same data, but because of the processing it would return a new object which caused the nodes to be re-created. (this is a known weakness and gets logged to the browser console). To deal with this I added a `useMemo` so that the same data should not create a new object if it runs multiple times.

A consequence of this issue was that nodes got added multiple times because "direct" children did not get deleted from the layoutPage object when removing nodes. (this was done intentionally since "direct" children are never expected to change within the same layout-set) But nevertheless I made sure these are removed as well. Additionally, the way children where removed from the `allChildren` list was a bit bad, since if the `indexOf` does not find the node it returns `-1` which in the `splice`function means remove the last element, but in that case we should not remove anything.

## Related Issue(s)

- https://github.com/Altinn/app-frontend-react/pull/2831

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels

### DIFF
--- a/src/core/contexts/queryContext.tsx
+++ b/src/core/contexts/queryContext.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useMemo } from 'react';
 import type { PropsWithChildren } from 'react';
 
 import type { UseQueryResult } from '@tanstack/react-query';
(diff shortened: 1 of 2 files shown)


olemartinorg (approved):
Looks good! :raised_hands:

PR #2860, opened 2025-01-02
Installing `cypress-fail-on-console-error` + console error fixes

olemartinorg:
## Description

I recently ran the `all.test.tsx` 'unit' tests that run app-frontend on many real-life apps to evaluate their layouts, etc. Quite alarmingly, every app I tested gave me a cryptic `Maximum update depth exceeded` error. I immediately looked to our test apps to figure out if it also happened when running in the browser - but luckily it didn't.

Still, I think errors in the browser console should make cypress tests fail. If we add some new functionality but mistakenly duplicate the `key` prop for some components (which happens somewhat regularly), nobody will notice unless it actually causes a bug - or someone decides to open the console to have a look. I think those cases (and lots of others) should alert us by failing tests outright instead.

So this PR introduces that, with the following fixes:
1. Being explicit about props we forward in `Button.tsx` to avoid spreading props that are unknown to native components downstream.
2. In `DataModelsProvider.tsx` there was a bug when loading data models for a subform (right after it had been created), where it didn't find the data element and caused it not to be loaded. It happened rarely, but seemed to be caused by the instance data not being updated in time before `DataModelsProvider` runs.
3. I removed parts of the `window.CypressLog()` stuff. I don't think many used it, and it would no longer log those lines to a file when a `console.log()` caused a test to fail (as `afterEach()` seemed to not be triggered correctly). Logging directly to the Cypress log works instead, and now bypasses the `cypress-fail-on-console-error`.
4. In `useGetOptions.ts` some props were set to undefined instead of being deleted, which caused these props to be passed to native radio components downstream (which logs an error to the console). I recently introduced this code, and hadn't noticed the error in the console, further driving home the point about why we need this stuff (or at least, I do). :wink: Removing the prop spread in `ControlledRadioGroup.tsx` that caused this as well.
5. `useSubformMutations.tsx` logged an error to the console when it was more logical to log it to devtools (the log for app developers, not the log for us).
6. I'm removing the warning about delayed selectors misuse. It turns out that warning was triggered even in cases where it's obviously a false alarm - i.e. when purely selecting data without any dependencies.
7. The actual cause behind the 'Maximum update depth exceeded` error I originally saw was that the `useFullErrorList()` hook re-created the object every render (after the zustand v5 upgrade). Fixing that by using a memo selector instead.
8. When running tests for all apps, I once saw a problem where selecting an array and spreading it caused an undefined error in `RepeatingChildrenPlugin.tsx`.

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels


src/features/options/useGetOptions.ts
```diff
@@ -219,7 +219,10 @@ export function useFilteredAndSortedOptions({
 
     // Always remove the rowNode and dataModelLocation at this point. It is only to be used in the filtering
     // process, and will not ruin the comparison later to make sure the state is set in zustand.
-    options = options.map((option) => ({ ...option, rowNode: undefined, dataModelLocation: undefined }));
+    options.map((option) => {
```
bjosttveit, line 222: Nitpick, but since you are now mutating the array instead of reassigning it, is it clearer if you switch to a `.forEach` instead?

src/utils/layout/plugins/RepeatingChildrenPlugin.tsx
```diff
@@ -231,7 +231,7 @@ export class RepeatingChildrenPlugin<E extends ExternalConfig = typeof defaultCo
     const out: string[] = [];
     for (let i = 0; i < rows.length; i++) {
       const row = rows[i];
-      row && out.push(...row.itemIds);
+      row && out.push(...(row.itemIds ?? emptyArray));
```
bjosttveit, line 234: Also a nitpick, but you can also just check if its undefined before pushing instead

```suggestion
      row?.itemIds && out.push(...row.itemIds);
```

bjosttveit (approved):
Very nice, I like it

src/features/options/useGetOptions.ts
olemartinorg: Good choice. I personally don't see much difference in those two, but I went for a `for (const option of options) {...}` instead.

PR #2890, opened 2025-01-09
Feature/signing cta

### LINKED ISSUE #2641: Signature Call-To-Action (CTA)
### Description

Button which invokes the signing action. 
Should be enabled when:
-  the user has confirmed that they have read and understood the contents that the are to sign for
-  the user has signing rights

Should maybe not be shown at all if the user does not have signing rights.

### Additional Information

_No response_

cammiida:
## Description

Kan fort lettest leses commit for commit 😊 

## Related Issue(s)

- closes #2641 

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels


src/layout/SigningStatusPanel/PanelAwaitingCurrentUserSignature.tsx
```diff
@@ -0,0 +1,81 @@
...
+          onClick={() => handleSign()}
+          disabled={!confirmReadDocuments}
+          size='md'
+          color='success'
+        >
+          <Lang id={signingButtonText} />
+        </Button>
+      }
+      description={checkboxDescription}
+      errorMessage={error ? langAsString('signing.error_signing') : undefined}
```
olemartinorg, line 70: This and `heading` would be better if you just pass the `<Lang />` component instead (and update the `SigningPanel` component to accept react components as input instead of just strings). By doing that, you:

1. Might not need to import and `useLanguage()` here, making sure this component doesn't have to re-render if anything in the language-related hooks change. Smaller parts of the tree that has to re-render = better performance.
2. `langAsString()` will remove HTML and markdown formatting from text resources, which we officially try to support whenever possible. I'm not sure if that makes sense to use here, but we've tried to err on the side of allowing it whenever possible. Using `<Lang />` will format the string appropriately.

src/layout/SigningStatusPanel/PanelNoActionRequired.tsx
```diff
@@ -0,0 +1,64 @@
+import React from 'react';
+import { useParams } from 'react-router-dom';
```
olemartinorg, line 2: I know me might disagree on this, but it would be nice for the entire code base to be consistently using `AppRoutingContext` for these things instead of react router directly.

src/layout/SigningStatusPanel/PanelSigning.tsx
```diff
@@ -0,0 +1,124 @@
...
+}
+
+type RejectTextProps = {
+  node: BaseLayoutNode<'SigningStatusPanel'>;
+};
+
+function Reject({ node }: RejectTextProps) {
+  const modalRef = useRef<HTMLDialogElement>(null);
+  const { next, busy } = useProcessNavigation() ?? {};
+  const { textResourceBindings } = useNodeItem(node);
```
olemartinorg, line 71: If you're only accessing one property, the selector mode is better:
```suggestion
  const textResourceBindings = useNodeItem(node, (i) => i.textResourceBindings);
```

This will only re-render when the actual text resources change (not when any other property changes), so it's similar to `SELECT justThatFieldINeeded FROM table` instead of `SELECT * FROM table`.

Also, I hope we one day get to run expressions _here_ instead of in the hierarchy generator, and at that point selecting just what you need will hopefully be more future-proof.

src/layout/SigningStatusPanel/SigningStatusPanel.module.css
```diff
@@ -0,0 +1,42 @@
...
+  text-decoration: none;
+}
+
+.signingPanel {
+  padding-top: var(--fds-spacing-8);
+  display: flex;
+  flex-direction: column;
+
+  .contentContainer {
+    padding: var(--fds-spacing-2); /* TODO: remove when MUI Grid is removed?*/
```
olemartinorg, line 18: MUI grid has been removed! :partying_face:

src/layout/SigningStatusPanel/config.ts
```diff
@@ -0,0 +1,161 @@
...
+export const Config = new CG.component({
+  category: CompCategory.Presentation,
+  capabilities: {
+    renderInTable: false,
+    renderInButtonGroup: false,
+    renderInAccordion: false,
+    renderInAccordionGroup: false,
+    renderInCards: false,
+    renderInCardsMedia: false,
+    renderInTabs: false,
```
olemartinorg, line 13: I don't see why this couldn't render in Accordion, Cards and Tabs? :thinking: Is is just because we haven't tested that, or because we want this to strictly not render inside anything else? I believe opening these up for everything that's not obviously breaking stuff is the best. (Such as rendering in `AccordionGroup` or `ButtonGroup`, or a table cell for that matter.) That way, app developers will get the final say instead of us. We've had pull requests from external contributors before with one-line changes that only allow one type of component to be rendered inside another, and at that point we're just slowing people down by being gatekeepers.

src/layout/SigneeList/SigneeListComponent.tsx
```diff
@@ -30,7 +30,12 @@ export function SigneeListComponent({ node }: PropsFromGenericComponent<'SigneeL
   });
 
   if (taskType !== ProcessTaskType.Signing) {
-    return <Lang id='signee_list.wrong_task_error' />;
+    return (
+      <Lang
+        id='signing.wrong_task_error'
+        params={['SigneeList']}
+      />
+    );
```
olemartinorg, line 38: This should probably be part of component validation, instead of just printed out to the user. Logged in DevTools, red border, all the bells and whistles. If we decide to integrate that log into (for example) application insights in the future, this thing won't circumvent that.

src/layout/SigningStatusPanel/SigningStatusPanelComponent.tsx
```diff
@@ -0,0 +1,88 @@
...
+  const canWrite = useIsAuthorised()('write');
+  const { langAsString } = useLanguage();
+
+  if (taskType !== ProcessTaskType.Signing) {
+    return (
+      <Lang
+        id='signing.wrong_task_error'
+        params={['SigningStatusPanel']}
+      />
+    );
```
olemartinorg, line 35: Same thing here, should use component validation. Take inspiration from `Subform`, the `renderLayoutValidators()` function and `<SubformValidator />` - I think that's the currently easiest approach to it.

olemartinorg (approved):
Good stuff!! :muscle: :rocket: I have quite a few comments here, and I can't really comment about the domain-specific parts about signing (I haven't followed the work in that task force closely).

src/layout/SigningStatusPanel/SigningStatusPanelComponent.tsx
```diff
@@ -0,0 +1,88 @@
...
+        isOnBottom
+      >
+        <div className={classes.loadingContainer}>
+          <Spinner title={langAsString('signing.loading')} />
+        </div>
+      </Panel>
+    );
+  }
+
+  if (currentUserStatus === 'awaitingSignature') {
```
HauklandJ, line 51: The logic looks correct here. 
We should have tests for all the cases:
User in signeeList has not yet signed -> AwaitingCurrentUserSignaturePanel
User in signeeList has signed -> NoActionRequiredPanel
User with write access and in signeeList has not yet signed -> AwaitingCurrentUserSignaturePanel
User with write access and in signeeList has signed, but other required has not -> AwaitingOtherSignaturesPanel
User with write access and in signeeList has signed, and all other required have too -> SubmitPanel

HauklandJ (approved):
🙌

src/language/texts/nb.ts
```diff
@@ -345,9 +345,40 @@ export function nb(): FixedLanguageList {
       validation_invalid_response_from_server: 'Det oppstod en feil. Vennligst prøv igjen senere.',
       unknown_error: 'Ukjent feil. Vennligst prøv igjen senere.',
     },
+    signing: {
+      awaiting_signature_panel_title: 'Signer skjemaet',
+      checkbox_label: 'Jeg bekrefter at informasjonen og dokumentene er korrekte.',
+      sign_button: 'Signer skjemaet',
+
+      submit_panel_title: 'Du kan nå sende inn skjemaet',
+      submit_panel_description: 'Alle parter har signert! Velg Send inn for å fullføre.',
```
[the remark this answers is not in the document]
cammiida: Agreed! I'll discuss it with the team 👌

src/layout/SigningStatusPanel/config.ts
```diff
@@ -0,0 +1,161 @@
...
+  )
+
+  .addTextResource(
+    new CG.trb({
+      name: 'no_action_required_panel_title_has_signed',
+      title: 'Go to inbox panel title signed',
+      description:
+        'The title of the panel that is displayed when the user has signed and no further action is required',
+    }),
+  )
```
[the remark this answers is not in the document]
cammiida: Ah, true! This was a slip up from my side. I'll make them camelCase 👌

src/layout/SigningStatusPanel/PanelAwaitingCurrentUserSignature.tsx
```diff
@@ -0,0 +1,81 @@
...
+    },
+    onSuccess: () => {
+      queryClient.invalidateQueries(signeeListQuery(partyId!, instanceGuid!));
+    },
+  });
+
+  const [confirmReadDocuments, setConfirmReadDocuments] = useState(false);
+
+  if (!canSign) {
+    return langAsString('signing.error_missing_signing_rights');
```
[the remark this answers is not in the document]
cammiida: Hm, yeah was wondering about this case. It shouldn't really happen, since you normally wouldn't even have access to the form if you haven't been delegated the rights. But in case there is some error between our code and authorization it is nice to have something displayed 🤷‍♀️ 

Can write a comment about that this shouldn't happen?

src/layout/SigneeList/api.ts
```diff
@@ -20,6 +21,7 @@ export const signeeListQuery = (partyId: string, instanceGuid: string) =>
   queryOptions({
     queryKey: ['signeeList', partyId, instanceGuid],
     queryFn: () => fetchSigneeList(partyId, instanceGuid),
+    refetchInterval: 1000 * 60, // 1 minute
```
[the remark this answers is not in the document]
olemartinorg: Ugh, I see.. About time we get websockets up and running, but auth in those would probably be a nightmare.

src/layout/SigningStatusPanel/PanelAwaitingCurrentUserSignature.tsx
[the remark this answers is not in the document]
olemartinorg: Ah, if it shouldn't happen, I think a hard crash is better - the 'Unknown error' page is much more noticeable than this. If a user saw this message they'd probably think "I'll wait and see it works later", but a hard crash informs them about how they should report bugs to us.

src/layout/SigningStatusPanel/PanelNoActionRequired.tsx
cammiida: You're right, we do disagree on this 🙈  I don't see that it would gain such performance benefits that it outweighs the issues of added complexity, maintenance required and the fact that this can hide bugs that would be hard to resolve. How many times does the URL actually change? The most changes that I can see is from task page to task page. And shouldn't the components then rerender anyways because the layout is different?

src/layout/SigningStatusPanel/PanelAwaitingCurrentUserSignature.tsx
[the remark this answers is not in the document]
cammiida: Yeah, good point!

src/layout/SigningStatusPanel/SigningStatusPanel.module.css
cammiida: I realise that it still has the same functionality with padding and stuff, so still need this to make it look right :/

src/layout/SigningStatusPanel/PanelNoActionRequired.tsx
olemartinorg: Yep, maybe it doesn't change all that much. When I implemented this, there was a very noticeable difference in performance, but that was probably also mostly because of hot paths (which this isn't). The main problem I have with using `react-router-dom` directly now is not necessarily because of performance or re-renders, but because of consistency in the code base. Having two ways of doing one thing in the same code base is a 'smell'.

PR #2901, opened 2025-01-13
Fix payment loading for custom payment pages

### LINKED ISSUE #2740: Components in payment layout-set show while create payment is loading
### Description of the bug

Patentstyret want to show an extra paragraph in the payment page to inform the user that they can try another payment method if Nets Easy fails for some reason.

However, the paragraph they added shows up while "Create payment" is being called. Our loading indicator shows next to it. It would be best if no other components than the loading indicator showed. 

See if it's possible to solve somehow. If we can't solve it, we probably need to solve the issue of showing the paragraph in another way. Some kind of failure handling in the payment component.

### Steps To Reproduce

1...

### Additional Information

_No response_

Magnusrm:
## Description

This PR moves the "automatic" logic from the PaymentComponent into a new context, "PaymentProvider".
This allows us to set a loading state that displays loading of the whole form, and not just the payment component, while we are navigating to Nets or other payment providers, and while we are navigating to the next process step after a payment is complete.

## Related Issue(s)

- closes #2740 

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels


bjosttveit:
> The payment provider seems to need to be placed within the PDFWrapper in order to be able to call next() from useProcessNavigation.

Could it be added to the `FormProvider` here instead? It should only require the `ProcessNavigationProvider`, not the `PDFWrapper`.

```tsx
<OrderDetailsProvider>
  {hasProcess ? (
    <ProcessNavigationProvider>
      <PaymentProvider>
        <Provider value={undefined}>{children}</Provider>
      </PaymentProvider>
    </ProcessNavigationProvider>
  ) : (
    <Provider value={undefined}>{children}</Provider>
  )}
</OrderDetailsProvider>

```

Magnusrm:
> Could it be added to the FormProvider here instead? It should only require the ProcessNavigationProvider, not the PDFWrapper.

I did try this, but it did not work. The backend would get stuck and eventually fail with PDF generation failed. I did not manage to find out why 🤔

src/components/wrappers/ProcessWrapper.tsx
```diff
@@ -148,9 +149,11 @@ export const ProcessWrapper = () => {
             path='*'
             element={
               <PDFWrapper>
-                <PresentationComponent type={ProcessTaskType.Data}>
-                  <Form />
-                </PresentationComponent>
+                <PaymentProvider>
```
olemartinorg, line 152: This could probably be placed late in `<FormProvider>` instead, otherwise the Payment component will now crash the app if placed in:
- Stateless forms
- Custom receipts (!)
- Sub forms

None of these are probably likely use-cases, but it's better to validate that in the component than having the whole app crash because of it. FormProvider is 'the place' for all providers needed by form components, at least.

src/features/payment/PaymentProvider.tsx
```diff
@@ -0,0 +1,80 @@
...
+};
+
+type PaymentContextProvider = {
+  children: ReactNode;
+};
+
+export const PaymentContext = createContext<PaymentContextProps | undefined>(undefined);
+
+export const PaymentProvider: React.FC<PaymentContextProvider> = ({ children }) => {
+  const [loading, setLoading] = useState<boolean>(false);
```
olemartinorg, line 27: It's not clear to me why this has to be a separate state. Why not use `busy` directly, instead of setting `loading` when `busy && !paymentError`? Double state-keeping like this can lead to brittleness, as I've already seen in the `busy` state for that matter.. If this is loading should probably be a function of 'is the call to `process/next` running', which you can get from tanstack (see for example `useIsSaving()`).

src/features/payment/PaymentProvider.tsx
```diff
@@ -0,0 +1,80 @@
...
+    if (paymentCompleted && next && !busy && !nextCalled.current) {
+      nextCalled.current = true;
+      setLoading(true);
+      next({ action: 'confirm', nodeId: 'next-button' });
+    }
+  }, [paymentCompleted, next, busy]);
+
+  if (loading) {
+    return <Loader reason='Navigating to external payment solution' />;
+  }
```
olemartinorg, line 69: There doesn't seem to be a way to get away from this loader, but that might be intentional? :thinking:

src/features/payment/PaymentProvider.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+  return <PaymentContext.Provider value={contextValue}>{children}</PaymentContext.Provider>;
```
olemartinorg, line 71: I suspect this can lead to re-renders of all form components when anything in this provider changes. The `contextValue` object is at least not memoized, so this context will provide a new value every re-render. Have you checked if this affects anything when (for example) submitting a form (not necessarily a form with payment)? If useProcessNavigation() causes this to re-render because process navigation became busy, that might lead to expensive re-renders immediately when the submit button is pressed.

olemartinorg: Btw, these things are difficult to get right, which is why we made `createZustandContext()` and used that pattern in many of these global form-providers. That way you can set all the state you want, change it however you want, but never cause full-tree re-renders unless every component in the tree uses the state from zustand.

olemartinorg (approved):
Looks good! I like the direction this is taking, but I have a few concerns - see comments.

olemartinorg:
> I did try this, but it did not work. The backend would get stuck and eventually fail with PDF generation failed. I did not manage to find out why 🤔

@Magnusrm What happens if you just load the instance with `?pdf=1` in the URL? In my experience, the full PDF generator adds a lot of complexities that can fail, so the provider placement may not be the issue there.

src/features/payment/PaymentProvider.tsx
Magnusrm: We want to also display loading when we are navigating to the external payment provider. This is not a call to process next, but to "performAction". We also want to keep displaying a loading state after the call is complete, since the redirect seems to have some slowness in displaying the page for the payment provider that is redirected to.

src/features/payment/PaymentProvider.tsx
Magnusrm: I'll do some testing and see if i can make use of the pattern with `createZustandContext` 😊 👍

olemartinorg (approved):
Looks good! :raised_hands:

PR #2917, opened 2025-01-16
Synchronizing shared expression tests with backend

olemartinorg:
## Description

I compared the folders with those in `app-lib-dotnet`, and updating tests to match.

A few notes:
- Backend is missing function tests for functions not yet implemented there (`_experimentalSelectAndMap`, `displayValue`, `externalApi`, `formatDate`, `hasRole`, `linkToComponent`, `linkToPage`, `text` and `value`).
- The backend also gets to test duplicate component IDs here, but that check happens so early in frontend now that catching it and handling it in shared tests is difficult.
- I added tests for `argv`, and made sure to test that every expression function has a folder with shared tests.

## Related Issue(s)

- https://github.com/Altinn/app-lib-dotnet/pull/1040

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels


src/features/expressions/shared-functions-exist.test.ts
```diff
@@ -0,0 +1,43 @@
...
+import { implementsDisplayData } from 'src/layout';
+import { getComponentConfigs } from 'src/layout/components.generated';
+
+describe('Shared function tests should exist', () => {
+  const sharedTests = getSharedTests('functions');
+
+  describe('DisplayValue tests', () => {
+    for (const [type, config] of Object.entries(getComponentConfigs())) {
+      if (implementsDisplayData(config.def)) {
+        it(`Component '${type}' should hava e matching test in functions/displayValue/type-${type}.json`, () => {
```
cammiida, line 12: Hehe
```suggestion
        it(`Component '${type}' should have a matching test in functions/displayValue/type-${type}.json`, () => {
```

src/features/expressions/shared-functions.test.tsx
```diff
@@ -183,6 +183,27 @@ describe('Expressions shared function tests', () => {
...
+          id: 'theCurrentExpression',
+          type: 'NavigationButtons',
+          ...({
+            // This makes sure that the expression is never evaluated, as it is not a valid property. All properties
+            // that can handle expressions will be evaluated in the hierarchy generation, but errors from there will
+            // not effect the actual test here. Still, DataModelsProvider will find the reference to any data models
+            // inside this expression, because it will simply traverse the entire layout looking for expressions.
+            notAnActualExpression: expression,
+            // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          } as any),
```
cammiida, line 203: Kan caste til typen det burde være i stedet? 🤷‍♀️ ev `object`?

src/features/expressions/shared-functions.test.tsx
```diff
@@ -183,6 +183,27 @@ describe('Expressions shared function tests', () => {
...
+        layouts = structuredClone(_layouts);
+        const firstPage = Object.values(layouts)[0];
+        firstPage?.data.layout.push({
+          id: 'theCurrentExpression',
+          type: 'NavigationButtons',
+          ...({
+            // This makes sure that the expression is never evaluated, as it is not a valid property. All properties
+            // that can handle expressions will be evaluated in the hierarchy generation, but errors from there will
+            // not effect the actual test here. Still, DataModelsProvider will find the reference to any data models
+            // inside this expression, because it will simply traverse the entire layout looking for expressions.
```
cammiida, line 200: Har lest denne kommentaren 10 ganger og forstår fortsatt ikke helt 😬 er det mulig å forenkle på noen måte? Kortere kan forøvrig ofte være enklere.

cammiida (approved):
Cool!

src/features/expressions/shared-functions.test.tsx
olemartinorg: Usikker på om kortere er enklere å forstå her, om denne kommentaren var uforståelig. Det forklarer jo hvorfor vi må gjøre dette her, og kobler på dypt inngående detaljer til flere ting (NodesProvider, DataModelsProvider osv).

Hva med denne?
> This makes sure that the expression is never evaluated, as it is not a valid property. All properties
that can handle expressions (like 'hidden') will be evaluated during hierarchy generation, but errors
from there (such as unknown extra properties like this one) will not cause test failures here (so doing
this is safe). DataModelsProvider however, will recursively look inside the layout and find anything
that resembles an expression and load the data model it refers to. In other words, this makes sure we
load any data models that are only references in the expression we're testing - not elsewhere in the
layout. For an example of a test that would fail without this, see 'dataModel-non-default-model.json'.
It has only a Paragraph component with no expressions in it, so without injecting the tested
expression into that layout, DataModelsProvider would not load the data model that the expression refers
to, and the test would fail.

src/features/expressions/shared-functions.test.tsx
olemartinorg: Den typen finnes ikke, det er litt av poenget :face_with_peeking_eye: Men jeg ser at `object` funker, så det får bort behovet for eslint-disable iallefall.

Issue #2027, opened 2024-04-16
Custom component not working in v4


SimenRokaas:
### Description of the bug

Custom component (address search) worked in v3, does not work anymore in v4. We did not change anything.

Example of working v3 form: https://altinn.studio/repos/dsb/siv-1002-personopplysninger/src/branch/master/App/ui/layouts/02PersonalInformation.json#L303

Example of non-working v4 form: https://altinn.studio/repos/dsb/int-1017-personopplysninger/src/branch/master/App/ui/form/layouts/02PersonalInformation.json#L302 (commented out)

### Steps To Reproduce

1. Clone app int-1017-personopplysninger
2. Remove /* */ around custom component so it gets included (see last description link)
3. Start form, go to page 2
4. Page fails with reference to custom component rendering

adamhaeger:
Was able to get the component to render by renaming the function L in the `dsb-components.js`.

This suggests there is a problem with the way the svelte code is compiled or minified, so you should look into this.

However, when clicking an item in the address list, I found an issue in our component related to how we save data so that has been fixed in this PR:

https://github.com/Altinn/app-frontend-react/pull/2920

PR #2920, opened 2025-01-17
Issues in Custom component

adamhaeger:
Allowing not including field when using simple binding, fixed bug when generating passThroughProps

## Description

- According to docs you should not have to pass a field in the custom event when using simplebinding, this is now fixed and giving you an error if you are not using simpleBinding _and_ not passing field in the event.
- Fixed an issue where you get a JSON circular dependency error when generating passThroughProps because we tried to include the div ref.

## Related Issue(s)

- closes https://github.com/Altinn/app-frontend-react/issues/2027

## Verification/QA

  - [x] Creator of the original issue (or service owner) has been contacted for manual testing (or will be contacted when released in alpha)
- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels

### DIFF
--- a/src/layout/Custom/CustomWebComponent.tsx
+++ b/src/layout/Custom/CustomWebComponent.tsx
@@ -15,7 +15,10 @@ export type ICustomComponentProps = PropsFromGenericComponent<'Custom'> & {
   summaryMode?: boolean;
 };
 
-export type IPassedOnProps = Omit<PropsFromGenericComponent<'Custom'>, 'node' | 'componentValidations'> &
+export type IPassedOnProps = Omit<
+  PropsFromGenericComponent<'Custom'>,
+  'node' | 'componentValidations' | 'containerDivRef'
+> &
   Omit<CompInternal<'Custom'>, 'tagName' | 'textResourceBindings'> & {
     [key: string]: string | number | boolean | object | null | undefined;
     text: string | undefined;
@@ -31,8 +34,11 @@ export function CustomWebComponent({
   const langTools = useLanguage();
   const { language, langAsString } = langTools;
   const { tagName, textResourceBindings, dataModelBindings, ...passThroughPropsFromNode } = useNodeItem(node);
+
+  const { containerDivRef: _unused, ...restFromGeneric } = passThroughPropsFromGenericComponent;
+
   const passThroughProps: IPassedOnProps = {
-    ...passThroughPropsFromGenericComponent,
+    ...restFromGeneric,
     ...passThroughPropsFromNode,
     text: langAsString(textResourceBindings?.title),
     getTextResourceAsString: (textResource: string) => langAsString(textResource),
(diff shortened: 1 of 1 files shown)


src/layout/Custom/CustomWebComponent.tsx
```diff
@@ -48,15 +49,23 @@ export function CustomWebComponent({
...
-        setValue(field, value);
+        if (!dataModelBindings?.simpleBinding && !field) {
+          throw new Error(
+            'If you are not using simpleBinding, you need to include a field in your change event to indicate what datamodel binding you want to save to. See docs: https://github.com/Altinn/altinn-studio/issues/8681',
+          );
+        }
+        if (field) {
+          setValue(field, value);
+        }
+        setValue('simpleBinding', value);
```
olemartinorg, line 60: This always sets `simpleBinding` anyway
```suggestion
        if (field) {
          setValue(field, value);
        } else {
          setValue('simpleBinding', value);
        }
```

src/layout/Custom/CustomWebComponent.tsx
```diff
@@ -88,11 +97,12 @@ export function CustomWebComponent({
       // eslint-disable-next-line @typescript-eslint/no-explicit-any
       prop = ReactDOMServer.renderToStaticMarkup(prop as any);
     } else if (['object', 'array'].includes(typeof prop)) {
-      prop = JSON.stringify(passThroughProps[key]);
+      if (key !== 'containerDivRef') {
+        prop = JSON.stringify(passThroughProps[key]);
+      }
```
olemartinorg, line 102: Eslint complained about this one, but i think you could either:
- Add `containerDivRef: _unused` to the function definition prop-destructor
- Rewrite it to be more explicit about which props we pass on to Custom components. I think this is wiser, as we don't necessarily want to expose future props here automatically. At the same time, if they're explicitly built it's easier for us to rewrite thing and notice when we break Custom components, and it's much easier to document what we pass to them when it's not relying on implicit behaviour.

olemartinorg (approved):
Looks promising! :raised_hands:

src/layout/Custom/CustomWebComponent.tsx
adamhaeger: Good catch :)

src/layout/Custom/CustomWebComponent.tsx
adamhaeger: yup, also good point :)

adamhaeger: As the 'docs' says _All other properties of a component are passed as HTML attributes (handled as strings) to the web component._ Im hestitant to pass props explicitly for now as we dont know what props people rely on. So Ive just removed the containerDivRef for now :)

olemartinorg: I guess that works. I've removed most values from `PropsFromGenericComponent` since the custom component support was introduced (there used to be a lot more props!), and I never got any hints from typescript indicating I removed props that broke an external contract with custom components. :-( Making them explicit would help both ways.

PR #2926, opened 2025-01-20
Feature/list component with checkboxes

walldenfilippa:
## Description
image
image

## Related Issue(s)

- closes #{issue number}

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels
  - [x] I don't have permissions to add labels, please help me out


src/layout/List/ListComponent.tsx
```diff
@@ -54,16 +57,23 @@ export const ListComponent = ({ node }: IListProps) => {
 
   const { data } = useDataListQuery(filter, dataListId, secure, mapping, queryParameters);
   const bindings = item.dataModelBindings ?? ({} as IDataModelBindingsForList);
-  const { formData, setValues } = useDataModelBindings(bindings);
+
+  const { formData, setValues } = useDataModelBindings(bindings, 1, 'raw');
```
olemartinorg, line 61: You should probably use `DEFAULT_DEBOUNCE_TIMEOUT` instead of passing `1` here - that will preserve the default value. 🙌

src/layout/List/ListComponent.tsx
```diff
@@ -75,46 +85,124 @@ export const ListComponent = ({ node }: IListProps) => {
     return JSON.stringify(selectedRow) === JSON.stringify(row);
   }
 
+  function isRowChecked(row: Row): boolean {
+    return (formData?.saveToList as Row[]).some((selectedRow) => {
+      const { altinnRowId, ...rest } = selectedRow;
```
olemartinorg, line 90:
```suggestion
      const { altinnRowId: _, ...rest } = selectedRow;
```

This gets rid of the eslint error 🥳

src/layout/List/ListComponent.tsx
```diff
@@ -75,46 +85,124 @@ export const ListComponent = ({ node }: IListProps) => {
...
+    }
+  };
+
+  const handleSelectedCheckboxRow = (row) => {
+    if (!item.dataModelBindings?.saveToList) {
+      return;
+    }
+    if (isRowChecked(row)) {
+      const index = (formData?.saveToList as Row[]).findIndex((selectedRow) => {
+        const { altinnRowId, ...rest } = selectedRow;
```
olemartinorg, line 113:
```suggestion
        const { altinnRowId: _, ...rest } = selectedRow;
```

Same here!

src/layout/List/ListComponent.tsx
```diff
@@ -75,46 +85,124 @@ export const ListComponent = ({ node }: IListProps) => {
...
   const title = item.textResourceBindings?.title;
   const description = item.textResourceBindings?.description;
+  const component = item.componentType;
+
+  const handleRowClick = (row) => {
+    if (item.componentType === 'CheckBoxes') {
+      handleSelectedCheckboxRow(row);
+    } else {
+      handleSelectedRadioRow({ selectedValue: row });
+    }
```
olemartinorg, line 104: I'm not entirely sold on the `componentType` property. My first thought was "is it likely this will be expanded to include any more component types in the future?" and I don't see how that can happen. It' either single-select og multi-select, so maybe it should just say that instead? (I.e. a `multiSelect: true/false`property).

But then, can't this also be detected by the precense of a `saveToList` binding? At least, setting `componentType: "CheckBoxes"` but not setting `dataModelBindings.saveToList` now just means your changes won't be saved (and you get a config error).
So I vote to remove the properly entirely, and we can just use `dataModelBindings.saveToList` to detect if checkboxes or radios should be displayed. That should make it less likely that app developers configure this wrong.

src/layout/List/ListComponent.tsx
```diff
@@ -75,46 +85,124 @@ export const ListComponent = ({ node }: IListProps) => {
     return JSON.stringify(selectedRow) === JSON.stringify(row);
   }
 
+  function isRowChecked(row: Row): boolean {
+    return (formData?.saveToList as Row[]).some((selectedRow) => {
+      const { altinnRowId, ...rest } = selectedRow;
+      return Object.keys(rest).every((key) => Object.hasOwn(row, key) && row[key] === rest[key]);
```
olemartinorg, line 91: This checks every key in in `rest`, not every key in `row`. That means that:
1. If you let users add info to the row later, it will not "detach" from the List component (no longer display as selected there)
2. If a key is in the List component data model bindings, but NOT in the repeating group row, that key will not be considered

I think inverting this logic to check keys in `row` (or, at least the keys that exist in `row`and `dataModelBindings`) should fix these issues.

olemartinorg (approved):
Sorry for the long delay in reviewing this! 🫣 Overall, this looks very good! Thank you!! 🫶 
There are a few things to consider when looking closely at the details (apart from the comments I wrote):

1. When you uncheck a checkbox, the repeating group row will be deleted. If the user had put any more info in the repeatin group row, that info is now lost - this is what we made `AlertOnDelete` to warn the user against. Not a hard requirement for now, but something we might consider in the future.
2. Again, if you allow users to add more info in the repeating group row (which is not part of your use-case, but will be part of other use-cases), unchecking and then re-checking a checkbox will wipe extra data from the repeating group row. I hope to implement a solution for that, for example in #1497.

(This is not something I expect you to fix, these are more like notes for myself for later!)

src/layout/List/ListComponent.tsx
walldenfilippa: Make sense, good idea!

src/layout/List/ListComponent.tsx
walldenfilippa: @olemartinorg were you thinking something like this? 
```
function isRowChecked(row: Row): boolean {
    return (formData?.saveToList as Row[]).some((selectedRow) =>
      Object.keys(row).every((key) => Object.hasOwn(selectedRow, key) && row[key] === selectedRow[key]),
    );
  }
```

olemartinorg: Yup! :raised_hands:

olemartinorg:
I merged this into `feat/list-checkboxes` instead of the `main` branch, so that I can take a look at the tests that were failing and create new ones for this functionality. Thanks for the contribution! :pray:

Issue #265, opened 2021-12-08
Support instantiation based on url params


lorang92:
## Description
We need the ability to start the form itnstansiation by providing a standard link including query parameters. The case for this is correspondance messages and regular emails containging a link to start the form. In Altinn 2 one of the most used features is to send a notification to citizens/businesses about the information they need to know (roughly 60 millions per year). Typically this is information about correspondence received or a new form they need to fill out. 

## Additional information
Example url from GitHub: https://github.com/Altinn/app-lib-dotnet/issues/new?assignees=&labels=kind%2Fchore%2Cstatus%2Fdraft&projects=&template=chore.yml
Documentation from GitHub: https://docs.github.com/en/issues/tracking-your-work-with-issues/creating-an-issue#creating-an-issue-from-a-url-query

## Acceptance criteria
- [ ] Instantiation can happen with a call containing url parameters which can be used for prefill for the new instance
- [ ] A hook which allows the application developers to read the parameteres in an easy way and to add custom logic **before** the instance is created. For example checking if a provided id is allready registered/used. The case for this is typically survey links sent to users which should not be shared between users. 
- [ ] If the instance shouldn't be created we should present information to the user that this link is no longer valid.
- [ ] The information page should be configurable.
- [ ] A way to map query parameters to the data model for Studio configurable prefil setup

FinnurO:
Digital gravferdsmelding

RonnyB71:
* Application developers should be in control of which query parameters should map to which data model fields
* Application developers should be able to define valid values to go into the data model fields
* How should we handle encoding of values to prevent security issues?
* Individual query params vs base64 encoded param => individual
* We need to fix redirect in Altinn Innboks otherwise you will end up in the inbox when you are not logged in when following the link
* Should we route this via an stateless form first instead of creating an instance directly? Yes
* Example url: https://app.serviceowner.altinn.no/org/app/layoutName/new?key1=value1&key2=value2
* Prefil config for data model connected to stateless layout should get url params as a new source.

olemartinorg:
Discussion about this on slack:
- https://altinn.slack.com/archives/C02EJ9HKQA3/p1724749995459069

FredM-NVE:
Denne er viktig for oss i NVE også, så bumper denne.

PR #2951, opened 2025-01-28
Feat/265 url param instantiation

adamhaeger:
## Description

Allows users to configure query params that will prefill the datamodel.

Depends on this backend PR: https://github.com/Altinn/app-lib-dotnet/pull/1068

Query params will prefill datamodel according to this flow diagram:

Team Apps - General - Query param prefill flyt (1)

## Related Issue(s)

- closes https://github.com/Altinn/app-frontend-react/issues/265

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels


src/features/datamodel/DataModelsProvider.tsx
```diff
@@ -295,12 +296,30 @@ function LoadInitialData({ dataType, overrideDataElement }: LoaderProps & { over
   const dataElements = useLaxInstanceDataElements(dataType);
   const dataElementId = overrideDataElement ?? getFirstDataElementId(dataElements, dataType);
   const url = useDataModelUrl({ dataType, dataElementId, includeRowIds: true });
+  const storedParams = sessionStorage.getItem('queryParams');
```
olemartinorg, line 299: Denne verdien burde slettes rett etter at den har blitt lest og brukt, for den gjelder jo bare for neste instans - ikke alle fremtidige instanser.

adamhaeger: Se oppdatering :)

src/features/datamodel/useBindingSchema.tsx
```diff
@@ -68,7 +70,15 @@ function getDataModelUrl({
   isAnonymous,
   isStateless,
   instanceId,
+  prefillFromQueryParams,
 }: DataModelDeps & DataModelProps) {
+  if (prefillFromQueryParams && isStateless && !isAnonymous && dataType) {
```
olemartinorg, line 75: I don't see why we should limit this not to work for anonymous stateless? :thinking:

src/utils/urls/appUrlHelper.ts
```diff
@@ -52,6 +52,12 @@ export const getFileTagUrl = (instanceId: string, dataGuid: string, tag: string
 
 export const getAnonymousStatelessDataModelUrl = (dataType: string, includeRowIds: boolean) =>
   `${appPath}/v1/data/anonymous?dataType=${dataType}&includeRowId=${includeRowIds.toString()}`;
+export const getStatelessDataModelUrlWithPrefill = (
+  dataType: string,
+  includeRowIds: boolean,
+  prefillFromQueryParams: string,
+) =>
+  `${appPath}/v1/data?dataType=${dataType}&includeRowId=${includeRowIds.toString()}&prefill=${prefillFromQueryParams}`;
```
olemartinorg, line 60: So you're passing prefill to the backend so that backend fills that into the model when first reading it? That's not quite what I expected here - I thought we talked about sending the first request as a POST with the prefill data in it, not sending a GET with prefill data for the backend to figure out. This works, I guess, but it adds complexity when we could have done this with existing APIs already (but then via POST).

olemartinorg (approved):

src/utils/urls/appUrlHelper.ts
adamhaeger: Uuuhm well wouldnt that require a new implementation instead of adding to what already happens?

Wouldnt that add even more complexity? 

Either way I think it makes sense to remove the GET /data endpoint altogether in this task: https://github.com/Altinn/app-frontend-react/issues/2065 as the GET and POST do more or less the same thing, so could merge this and fix it as part of that I guess.

adamhaeger: Or did you mean that we just do a setValue or something? We also have to let TE run validation code of the queryparam values so that would have to hook into it somehow also, see line 132 here: src/Altinn.App.Api/Controllers/StatelessDataController.cs

olemartinorg: Aha! Yep, I think it makes sense to never call the GET endpoint at all, and just start off by POSTing an empty object (or one with prefill) in stateless. Might want to double-check that with Ivar as well, but I think that would work.

adamhaeger: After a huddle we have decided it makes sense to switch to using the POST endpoint and deprecating the GET endpoint.

Issue #1299, opened 2023-06-30
More expression functions


olemartinorg:
### Description

- Date handling
  - `isDateBefore`
  - `isDateBeforeOrEq`
  - `isDateAfter`
  - `isDateAfterOrEq`
- String handling
  - `substr` or similar
  - `indexOf`/`strpos` or similar
  - `replace` or similar
  - `ucfirst` or similar
  - `lcfirst` or similar (our `smartLowerCaseFirst` would be cool to use, but maybe too complex to port to the backend)
- Data model lookups
  - `value`, just like `component` but looks up the value for _this_ component (and also returns it even if this component is hidden)
  - `relDataModel` (just a suggestion), a function to look up a relative path in the data model. We'd have to figure out a syntax for relative data model paths, which should be possible to use in future component data model bindings as well (i.e. binding a component in a group to `./FirstName` instead of `My.Deep.Group.FirstName`)
- Options
  - `optionLabel` - A function that returns the option label for a given option value (should probably only work inside a function with defined options, and probably only something that could work on the frontend). It could probably also work without any arguments, returning the current label (works for radio/dropdown/List, but not checkboxes/multipleselect, as they have multiple options).

### Additional Information

- #1176
- #1179
- #1208

olemartinorg:
Diskutert i statusmøte med SSB i dag. De ønsker sterkt replace-funksjonen.

rvessb:
> Diskutert i statusmøte med SSB i dag. De ønsker sterkt replace-funksjonen.
https://github.com/orgs/Altinn/projects/34/views/2?pane=issue&itemId=32409141

olemartinorg:
Out now! 🥳 Requires `8.6.0-preview.3` on the backend (if used in `required` or `hidden` props).

Docs for new functions here:
- https://docs.altinn.studio/nb/altinn-studio/reference/logic/expressions/#func-compare
- https://docs.altinn.studio/nb/altinn-studio/reference/logic/expressions/#func-stringSlice
- https://docs.altinn.studio/nb/altinn-studio/reference/logic/expressions/#func-stringIndexOf
- https://docs.altinn.studio/nb/altinn-studio/reference/logic/expressions/#func-stringReplace
- https://docs.altinn.studio/nb/altinn-studio/reference/logic/expressions/#func-lcFirst-ucFirst
- https://docs.altinn.studio/nb/altinn-studio/reference/logic/expressions/#func-countDataElements

PR #2961, opened 2025-01-31
New expression functions: `compare`, string functions and more

olemartinorg:
## Description

This implements the following new expression functions:
- `countDataElements` - counts the number of data elements in the current instance, given a type. Can for example count attachments or subform elements.
- `compare`
  - Including the operators `equals`, `greaterThan`, `greaterThanEq`, `lessThan` and `lessThanEq` - all of which had dedicated functions already
  - `isBefore`, `isBeforeEq`, `isAfter`, `isAfterEq` and `isSameDay` - these are all brand new, and relies on date parsing
- `stringIndexOf`, `stringReplace`, `stringSlice`, `lowerCaseFirst` and `upperCaseFirst`

Also, review comments from #2948 have also been solved here.

## Related Issue(s)

- https://github.com/Altinn/app-lib-dotnet/pull/1067
- #1299
- #2122

## Verification/QA

- Automated tests
  - [x] Has been added/updated
    - https://github.com/Altinn/altinn-studio-docs/pull/1983
  
- Support in Altinn Studio
  
- Sprint board
- Labels


schemas/json/layout/expression.schema.v1.json
```diff
@@ -463,6 +518,29 @@
...
+        { "const":  "stringSlice" },
+        { "$ref": "#/definitions/string" },
+        { "$ref": "#/definitions/number" },
+        { "$ref": "#/definitions/number" }
+      ],
+      "additionalItems": false
+    },
+    "func-stringIndexOf": {
+      "title": "String index of function",
+      "description": "This function returns the index of the first occurrence of a substring in a string",
```
bjosttveit, line 535: What happens if the substring does not occur? That should probably be mentioned

src/features/expressions/expression-functions.ts
```diff
@@ -570,7 +610,45 @@ export const ExprFunctionImplementations: { [K in Names]: Implementation<K> } =
...
+    if (length !== null && length !== undefined && length < 0) {
+      throw new ExprRuntimeError(this.expr, this.path, `Length cannot be negative`);
+    }
+
+    if (length === null || length === undefined) {
+      return string.substring(start);
+    }
+
+    return string.substring(start, start + length);
+  },
```
bjosttveit, line 643: Open question: Should this be called `stringSlice`?

There are different ways of slicing strings or arrays, and strings can be seen as an array of characters. The js function used here is `substring` but it is implemented in a way similar to the [legacy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring#the_difference_between_substring_and_substr) `substr` (where the second parameter is a `length` instead of `end`)

This definition (start, length) is intuitive and easy to understand, but is not as flexible as what I would consider a `slice` function, which would be like the [array slice](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice) function in js, which is also how array/string indexing works in languages like python. The benefit of these is that they allow negative indices.

Say you want everything except the last five characters, but the total length of the string is dynamic and therefore not known at "config time". With a proper slice you could simply specify `slice(0,-5)`. Now, you can get the length of the string using the `stringLength` function, but we don't have any arithmetic operators to be able to subtract 5 from that value.

I am not saying we absolutely need this for strings right now, but when introducing list support, I would think that we would want to have a function that "slices" lists in a similar way to `Array.slice` in js (due to its versatility), and if that takes the same number of arguments but behaves completely different than `stringSlice` that would be confusing. So maybe we should consider renaming this to `substring` or something else that does not seize the term `slice`.

bjosttveit: Also:
If you wanted to combine (the current) `stringSlice` with the `stringIndexOf`-function this is much more cumbersome when you need to specify a `length` instead of an `end`.

olemartinorg: Great feedback! Thank you! :heart_hands: I agree completely - I'll do another pass to implement it this way.

schemas/json/layout/expression.schema.v1.json
olemartinorg: It's a bitt iffy, because now it returns `null` (not `-1`), and `stringSlice` breaks with a runtime error when you pass `null` to it. That's intentional, of course, and it especially useful when I add support for negative values in `stringSlice`, as that would just behave strangely if used together with `stringIndexOf` when it returns `-1`.

src/features/expressions/shared-tests/functions/stringSlice/together.json
```diff
@@ -0,0 +1,17 @@
...
+      0,
+      ["stringIndexOf", ["dataModel", "string"], "world"]
+    ],
+    "expressions!"
+  ],
+  "expects": "Hello, expressions!",
+  "dataModel": {
+    "string": "Hello, world!"
+  }
+}
```
bjosttveit, line 17: This is a good example for an alternate `stringSlice` implementation. This kind of thing works only because the start is `0`. If the data model string was something like `message: Hello, world!`

This:
```json
[
  "stringSlice",
  ["dataModel", "string"],
  ["stringIndexOf", ["dataModel", "string"], "Hello"],
  ["stringIndexOf", ["dataModel", "string"], "world"]
],
```
Would result in: `Hello, world!expressions!`

To get the same result you would need to subtract the first `stringIndexOf` from the second `stringIndexOf` to get the correct length.

bjosttveit (approved):

PR #2985, opened 2025-02-06
Allow instantiation retry after failed attempt

danielskovli:
## Description
Bug report from service desk: https://github.com/Altinn/altinn-support-private/issues/4002

Slack thread: https://digdir.slack.com/archives/C076GU2QHCG/p1738842092707849

## Process
This is a process that doesn't actually exist for us at the moment, but when this fix eventually gets _deployed_, we need to communicate this issue back to service desk/interested parties.

## Verification/QA
- Manual functionality testing
  - [x] I have tested these changes manually
  - [ ] Creator of the original issue (or service owner) has been contacted for manual testing (or will be contacted when released in alpha)
  - [ ] No testing done/necessary
- Automated tests
  - [ ] Unit test(s) have been added/updated
  - [ ] Cypress E2E test(s) have been added/updated
  - [ ] No automatic tests are needed here (no functional changes/additions)
  - [x] I want someone to help me make some tests
- UU/WCAG ([follow these guidelines](https://aksel.nav.no/god-praksis/artikler/utvikling?tema=universell-utforming) until we have our own)
  - [ ] I have tested with a screen reader/keyboard navigation/automated wcag validator
  - [ ] No testing done/necessary (no DOM/visual changes)
  - [ ] I want someone to help me perform accessibility testing
- User documentation @ [altinn-studio-docs](https://github.com/Altinn/altinn-studio-docs)
  - [ ] Has been added/updated
  
  - [ ] No functionality has been changed/added, so no documentation is needed
  - [ ] I will do that later/have created an issue
  
- Support in Altinn Studio
  - [ ] Issue(s) created for support in Studio
  
  - [ ] This change/feature does not require any changes to Altinn Studio
- Sprint board
  - [ ] The original issue (or this PR itself) has been added to the Team Apps project and to the current sprint board
  - [ ] I don't have permissions to do that, please help me out
- Labels
  - [x] I have added a `kind/*` label to this PR for proper release notes grouping
  - [ ] I don't have permissions to add labels, please help me out

PR #2999, opened 2025-02-11
Adding tests for re-instantiation, fixing rare eternal loading 

### LINKED ISSUE #2862: App gets stuck on instantiation after switching party
### Description of the bug

If you instantiate an app, then go back to party selection and select a different party, you will see an infinite loader with `reason='instantiating'`. The reason is the following check:
```ts
const shouldCreateInstance = !!party && !instantiation.lastResult && !instantiation.isLoading;
```
This check fails due to `instantiation.lastResult` being truthy, probably because we already created an instance with a different party before.

It may also be prevented by the implementation of the `instantiate` function:
```ts
if (instantiate.data || instantiate.isPending || instantiate.error || isInstantiatingRef.current) {
  return;
}
```

### Steps To Reproduce

1. Go to https://ttd.apps.tt02.altinn.no/ttd/frontend-test
2. Login with a testid with at least two parties
3. Select 1st party, and create a new instance
4. Navigate to `/#/party-selection` and select a different party
5. See infinite loading screen

### Alternative steps to reproduce

- https://digdir.slack.com/archives/C076GU2QHCG/p1738842092707849

### LINKED ISSUE #2997: Add automated tests for party selection fix
## Description

https://github.com/Altinn/app-frontend-react/pull/2985 patches the instantiation handler so that a user is allowed to instantiate as expected, after re-selecting a correct party if they failed on the first attempt.

This scenario should be covered by an automated test.

## Acceptance criteria
- [ ] Write test for scenario descibed above (and in PR)

olemartinorg:
## Description

Adding automated tests for re-instantiation. Also removing ref-states and trying to simplify how this works, and making sure to clear an ongoing/failed instantiation when the user leaves the page (such as navigating back to party selection manually) which previously made the app display the loading-spinner indefintely.

## Related Issue(s)

- closes #2862
- closes #2997
- tests functionality implemented in #2985

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels

### DIFF
--- a/src/features/instantiate/containers/InstantiateContainer.tsx
+++ b/src/features/instantiate/containers/InstantiateContainer.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect } from 'react';
 
 import { Loader } from 'src/core/loading/Loader';
 import { InstantiateValidationError } from 'src/features/instantiate/containers/InstantiateValidationError';
@@ -6,6 +6,7 @@ import { MissingRolesError } from 'src/features/instantiate/containers/MissingRo
 import { UnknownError } from 'src/features/instantiate/containers/UnknownError';
 import { useInstantiation } from 'src/features/instantiate/InstantiationContext';
 import { useCurrentParty } from 'src/features/party/PartiesProvider';
+import { useAsRef } from 'src/hooks/useAsRef';
 import { AltinnPalette } from 'src/theme/altinnAppTheme';
 import { changeBodyBackground } from 'src/utils/bodyStyling';
 import { isAxiosError } from 'src/utils/isAxiosError';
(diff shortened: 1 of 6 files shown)


src/features/instantiate/containers/InstantiateContainer.tsx
```diff
@@ -15,14 +16,21 @@ export const InstantiateContainer = () => {
...
     if (shouldCreateInstance) {
       instantiation.instantiate(undefined, party.partyId);
     }
   }, [instantiation, party]);
 
+  // Clear the instantiation when the component is unmounted, to allow users to start a new instance later
+  useEffect(() => {
+    const clear = clearRef.current;
+    return () => clear();
+  }, [clearRef]);
```
bjosttveit, line 32: Do you expect the ref value to change during the lifetime of this component? If so the cleanup function uses the ref value at the time the component mounts not the current ref value at the time it unmounts

bjosttveit (approved):

src/features/instantiate/containers/InstantiateContainer.tsx
olemartinorg: Yup, it might change, but that's just because `InstantiationContext` doesn't do any memoization. :face_with_peeking_eye: But even if it changes, it shouldn't really matter - the `setState()` call for `busyWithId` updates the same state, and the rest works directly on the queryClient. I had to write it this way because the ref itself may be gone when the cleanup function runs, and just writing a `useEffect()` with the `clear` function as a dependency would clean up everything as soon as anything happens (because of the missing memoization). :shrug:

PR #3004, opened 2025-02-12
`optionLabel` expression function

olemartinorg:
## Description

This adds the `optionLabel` function to expressions. This will look more complex than it needs to be, but there's a reason.

The signature for this is `["optionLabel", <optionsId>, <some-option-value>]`. That means we'll have to load that code list (given in the `optionsId` argument) from the backend, and since the expression engine is (not yet) async, it has to be ready when we resolve expressions. It would be much easier (for frontend) if we used a component id here instead, since we have the calculated options for all option-based components on-hand already. However, I feel it's far better to bear the complexity now and leave these things decoupled instead tying even more functionality into the need for having node/component state in a store.

To make this work, I introduced two new things:
- `CodeListsProvider` replaces `StaticOptionPrefetcher` and loads code lists from an `optionsId` (as used in `optionLabel` calls) into a zustand store for the expression function to select from.
- `LoadingRegistry` is a generic solution to a problem we've had for a while (originally an idea I should give credit to @ivarne for). Currently each provider handle their own loading state (and block the form with the `<Loading />` component until they finish), which causes all providers to work in series (not in parallel). With this approach, we could instead register every 'thing' that is loading something, and when they're all done loading what they need (even if that happens in parallel), only _then_ is the actual form displayed. To do this it needs two parts, the provider itself and the 'outlet' where it blocks rendering (which should be the last 'provider' to render in `FormProvider`, or before we render stuff out). 

The rest of the PR is fairly straight forward, although I had to re-arrange some things in `shared-functions.test.tsx` for those tests to work. 

## Related Issue(s)

- #1669

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels


bjosttveit (approved):
Nice!

Issue #2490, opened 2024-09-25
Support thousands of rows in RepeatingGroup


olemartinorg:
Source:
- https://altinn.slack.com/archives/C02EJ9HKQA3/p1727189717723409

After the hierarchy rewrite, although the work there meant to improve performance, loading thousands of rows is still not performing optimally. I would assume that fixing this properly requires hard architecture decisions in frontend. Making an issue for this here to have something in the backlog, but the broad architecture strokes are absolutely still a subject we're discussing.

olemartinorg:
Closing this, as it's been replaced by https://github.com/Altinn/altinn-studio/issues/16359

PR #3009, opened 2025-02-14
Removing most node traversal usages

olemartinorg:
## Description

My plan has, for a while, been to remove the node traversal concept entirely. When that 's done, we can also remove the `LayoutPages`, `LayoutPage` and even the `LayoutNode` object. At that point we can simplify the node hierarchy and refactor to remove the expression evaluation from the node generator.
This is just one step along that road, removing some of the node traversal usages.

## Related Issue(s)

- #2490
- #2460

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -39,7 +39,7 @@ export const App = () => (
     />
 
     <Route
-      path='/instance/:partyId/:instanceGuid/*'
+      path='/instance/:instanceOwnerPartyId/:instanceGuid/*'
       element={
         <InstanceProvider>
           <Outlet />
--- a/src/components/PDFGeneratorPreview/PDFGeneratorPreview.tsx
+++ b/src/components/PDFGeneratorPreview/PDFGeneratorPreview.tsx
@@ -5,7 +5,7 @@ import { FilePdfIcon } from '@navikt/aksel-icons';
 
 import { Button } from 'src/app-components/Button/Button';
 import classes from 'src/features/devtools/components/PDFPreviewButton/PDFPreview.module.css';
-import { useLaxInstance } from 'src/features/instance/InstanceContext';
+import { useLaxInstanceId } from 'src/features/instance/InstanceContext';
 import { useTaskTypeFromBackend } from 'src/features/instance/ProcessContext';
 import { Lang } from 'src/features/language/Lang';
 import { useCurrentLanguage } from 'src/features/language/LanguageProvider';
@@ -28,7 +28,7 @@ export function PDFGeneratorPreview({
   const [errorText, setErrorText] = React.useState<string | null>(null);
 
   const taskType = useTaskTypeFromBackend();
-  const instanceId = useLaxInstance((state) => state.instanceId);
+  const instanceId = useLaxInstanceId();
   const language = useCurrentLanguage();
 
   const disabled = taskType !== ProcessTaskType.Data || !instanceId || !isLocalOrStaging();
--- a/src/components/form/Form.tsx
+++ b/src/components/form/Form.tsx
@@ -155,12 +155,12 @@ export function FormFirstPage() {
  */
 function useRedirectToStoredPage() {
   const pageKey = useCurrentView();
-  const partyId = useNavigationParam('partyId');
+  const instanceOwnerPartyId = useNavigationParam('instanceOwnerPartyId');
   const instanceGuid = useNavigationParam('instanceGuid');
   const { isValidPageId, navigateToPage } = useNavigatePage();
   const applicationMetadataId = useApplicationMetadata()?.id;
 
-  const instanceId = `${partyId}/${instanceGuid}`;
+  const instanceId = `${instanceOwnerPartyId}/${instanceGuid}`;
   const currentViewCacheKey = instanceId || applicationMetadataId;
 
   useEffect(() => {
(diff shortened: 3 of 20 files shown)


src/features/devtools/components/ExpressionPlayground/ExpressionPlayground.tsx
```diff
@@ -76,8 +76,11 @@ export const ExpressionPlayground = () => {
 
   const traversalSelector = useNodeTraversalSelector();
 
-  const componentOptions = useNodeTraversal((t) =>
-    t.allNodes().map((n) => ({ label: n.id, value: `${n.page.pageKey}|${n.id}` })),
+  const componentOptions = NodesInternal.useShallowSelector((state) =>
+    Object.values(state.nodeData).map((nodeData) => ({
+      label: nodeData.layout.id,
+      value: `${nodeData.pageKey}|${nodeData.layout.id}`,
+    })),
```
bjosttveit, line 83: A shallow selector will not do anything in this case no? You are returning a list, so the elements of the list are shallow-compared for equality, but you are creating new objects to populate the list every time so it will always fail the equality check on the first element

olemartinorg: Absolutely true! Sloppy mistake on my part. :face_with_peeking_eye:

src/features/options/effects/EffectPreselectedOptionIndex.tsx
```diff
@@ -30,12 +32,17 @@ export function EffectPreselectedOptionIndex({ preselectedOption, valueType, opt
   const shouldSelectOptionAutomatically =
     !hasValue && !hasSelectedInitial.current && preselectedOption !== undefined && isNodeHidden !== true;
 
-  NodesInternal.useEffectWhenReady(() => {
+  useEffect(() => {
     if (shouldSelectOptionAutomatically) {
+      if (nodeStore.getState().readiness !== NodesReadiness.Ready) {
+        requestAnimationFrame(() => setForceUpdate((prev) => prev + 1));
+        return;
+      }
```
bjosttveit, line 40: That seems like a quite the machine gun of re-renders 🤯  The original `useEffectWhenReady` was a bit complex maybe, what was the issue here?

src/utils/layout/NodesContext.tsx
```diff
@@ -838,48 +838,53 @@ function isHiddenPage(state: NodesContext, page: LayoutPage | string | undefined
...
+    const childHidden = parentDef.isChildHidden(parent as any, id);
     if (childHidden) {
       return true;
     }
   }
 
-  return isHidden(state, parent, nodes, options);
+  if (parent) {
+    return isHidden(state, 'node', parent?.layout.id, options);
+  }
```
bjosttveit, line 885: Nitpick, parent cannot be null 😅 

```suggestion
  if (parent) {
    return isHidden(state, 'node', parent.layout.id, options);
  }
```

test/e2e/integration/frontend-test/pdf.ts
```diff
@@ -308,6 +308,8 @@ describe('PDF', () => {
     });
 
     cy.goto('changename');
+    cy.get(appFrontend.changeOfName.newFirstName).should('be.visible');
+    cy.waitUntilSaved();
```
bjosttveit, line 312: Was this needed for the test to pass? 🤯  `testPdf` also waits for save before doing anything

bjosttveit (approved):
Looks good, just curious about the implications of the `useEffectWhenReady`-replacement in the three `Effect*`-components

test/e2e/integration/frontend-test/pdf.ts
olemartinorg: Yes, it does - but it doesn't wait for the page to load. So when you `cy.goto(...)` it will start loading, and a `cy.waitUntilSaved()` right after that will just check to make sure the `body` element does _not_ have `[data-unsaved-changes=true]` - which of course it doesn't, the page is still loading. The problem here is that `changename` has a few `preselectedOptionIndex` properties set, and will save 1-2 times before the page is ready. I had trouble with this test crashing because it tried clearing timers in `FormDataEffects` when `cy.clock()` tried to freeze them. :thinking:

When I think more about it, this could possibly also be solved by waiting for saves before displaying `ReadyToPrint`, but on the other hand this is only a case in tests - if the real PDF engine tries to save data, it will crash hard when the data elements are locked.

src/features/options/effects/EffectPreselectedOptionIndex.tsx
bjosttveit: When cy.clock is called in testPdf? Should any of these effects be running in pdf mode?

olemartinorg: They didn't get a chance to run before `?pdf=1` got added, so yes, they do. We don't have any checks to make sure we don't try to write to a datamodel when in PDF mode. For a while now the PDF generation would just end up crashing instead. I don't think it's better to introduce those checks either, because we'd just be masking configuration mistakes in the apps - and we'd have to remember to add that check as early as possible everywhere the data model can be changed (which is pretty much all over the place).

olemartinorg: I reverted these changes, ran the pdf test (which I believe is the one that failed) in a loop. After 10+ iterations it hasn't failed, so I would think it's all OK now. :raised_hands:

PR #2487, opened 2024-09-24
scaling down font size in pdf by overriding base font-size

adamhaeger:
## Description

Before:
Screenshot 2024-09-24 at 14 39 42

After:

Screenshot 2024-09-24 at 14 39 06

## Related Issue(s)

- closes #2401 

## Verification/QA

- Manual functionality testing
  - [x] I have tested these changes manually
  - [x] Creator of the original issue (or service owner) has been contacted for manual testing (or will be contacted when released in alpha)
  - [ ] No testing done/necessary
- Automated tests
  - [ ] Unit test(s) have been added/updated
  - [ ] Cypress E2E test(s) have been added/updated
  - [ ] No automatic tests are needed here (no functional changes/additions)
  - [ ] I want someone to help me make some tests
- UU/WCAG ([follow these guidelines](https://aksel.nav.no/god-praksis/artikler/utvikling?tema=universell-utforming) until we have our own)
  - [ ] I have tested with a screen reader/keyboard navigation/automated wcag validator
  - [ ] No testing done/necessary (no DOM/visual changes)
  - [ ] I want someone to help me perform accessibility testing
- User documentation @ [altinn-studio-docs](https://github.com/Altinn/altinn-studio-docs)
  - [ ] Has been added/updated
  
  - [ ] No functionality has been changed/added, so no documentation is needed
  - [ ] I will do that later/have created an issue
  
- Support in Altinn Studio
  - [ ] Issue(s) created for support in Studio
  
  - [ ] This change/feature does not require any changes to Altinn Studio
- Sprint board
  - [ ] The original issue (or this PR itself) has been added to the Team Apps project and to the current sprint board
  - [ ] I don't have permissions to do that, please help me out
- Labels
  - [ ] I have added a `kind/*` label to this PR for proper release notes grouping
  - [ ] I don't have permissions to add labels, please help me out

PR #3021, opened 2025-02-18
Fix map not centered in PDF

bjosttveit:
## Description

In the PDF, the map component has been rendering with its bounds off center for some time, and it appears that the problem was introduced in https://github.com/Altinn/app-frontend-react/pull/2487. This change scaled down the global font-size when using `@media print`, and since the map component's height was defined in `rem` this caused the height to get much smaller when the browser calls print, which results in the bottom part getting cut off like this (the pin should be roughly in the center of the map):

bilde

In general we cannot change the bounds of the map at this point since we cannot load new tiles or do anything else asynchronously in the `beforePrint` event, so we have to make sure that the bounds are fit correctly before the browser prints. Changing the size to use `px` instead restores the correct behavior.

This problem was not caught by percy until https://github.com/Altinn/app-frontend-react/pull/2965 where the `cy.clock()` call prevents the map from updating after we change the viewport which better matches the actual behavior of the PDF generator.

The changes in `custom.ts` are unrelated, I noticed that the viewport did not get restored correctly when returning to the form after calling `testPdf`.

## Related Issue(s)

- closes #{issue number}

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels

### DIFF
--- a/src/layout/Map/Map.tsx
+++ b/src/layout/Map/Map.tsx
@@ -119,7 +119,7 @@ export function Map({
   return (
     <MapContainer
       ref={map}
-      className={cn(classes.map, { [classes.mapReadOnly]: !isInteractive, [classes.printHack]: isPdf }, className)}
+      className={cn(classes.map, { [classes.mapReadOnly]: !isInteractive, [classes.print]: isPdf }, className)}
       center={center}
       zoom={zoom}
       bounds={bounds}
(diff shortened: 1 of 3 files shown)


olemartinorg (approved):
Sounds like a pain to debug! Well done! :clap: :clap: :clap:

PR #3066, opened 2025-03-03
Multiple minor bug-fixes, removing type union for layout-sets, compatibility in `all.test.tsx`

olemartinorg:
## Description

I needed to run the unit test we use to evaluate all hierarchies in all apps, and it was in need of a touch-up to make it work well again.

Changes for the test-runner:
- Implemented proper support for subforms when running these tests
- Fixed many minor issues when running the tests, including:
  - Mocking a newer backend version to make sure `PDFPreviewButton` does not fail
  - Allowing data models to have a filename on disk that is differently caSeD from the data type ID in the application metadata file (sketchy on linux, but works on windows).
- Marking a few more errors as non-critical

Other changes that might/will affect apps and config:
- Fixed an issue where `PDFPreviewButtonRenderLayoutValidator` would set state during the render call.
- Removed quirks for two apps where these quirks now had been fixed.
- When `FileUpload` components are bound to the data model with a `list` binding, we'd automatically set the list to an empty array when starting the form. I believe this was a mistake, and probably won't work as intended, because the data model (when stored as an XML file) doesn't really see the difference between an empty array and a `null`. The code we had just leads to extra PATCH calls, with no actual difference saved to the data model.
- Input components could end up passing the `textonly` property down to native elements, which would print errors.
- Comparing existing values when setting data in `DataModelsProvider`. I found a case where this would lead to an infinite loop, setting data over and over again, but it was probably limited to these tests. Still, setting data should not re-render everything when there are only object reference differences, not actual data differences.
- The most breaking one, at last: We had a type union for subforms, but when looking closely at the logic that tried to differentiate between them (and display a validation error), the code could never reach that path. As a result, it has always been valid to just use a regular layout-set configuration for subforms as well, and it turns out only our test-apps followed the docs to set up a layout-set with `"type": "subform"`. :facepalm: I'll open a PR for docs as well, as that is also misleading app developers now.

## Related Issue(s)

- closes #{issue number}

## Verification/QA

- Automated tests
  - [x] Has been added/updated
    - https://github.com/Altinn/altinn-studio-docs/pull/2055
  
- Support in Altinn Studio
  
- Sprint board
- Labels


bjosttveit (approved):
Very nice

PR #3069, opened 2025-03-04
Subform table expressions and navigation

### LINKED ISSUE #2475: Subform: cellContent display value based on optionList
## Context
We have `1-n` subforms that have a option list with text resources attached. We need the `cellContent` to display the text that is connected via the option list text resource value. 

### Current behaviour
Under is an example of todays behaviour. `Rapport-RR0007U.aarsregnskap.type.value` is connected to a option list in the subform
#### A table column in subform
```json
          {
            "headerContent": "Test headercontent 2",
            "cellContent": {
              "query": "Rapport-RR0007U.aarsregnskap.type.value",
              "default": "default tekst da lookup failed"
            }
          }
```
#### Option list
```json
[
  {
    "value": "S",
    "label": "Regnskapstype-Selskapsregnskap.label"
  },
  {
    "value": "K",
    "label": "Regnskapstype-Konsernregnskap.label"
  }
]
```

#### Result
image

## Proposed solution 1
Specify option list to render text from 
```json
    "cellContent": {
    "query": {
           "dataModelBinding": "Rapport-RR0007U.aarsregnskap.type.value",
           "optionsId": "option_list_id"           
         },
     "default": "default tekst da lookup failed"
      }
 ```

## Proposed solution 2 (preferred)
Dynamic expressions in context of the subform datamodel. `Rapport-RR0007U.aarsregnskap.type.value` is the value in the subform datamodel. 
```json
    "cellContent": {
    "text":  [
           "if", [
               "equals",
               ["Rapport-RR0007U.aarsregnskap.type.value", "S"],
               "Regnskapstype-Selskapsregnskap.label"
           ],
          "else", "Regnskapstype-Konsernregnskap.label"
       ],
     "default": "default tekst da lookup failed"
      }
 ```
 
 To target other datamodels (not the subform datamodel) you could still do something like this `["dataModel", "Path.Inside.DataModel", "someDataType"]`
 
 

### Additional Information

_No response_

### PR DESCRIPTION
**Pull request by bjosttveit**
## Description

Test app: https://ttd.apps.tt02.altinn.no/ttd/navigation-test-subform/

This adds expression support to Subform `tableColumns.cellContent` as well as a new `entryDisplayName` that are evaluated for each subform data element. The `entryDisplayName` gets used in summary as well as in the new navigation. There you now get a list of data elements for each subform component under its page with direct links to open it.

In this example the car brand is an option label, shown using the `optionLabel` expression, and the brackets around the reg.number are added using a `concat`-expression. The displayName in the navigation concatenates things:
bilde

This also includes some other changes and bugfixes for the navigation in general:
- Increased font size to `16px`
- Don't show navigation on instance-selection (or other weird places)
- Fix behavior in stateless when `taskNavigation` exists
- Fix navigation between subform and summary
- Fix `undefined` label-id
  - Using `node.id` instead of `node.item.id`. The type for `overrideItem` explicitly forbids overriding the id anyway so there is nothing wrong with getting it directly from the node object instead of the item.
- Processing could get stuck if component unmounting causes callback to never finish.
  - `useWaitForState` has this behavior, maybe it should reject all of the waiting promises when unmounting? That seemed to do the trick, but caused random uncaught exceptions here and there, and i didn't bother investigating further and dealt with it in the processing hook instead.

## Related Issue(s)

- closes #2475
- https://github.com/Altinn/app-frontend-react/pull/2831

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels

### CODE REVIEW

File: src/features/navigation/components/SubformsForPage.tsx [hunk 1]
```diff
@@ -0,0 +1,146 @@
...
+import { useNodeItem } from 'src/utils/layout/useNodeItem';
+import type { ExprVal, ExprValToActualOrExpr } from 'src/features/expressions/types';
+import type { IData } from 'src/types/shared';
+
+export function SubformsForPage({ pageKey }: { pageKey: string }) {
+  const subformIds = NodesInternal.useLaxMemoSelector(({ nodeData }) =>
+    Object.values(nodeData)
+      .filter((node) => node.pageKey === pageKey && node.layout.type === 'Subform')
+      .map((node) => node.layout.id),
+  );
```
olemartinorg, line 30: Could technically be done using `layoutLookups` now, without having a selector at all. :partying_face:

olemartinorg: But I guess it wouldn't properly handle any `Subform` components inside repeating groups, but I'm not sure there's a use-case for that anyway.

bjosttveit:
Failing cypress test seems to be due to a new version of stateless-app being published today. It also fails in `main`:
bilde

src/features/navigation/components/SubformsForPage.tsx
```diff
@@ -0,0 +1,146 @@
...
+}
+
+function SubformGroup({ nodeId }: { nodeId: string }) {
+  const [isOpen, setIsOpen] = useState(false);
+  const node = useNode(nodeId);
+  if (!node?.isType('Subform')) {
+    // This should never happen, @see SubformsForPage
+    throw new Error(`Navigation expected component: "${nodeId}" to exist and be of type: "Subform"`);
+  }
+  const subformIdsWithError = useComponentValidationsForNode(node).find(isSubformValidation)?.subformDataElementIds;
```
olemartinorg, line 51: Is there always just one validation with a list of ids? :thinking: I haven't looked into this deeply, but I would assume there could be other validations skipped here because you're using `.find()` instead of `.filter()`.

test/e2e/integration/navigation-test/navigation.ts
```diff
@@ -116,21 +116,21 @@ describe('navigation', () => {
...
 
       isUsingDialog && cy.showNavGroups();
-      cy.navGroup('Utfylling', 'Biler').find(ICON_COMPLETE).should('not.exist');
-      cy.navGroup('Utfylling', 'Biler').find(ICON_ERROR).should('not.exist');
+      cy.navGroup('Utfylling', 'Kjøretøy').find(ICON_COMPLETE).should('not.exist');
+      cy.navGroup('Utfylling', 'Kjøretøy').find(ICON_ERROR).should('not.exist');
       isUsingDialog && cy.hideNavGroups();
 
-      cy.findByRole('textbox', { name: /Fødselsdag/ }).type('01011983');
+      cy.findByRole('textbox', { name: /Fødselsdag/ }).type('09061934');
```
olemartinorg, line 133: Nice :sunglasses: (I had to look that up! :sweat_smile:)

olemartinorg (approved):
Nice work!! :rocket:

src/features/navigation/components/SubformsForPage.tsx
bjosttveit: I don't think you can use subform inside of repeating groups currently? We don't have any datamodel bindings, only a dataType, and it fetches all data elements for that type, its not like attachments that can map elementIds to a list

src/features/navigation/components/SubformsForPage.tsx
bjosttveit: Yeah its just one, and it says something like "There are errors in one or more subform entries", and the `subformDataElementIds` allow us to highlight which rows contain errors. I think it was done this way to not spam lots of validation errors that would look pretty much the same for each row. Though, now that we have `entryDisplayName`, we could show a unique error message per row, but previously they would either look exactly the same or include the uuid, which was not desirable.

test/e2e/integration/navigation-test/navigation.ts
bjosttveit: I had to lie about Scrooge McDuck's birthday unfortunately, since I had set up a minimum date of `01.01.1900` 😆

src/features/navigation/components/SubformsForPage.tsx
olemartinorg: Yep, true. Only a matter of time until that feature-request! :sweat_smile: Nothing to worry about for now.

PR #3077, opened 2025-03-04
Removing node traversal, fixing bugs in component lookups

olemartinorg:
## Description

This PR finally removes node traversal. However, to make component lookups work as before, I had to introduce some new functionality to make it possible to look up components:

- `LayoutContext` now gives you `LayoutLookups`, generated by the code in `makeLayoutLookups.ts`. This may look like a hierarchy, but it doesn't include information about how many rows there are in repeating groups - it's a fully static mapping that can be generated once for an entire layout, and makes it easy and performant to find out relations between components (such as `componentToParent[component.id]` will give you which component or page is the parent). It is also possible to look up which components bind to a given data model path, which page is a component on, what are the top-level components on that page, and more. To make this work, a tiny part of the node hierarchy code is re-used, specifically calling `def.claimChildren()` on every node type.
- `makeIndexedId()` was added as a function to let you find the indexed component ID matching your current position in the data model. So, if the current position in the first row in a repeating group and `currentValue` is a child of that repeating group, looking up `currentValue` will produce `currentValue-0` (the first instance of that component in the repeating group). This, and indexed ids, are still needed because the `component` and `displayValue` lookup functions need to find out if the target component is hidden - and as long as we still need to support `ruleHandler.js` as a dynamics engine, it's difficult to get away from these unique indexed component ids.
-  `DataModelLocationProvider` was added, partially for forwards-compatibility, so that we can provide the current location in the data model for expressions that run for components 'inside' that provider. This provider is used to tell the expression engine that 'you're now in row 0 in this repeating structure', and allows it to find the correct component instances when looking up components, and looking up in the data model.

In addition to the new stuff, this also adds shared tests for, and fixes a few issues the older code had:

1.  When outside of a repeating group, if you asked the former node traversal code to find a component _inside_ a repeating group, it would always find the first instance of that component (so, in row 0). Now it will just refuse to find any component - when we don't know which row you're looking for, that's the better approach (and it's what the backend does). Testing all currently deployed apps found that this doesn't break anything - the only thing it broke was a couple of our cypress tests that relied on this functionality.
2. Much worse, when you have a group (or other non-repeating child-having component) inside a repeating group, and you look up a component inside that group from _outside the group_ (but from inside the repeating group), lookups would fail. Because of the behavior described in the first point, we'd still find a component, but it would always be the first one (not the one in the row you're in right now). Testing currently deployed apps found that this affects one app in prod (`krt-1008a-1`), and two more in tt02 (`krt-1157a-1` and `ttd/signering-brukerstyrt`). All of which I've investigated manually, and found that this new functionality fixes the issues these apps had.

## Related Issue(s)

- closes #{issue number}

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels


bjosttveit (approved):
🙌

PR #3094, opened 2025-03-06
Detecting and crashing upon runaway save requests

olemartinorg:
## Description

I've seen this happen a few times. Mostly when we have bugs in our code, but I also saw an app that was configured to save options to a path in the datamodel where the backend would overwrite the data with invalid options, and we ended up in a loop where frontend tried to remove the invalid data and backend tried to add it back. Regardless of the reason, and who to blame, I think it's better to crash the app in these cases instead of spamming requests in a loop - that's something the user might not even notice.

Alternatively, consider this an effort to prevent self-DoS. :wink:  

## Related Issue(s)

- closes #{issue number}

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels

### DIFF
--- a/src/features/formData/FormDataWrite.tsx
+++ b/src/features/formData/FormDataWrite.tsx
@@ -118,6 +118,29 @@ function useFormDataSaveMutation() {
     }
   }
 
+  function checkForRunawaySaving() {
+    const now = Date.now();
+    const lastRequests = queryClient
+      .getMutationCache()
+      .findAll({ status: 'success', mutationKey: ['saveFormData'] })
+      .sort((a, b) => a.state.submittedAt - b.state.submittedAt)
+      .filter((req) => now - req.state.submittedAt < 10000) // Must have happened withing the last 10 seconds
+      .map((request) => request.state.data);
+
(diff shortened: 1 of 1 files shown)


bjosttveit (approved):
Nice, though I guess a user would be able to achieve 3 identical saves in 10 seconds if the app somehow reverts the users input in a radio-button or checkbox or something. Maybe 3 saves in 10 seconds is a bit to lax?

olemartinorg:
@bjosttveit I agree. Re-implemented this now with a 'last 10 requests' window, and made multiple improvements:
- This is now checking for variation in timing between requests, but we're no longer limited by the speed of the network, latency, or processing time on the server - the only thing that matters is that timing is somewhat consistent between saves.
- Even if you have inconsistencies in the network, since this is working on a time window, things will fail as soon as the network is consistent enough for 10 requests in a row.
- I imagine some cases may have an alternating pattern where every other response is equal to each other, so I'm detecting that as well.
- I tested this in multiple browsers, locally and on tt02, with simulated throttling (down to unbearable speeds), and it caught the problem every time (even when, and especially when, response times were long).

PR #3097, opened 2025-03-10
Fixing 2 sources of flakiness in Cypress tests

olemartinorg:
## Description

In an upcoming branch, I had trouble with some flaky cypress tests. These all seem timing-related, so I've quite possibly affected those timings in that branch - but tests shouldn't be failing for that reason alone.

1. In `summary.ts` the `cy.gotoAndComplete()` function was used, and right after that `injectExtraPageAndSetTriggers()` was called, which reloads the page. However, the test didn't wait until form data was saved first, so sometimes you could end up with a cancelled PATCH request because of that reload. As the last field to be edited in `cy.gotoAndComplete()` was the date picker field,  which is required, the required validation for that field would block navigation in `Navigation between summary and pages` and fail the test.
2. The new navigation component tests and the new `cy.showNavGroups()` function was based on checking the DOM and conditionally open/close the dialog if used. In my case this cypress function was called to open the dialog right before it was closed, causing the code to detect that the dialog was open and do nothing. When later assertions tried to read data from the dialog, they would fail. By rewriting this to more idiomatic cypress testing, we remove a source of flakiness. After this fix, cypress will see that it can't click on 'Skjemasider' because it is being covered by the dialog, and the automatic retrying will wait until the dialog disappears before trying to click the button again. 

## Related Issue(s)

- closes #{issue number}

## Verification/QA

  - [x ] No testing done/necessary
- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels


test/e2e/support/custom.ts
```diff
@@ -862,75 +862,54 @@ Cypress.Commands.add('getCurrentViewportSize', function () {
...
-    cy.get('[data-testid=page-navigation]').then((container) =>
-      cy
-        .findByRole('button', { name: groupName, container })
-        .parent()
-        .then((container) => cy.findByRole('button', { name: pageName, container })),
-    );
+    cy.get('[data-testid=page-navigation]')
+      .findByRole('button', { name: groupName })
+      .parent()
+      .findByRole('button', { name: pageName });
```
bjosttveit, line 881: `findByRole` does not get scoped by the subject its called on, so in this case only the last `.findByRole` really matters (except for asserting that a group with groupName also exists, but it is not necessarily the "parent" of pageName). If there are multiple pages with the same display name but in different groups this will not be able to distinguish the two. It will not distinguish between the navigation menu and other buttons with the same name either

test/e2e/support/custom.ts
```diff
@@ -862,75 +862,54 @@ Cypress.Commands.add('getCurrentViewportSize', function () {
...
-          cy.navGroup(groupName).click();
-        }
-      });
-      cy.navGroup(groupName).should('have.attr', 'aria-expanded', 'true');
-      cy.navGroup(groupName, pageName).click();
-      if (isUsingDialog) {
-        cy.findByRole('dialog', { name: 'Skjemasider' }).should('not.exist');
-      } else {
-        cy.navGroup(groupName, pageName).should('have.attr', 'aria-current', 'page');
+Cypress.Commands.add('gotoNavGroup', (isUsingDialog, groupName, pageName) => {
```
bjosttveit, line 887: It's only when detecting whether or not the dialog is currently open that caused problems right? So adding `isUsingDialog` as a prop here was not really needed? I just think it makes the command signature a bit messy, and I would like to have a nice utility command for navigating with the new navigation, similar to our `gotoNavPage` command for the old `NavigationBar` that gets used a lot in other tests

olemartinorg: >So adding `isUsingDialog` as a prop here was not really needed?

Maybe I'm not getting what you're saying, but it's needed as long as the behavior is supposed to be different depending on that prop. I guess we could try to read the current viewport size and infer if the dialog should be used from that? Or, which style/approach do you suggest?

bjosttveit: I was suggesting the original approach:
```ts
cy.get('body').then((body) => {
  const isUsingDialog = !!body.find('[data-testid=page-navigation-trigger]').length;
```
Since this should be static as long as the viewport is not changed during the test, as opposed to `isDialogOpen` where I can see that there is a problem.

test/e2e/support/custom.ts
olemartinorg: Hmm, I'll revert the changes I made to this function. To me it just seemed like wrapping in `.then()` was not really need (and I suspected it broke automatic retries in Cypress), but having run the test 5x successfully now in the problematic branch, I guess this was not really the problem. :pray:

test/e2e/support/custom.ts
olemartinorg: Duh, of course! Thanks! Sorry for being slow.. :sweat_smile:

bjosttveit (approved):
Very nice!

PR #3099, opened 2025-03-10
Making component validation plain hooks

olemartinorg:
## Description

As a step to simplify the framework code and make us independent of `NodeData.item` having resolved expression values, I'm moving component-level validation to use plain hooks (i.e. `useEmptyFieldValidation(node)` instead of `runEmptyFieldValidation(node, validationDataSources)`).

This has a few advantages:
- It should be slightly easier to reason about what every node needs in order to validate, and make changes to that code. No need to extend `ValidationDataSources`, just use a hook to select the data you need from a store.
- In places we now select from `NodesContext` to get parts of `NodeData`, we can probably just resolve the expressions we need values for in the future - it should be better and more performant to evaluate an expression twice than having to store the results of all expressions and evaluate all of them regardless of the need to actually read those values.
- This won't actually break the rule of hooks as long as components don't change types during runtime - and luckily they don't.

The changes in this PR, simply put:
- Adding a utility function `layoutLookups.getComponent(baseId, expectedType)` that lets us get `CompExternal<ExpectedType>`, i.e. the configuration from the layout file.
- Removing `ValidationDataSources`
- Moved component-level validation functions to hooks, and extracted them from the `index.tsx` files
- Rewrote `useNodeValidation()` to use the hooks instead

## Related Issue(s)

- #2490

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels

### DIFF
--- a/src/features/form/layout/LayoutsContext.tsx
+++ b/src/features/form/layout/LayoutsContext.tsx
@@ -16,12 +16,10 @@ import { useHasInstance } from 'src/features/instance/InstanceContext';
 import { useLaxProcessData } from 'src/features/instance/ProcessContext';
 import { useNavigationParam } from 'src/features/routing/AppRoutingContext';
 import type { QueryDefinition } from 'src/core/queries/usePrefetchQuery';
-import type { LayoutLookups } from 'src/features/form/layout/makeLayoutLookups';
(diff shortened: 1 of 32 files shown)


bjosttveit (approved):
### Looks good to me 🚀

PR #3101, opened 2025-03-11
Expressions: Using data sources as-needed, direct hooks for `displayValue`

olemartinorg:
## Description

A continuation in the refactoring effort to make the node hierarchy simpler. I have a longer-term goal of getting to a place where we can evaluate expressions on the fly when displaying the components themselves, and thus not having to evaluate them in the node hierarchy at all. This refactor gets us closes to that, by:

- Making `displayValue` just a simple hook (just like I did with component-level validation in #3099). We no longer have to share data sources between `DisplayDataProps` and expression data sources, instead we now iterate all the calls to `displayValue` functions and call hooks for getting display values for each of those target components immediately. This works as long as expressions don't change, which they only do in the expression playground (not in real layouts).
- `useExpressionDataSources()` now requires you to pass what you intend to evaluate, and will find which expression functions used inside. From that info, it can selectively call the hooks needed to construct just the number of data sources the expression needs to run. This way we never need to re-render when the language changes _unless_ you're calling an expression function that uses the current language, such as `language`. This also means that when every rendered component gets to run their own expressions in the future, they will only use the hooks that expression needs - so no huge duplication of all expression data sources everywhere.

## Related Issue(s)

- #2490
- #3099

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels


bjosttveit (approved):
Very cool indeed, I guess I have to make some changes on my end to support this with subform table, but now I can just throw an exception if an "illegal" expression source is used instead of polluting the expression-function logic 🙌

Issue #273, opened 2021-09-02
Checkbox component should support mapping to an array


lorang92:
## Description
The current implementation of the checkbox component maps answers to a comma-separated string. As an app developer, it feels more natural that these values are mapped as an array. We should support both cases 🦅 and make this setting configurable on the component.

## Considerations
- Default config? We have to stick to the current implementation as default as this could break existing apps. (or if we strongly belive that an array is better we should create a new major versjon.)
- Adding a checkbox-component from studio could potentially be setup as array mapping as default for all new boxes

Depends on #345 to be solved first.

## Acceptance criteria
- Checkbox component should support mapping data to an array structure
- Make this mapping configurable at the component
- Default mapping should be a comma-separated string

## Specification tasks

- [x] Development tasks are defined

## Development tasks
- [ ] Implement configurable setting in formlayout.json
  - [ ] update json-schema with new possible config values
- [ ] update checkbox-component to handle both cases 
- [ ] unit tests

## Test
> Add test cases here as checkboxes that are being tested as part of the changes.

## Definition of done
Verify that this issue meets [DoD](https://digdir-digitalisering.atlassian.net/wiki/spaces/T3KP/pages/5049246/Definition+of+Done) (Only for project members) before closing.

- [ ] Documentation is updated (if relevant)
  - [ ] Technical documentation (docs.altinn.studio)
  - [ ] User documentation (altinn.github.io/docs)
- [ ] QA
- [ ] Manual test is complete (if relevant)
- [ ] Automated test is implemented (if relevant)
- [ ] All tasks in this userstory are closed (i.e. remaining tasks are moved to other user stories or marked obsolete)

olemartinorg:
Psst! We now have backend support for scalar arrays, and the frontend includes support for the `list` binding for attachment references (see #246). I've noticed this issue myself, and it should be fairly easy to add support for the `list` binding here as well.

FinnurO:
I am sure I saw an issue where the need for checkbox mapping to datamodel fields was mentioned. It said it is a common pattern i Altinn 2.

PR #3136, opened 2025-03-19
Renaming new binding in `List` from `saveToList` to `group`

olemartinorg:
## Description

With this change, the data model binding in `List` matches that used for `object[]` types in `RepeatingGroup` (which also adds/removes rows). This should make the setting less confusing when we port this over to `Checkboxes` and `MultipleSelect` later, which hopefully should also get a `list` binding in the future (ref #273). Having both `list` and `saveToList` at that point would be confusing, so we should fix this before releasing #3007.

Also fixing a few minor issues I found:
- `useDisplayData()` for the `List` component now works with the `group` binding (it returns comma-separated strings, using `summaryBinding` when set).
- Data model binding validation didn't account for deeply nested properties in the data model, and also assumed the binding key was the same as the property name in the model (which is not always the case).
- Empty field validation didn't work properly with the `group` binding.

## Related Issue(s)

- #3007

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
    - https://github.com/Altinn/altinn-studio-docs/pull/2030
- Support in Altinn Studio
  
- Sprint board
- Labels

### DIFF
--- a/src/layout/List/ListComponent.tsx
+++ b/src/layout/List/ListComponent.tsx
@@ -80,14 +80,13 @@ export const ListComponent = ({ node }: IListProps) => {
   }
 
   function isRowSelected(row: Row): boolean {
+    if (groupBinding) {
+      const rows = (formData?.group as Row[] | undefined) ?? [];
+      return rows.some((selectedRow) =>
+        Object.keys(row).every((key) => Object.hasOwn(selectedRow, key) && row[key] === selectedRow[key]),
+      );
+    }
     return JSON.stringify(selectedRow) === JSON.stringify(row);
-  }
-
-  function isRowChecked(row: Row): boolean {
-    const rows = (formData?.group as Row[] | undefined) ?? [];
-    return rows.some((selectedRow) =>
-      Object.keys(row).every((key) => Object.hasOwn(selectedRow, key) && row[key] === selectedRow[key]),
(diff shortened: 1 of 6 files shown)


bjosttveit (commented):
I was about to write a comment on the original PR to ask if the binding should have been called `group` instead. I finally remembered why I didn't 😅  There is some special treatment of that name in app-lib here:

https://github.com/Altinn/app-lib-dotnet/blob/a79c86d6d72e41b9d54122d419564cdbafd9c42a/src/Altinn.App.Core/Internal/Expressions/LayoutEvaluator.cs#L66-L72

Not sure if that is a problem or not, but has this been taken into consideration?

ivarne:
I think it is just a lady way to check for groups that we need to handle somewhat differently

olemartinorg:
Oooh, yep, I agree that should be handled differently. In fact, `group` (and `saveToList` for that matter) was always a valid binding key for the `List` component - the keys didn't matter there.

> just a lady way to check for groups

:joy: I assume you meant lazy!

ivarne:
Yes, lazy. Can I blame it on autocorrect?

bjosttveit (approved):
We need new backend support for this anyway, so I guess it does not matter if its broken for unrelated reasons as well

PR #2871, opened 2025-01-06
Signee List Component (#2807)

cammiida:
## Related Issue(s)

- closes #{issue number}

## Verification/QA

- Manual functionality testing
  - [ ] I have tested these changes manually
  - [ ] Creator of the original issue (or service owner) has been contacted for manual testing (or will be contacted when released in alpha)
  - [ ] No testing done/necessary
- Automated tests
  - [ ] Unit test(s) have been added/updated
  - [ ] Cypress E2E test(s) have been added/updated
  - [ ] No automatic tests are needed here (no functional changes/additions)
  - [ ] I want someone to help me make some tests
- UU/WCAG ([follow these guidelines](https://aksel.nav.no/god-praksis/artikler/utvikling?tema=universell-utforming) until we have our own)
  - [ ] I have tested with a screen reader/keyboard navigation/automated wcag validator
  - [ ] No testing done/necessary (no DOM/visual changes)
  - [ ] I want someone to help me perform accessibility testing
- User documentation @ [altinn-studio-docs](https://github.com/Altinn/altinn-studio-docs)
  - [ ] Has been added/updated
  
  - [ ] No functionality has been changed/added, so no documentation is needed
  - [ ] I will do that later/have created an issue
  
- Support in Altinn Studio
  - [ ] Issue(s) created for support in Studio
  
  - [ ] This change/feature does not require any changes to Altinn Studio
- Sprint board
  - [ ] The original issue (or this PR itself) has been added to the Team Apps project and to the current sprint board
  - [ ] I don't have permissions to do that, please help me out
- Labels
  - [ ] I have added a `kind/*` label to this PR for proper release notes grouping
  - [ ] I don't have permissions to add labels, please help me out

PR #3232, opened 2025-04-03
add expression support for organisation instance owner name

HauklandJ:
## Related Issue(s)

- closes #{issue number}

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels


src/types/shared.ts
```diff
@@ -96,6 +96,7 @@ export interface IInstanceOwner {
   personNumber?: string;
   organisationNumber?: string | null;
   username?: string;
+  party?: IParty | null;
```
martinothamar, line 99: I'm assuming `IParty` here is modelled after the platform register model and not the new response/dto? Some fields will always be null but maybe that's good enough

HauklandJ: correct

src/utils/instanceDataSources.ts
```diff
@@ -12,10 +12,14 @@ export function buildInstanceDataSources(instance?: IInstance | null | undefined
         ? 'selfIdentified'
         : 'unknown';
 
+  // instanceOwnerName is set to 'unknown' if the instanceOwner is not an organisation
+  const instanceOwnerName = instance.instanceOwner.organisationNumber ? instance.instanceOwner?.party?.name : 'unknown';
+
```
bjosttveit, line 17: Why should it be set to `unknown`?

HauklandJ: I did not want to exposed more than necessary. The instanceContext expressions are limited "to ensure that information not (yet) exposed via a key here is not attempted to be retrieved from the instance."

Could expose it or have a different placeholder

HauklandJ: Open to change it, if you don't see any issues with exposing names for person instance owners

ivarne: As long as backend exposes it in the API, I don't really see the reason for this logic.

src/utils/instanceDataSources.ts
```diff
@@ -1,7 +1,7 @@
 import type { IInstance, IInstanceDataSources } from 'src/types/shared';
 
 export function buildInstanceDataSources(instance?: IInstance | null | undefined): IInstanceDataSources | null {
```
ivarne, line 3: The instance returned from PATCH, actions, process/next ... does not contain the instance owner party. Should we do something to ensure that an instance from the correct API is used?

PS: when we add `["instanceContext", "dataElementCount", dataType]`, or something similar, we need to use the most recent instance to resolve.

HauklandJ: We probably do not want to extend all the endpoints to include the instance owner party. 
For the instance owner party we do not need the most recent instance as it is not goingto change. 
Hmm.. I do not immediately see a solution that I like

HauklandJ: > The instance returned from PATCH, actions, process/next ... does not contain the instance owner party. 

Looking at the [instaceContext store](https://github.com/Altinn/app-frontend-react/blob/main/src/features/instance/InstanceContext.tsx) I thought so too, seeing how `mutateDataElement`/`changeData` etc calls `buildInstanceDataSources` with `next`. 

However, I set up an example app with `resource.nb.json`:

```
    {
      "id": "singing.summary.title.override",
      "value": "Disse har signert på vegne av {0}",
      "variables": [
        {
          "key": "instanceOwnerName",
          "dataSource": "instanceContext"
        }
      ]
    }
```
Which is used to override the title of a summary2 component in a custom pdf-layout `pdf.json`:

```
{
  "$schema": "https://altinncdn.no/schemas/json/layout/layout.schema.v1.json",
  "data": {
    "layout": [
      {
        "id": "signeeListSummary",
        "type": "Summary2",
        "target": {
          "type": "component",
          "id": "signee-list"
        },
        "overrides": [
          {
            "componentId": "signee-list",
            "title": "singing.summary.title.override"
          }
        ]
      }
    ]
  }
}
```

Which is used as `pdfLayoutName` in `Settings.json`:

```
{
  "$schema": "https://altinncdn.no/schemas/json/layout/layoutSettings.schema.v1.json",
  "pages": {
    "order": ["Side1"],
    "pdfLayoutName": "pdf"
  }
}
```

And this results in the instance owner name being used in the title.. (ignore timestamp)
image

There are plenty of PATCHes and a process next between the instance creation and the pdf generation, so why would this work if the hypothesis is correct?

ivarne: I'm not sure I have an hypothesis, or what it might be. I just pointed out that we currently have two different instance documents around, and your example seems to indicate that we currently use the right one (meaning we don't update), but it will be very confusing when we start adding more context keys that require us to use the latest instance document.

HauklandJ: [The only available keys](https://docs.altinn.studio/altinn-studio/reference/logic/expressions/#lookup-components-and-data) for `instanceContext` at the moment are `instanceId`, `instanceOwnerPartyId`, `instanceOwnerPartyType` and `appId`. None of these require an update, so that makes sense, right. If we need keys for the latest instance document we would have to rethink the approach.

ivarne: Yes, there is no bug yet. The question is if there is something easy we can do to prevent this from happening.

HauklandJ: I am open to suggestions

martinothamar: Made an issue for it: https://github.com/Altinn/app-lib-dotnet/issues/1236
I'm open to any solution. We could also update the frontend to have two instance models to match the new OpenAPI spec. If we were to move to OpenAPI spec driven codegen for models then this would probably not be an issue, and we would just realize that we needed to keep instance owner information separate, which IMO makes sense any way since it never changes

src/features/instance/InstanceContext.tsx
```diff
@@ -206,10 +213,16 @@ export const useLaxInstanceStatus = () => useLaxInstance((state) => state.data?.
...
-export const useLaxInstanceDataSources = () => useLaxInstance((state) => state.dataSources) ?? null;
 export const useLaxChangeInstance = (): ChangeInstanceData | undefined => useLaxInstance((state) => state.changeData);
 export const useHasInstance = () => useHasProvider();
 
+export function useLaxInstanceDataSources(): IInstanceDataSources | null {
+  const instance = useLaxInstanceData((i) => i);
+  const instanceOwnerParty = useInstanceOwnerParty();
+
+  return buildInstanceDataSources(instance, instanceOwnerParty);
+}
```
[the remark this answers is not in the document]
HauklandJ: Nice! Did this on the target branch of this pr: https://github.com/Altinn/app-frontend-react/pull/2871/commits/821cb60574a22891ea633cef77864c89e9443a1c

PR #3264, opened 2025-04-10
Sync with azure cdn directly

### LINKED ISSUE #3201: Stop using altinn-cdn git repo to push release assets (in addition Azure CDN)
### Description

Currently we are pushing release assets both directly to Azure CDN and to the altinn-cdn repo.
We have had a failure recently where during release we succeeded in pushing to Azure CDN, but failed to push the changes to the git repo, which failed the pipeline. Thus we have a release that practically succeeded (apps were updated, assets are available on CDN) but wasn't logged as a success (e.g. in Swarmia for DORA purposes).

We should consider stopping pushing assets to both CDN and the git repo
In that case we also need to make the rollback workflow work based Azure CDN instead of the git repo

### Additional Information

_No response_

### Tasks

- [ ] Refactor rollback workflow
- [ ] Refactor release script 

### Acceptance Criterias

* When releasing, the pipeline succeeds and fails atomically (the pipeline should never fail if assets have actually landed in CDN)

bjosttveit:
## Description

Remove everything related to syncing with Altinn-CDN repo:
- Removed checkout Altinn-CDN
- Removed commit to Altinn-CDN from release / revert scripts
- Removed push Altinn-CDN
- Revert script now works by directly calling `azcopy sync` between two virtual directories in azure
- Making the `index.json` file was previously done by listing the directories in the Altinn-CDN repo. I considered fetching the existing file and appending to it, but if there are race-conditions (multiple releases at the same time) things will not necessarily be eventually consistent. So I instead optet for moving it into a separate script / workflow step, and using `azcopy ls` to list the versions and generate the index file _after_ we are done with pushing the files. I tested generating the index file this way on my personal azure account.

## Related Issue(s)

- closes #3201

### DIFF
--- a/.github/workflows/revert.yml
+++ b/.github/workflows/revert.yml
@@ -33,13 +33,6 @@ jobs:
           sparse-checkout: |
             .github
 
-      - name: Checkout Altinn-CDN repository
-        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
-        with:
-          repository: 'Altinn/altinn-cdn'
-          token: ${{secrets.ALTINN_CDN_TOKEN}}
-          path: cdn
-
       - name: Azure login
         uses: azure/login@a457da9ea143d694b1b9c7c869ebb04ebe844ef5 # v2
         with:
@@ -48,6 +41,7 @@ jobs:
           subscription-id: ${{ secrets.AZURE_CDN_SUBSCRIPTION_ID_FC }}
 
       - name: Run roll back script
+        working-directory: app-frontend
         env:
           FD_DOMAIN: ${{ vars.AZURE_DOMAIN_FC }}
           FD_ENDPOINT: ${{ vars.AZURE_ENDPOINT_FC }}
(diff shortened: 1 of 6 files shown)


tjololo:
Pre reive comment, remember to check that you dont have resources outside og these paths:
- /toolkits/altinn-app-frontend/*
- /schemas/json/component/*
- /schemas/json/layout/*

Any other resources are served from resources synced from the altinn-cdn repo. [Relevant IaC code](https://github.com/Altinn/altinn-cdn-infra/blob/main/terraform/modules/frontdoor/frontdoor.tf#L63-L118)

bjosttveit:
We don't ever update `/schemas` since v4, because that would have breaking changes for v3 apps. So we only really push stuff to `/toolkits/altinn-app-frontend`, and the new schemas are generated and stored for each version

.github/workflows/revert.yml
```diff
@@ -56,15 +50,6 @@ jobs:
         run: |
           bash app-frontend/.github/scripts/revert.sh \
```
martinothamar, line 51: did you mean to remove  `app-frontend/` prefix here since you added working directory?

martinothamar (approved):

.github/workflows/revert.yml
bjosttveit: yes, good catch

PR #3287, opened 2025-04-24
Fix: '1.00' marked as invalid data in an Input field

### LINKED ISSUE #2368: Invalid data in numeric inputs: `1.00` should be converted to `1` instead of giving an error
### Description of the bug

Se discussion here: https://altinn.slack.com/archives/C02EJ9HKQA3/p1724761462469779

olemartinorg:
## Description

When combining number formatting with storing data in a `decimal` field in the data model, we have to convert the typed number to an actual number when storing it. To make sure we don't trick the user into thinking we can store numbers that cannot be stored, we _compare the typed number with the stringified number we stored in the data model_ and mark it as invalid if they aren't equal. This caused some unexpected friction for some use-cases:

- Typing `1` is fine
- Typing `1.` shows up as an error (if you stop typing at that point). We cannot store just the decimal point without any numbers after it.
- Typing `1.0` used to show an error (because we cannot tell a numeric type to store extra zeroes after the comma, this value is essentially equal to `1`). Now it stores the number (`1`) but temporarily stores local state in the input component for the string you're actually typing. When blurring the input field the value will revert to what we stored in the data model (`1`). 
- Typing `1.02` suddenly turn this into something we can store again, so the data model goes from storing `1` to `1.02`. This shows why it's important we show this value until the user blurs the field.

## Related Issue(s)

- closes #2368

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels

### DIFF
--- a/src/features/formData/FormDataWriteStateMachine.tsx
+++ b/src/features/formData/FormDataWriteStateMachine.tsx
@@ -100,6 +100,7 @@ export interface FDChange {
   // timeout is used. The debouncing may also happen sooner than you think, if the user continues typing in
   // a form field that has a lower timeout. This is because the debouncing is global, not per field.
   debounceTimeout?: number;
+  callback?: (result: FDSetValueResult) => void;
 }
 
 export interface FDNewValue extends FDChange {
@@ -168,6 +169,23 @@ export interface FDSaveFinished extends FDSaveResult {
     [dataType: string]: object;
   };
 }
+
+export interface FDSetValueSuccessful {
+  newValue: FDLeafValue;
+  convertedValue: FDLeafValue;
+  error: boolean;
+  hadError: boolean;
+}
+
+export const FDSetValueReadOnly = Symbol('FDSetValueReadOnly');
+export const FDSetValueEqual = Symbol('FDSetValueEqual');
+export const FDSetValueUnset = Symbol('FDSetValueUnset');
+
+export type FDSetValueResult =
+  | FDSetValueSuccessful
+  | typeof FDSetValueEqual
+  | typeof FDSetValueReadOnly
+  | typeof FDSetValueUnset;
 
 export interface FormDataMethods {
   // Methods used for updating the data model. These methods will update the currentData model, and after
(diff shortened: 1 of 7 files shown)


ivarne:
The algorithm you describe seems overly complex.
Would it be simpler to have 3 states on the input component
1. string: Text as shown/written in the input field
2. string: Last successful parse into number
3. bool: is current value parsable

* saving to backend and local expressions will only see 2
* if the current value isn't parsable, show a visual clue/error
* onBlur, we update the displayed value to see the actually saved value (if the current value is parsable).

olemartinorg:
I don't agree that's a better user experience. Currently we have a system of 'invalid data' in the data model (i.e. data that was set, but we cannot store in the data model as-is). This goes beyond just the Input component, as it's a function of the data model structure (JsonSchema) and the interaction with components. Every piece of data stored as 'invalid data' (in-memory) will trigger a validation error that prevents the user from proceeding/submitting the form until they change the data to something we _can_ store.

So, with `onBlur, we update the displayed value to see the actually saved value` it means we'll ignore changes made by the user, revert to some previous value (although it may be wildly different) and just never tell the user about it (unless they notice the change when they blur the field).

ivarne:
I said `(if the current value is parsable)`, with I intended to mean that it has one obvious transformation into a single valid value. For numbers this means that we can remove any spaces, traling zeros and a trailing `.` (not sure what rules we have for `2,000,000.44` vs `2.000.000,44`). Otherwise we obviously need to keep the error in memory and make the user explicitly fix it (or refresh ).

olemartinorg:
Hmm, yep, well, that sounds like what we already do (implemented in #1742 and later improved in #1877). This is a workaround/fix for a very specific problem introduced then, which caused `1.00` to fail. Numbers like `1.200` already worked fine, because `react-number-format` handles trailing zeroes already (showing them visually, but allowing them to be trimmed off the form data we pass the input field).

bjosttveit (approved):

PR #3323, opened 2025-04-26
Flattening language files, adding eslint rule to catch broken text key references

olemartinorg:
## Description

I noticed that the text key used in the helpText for 'bolignummer' in the Address component was broken:

20250426_13h51m11s_grim

Typescript can't help us much here, because the expected type is either just any string (for text resources defined in an app) or a valid text key as defined in `en.ts`/`nb.ts`/`nn.ts`. However, literal strings given as a language key should always point to text keys that are defined in app-frontend, so eslint can do that (and differentiate between a variable containing a string and a literal string) to warn us when making this mistake.

Also flattening our text files is something I've been thinking about for a long time (it may look nicer with nested objects, but it's just indirection and a format you 'need to know about'). By flattening them, we get back proper autocomplete support via the typescript types, along with simpler code to look them up.

## Related Issue(s)

- closes #{issue number}

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels


src/layout/Custom/useLegacyNestedTexts.ts
```diff
@@ -0,0 +1,41 @@
...
+}
+
+interface NestedTexts {
+  [n: string]: string | NestedTexts;
+}
+
+/**
+ * Splits the text keys by '.' and creates a nested object.
+ */
+export function convertToLegacyNestedTexts(language: FixedLanguageList): NestedTexts {
```
bjosttveit, line 19: Isn't the `dot-object` package made for this?

src/layout/Custom/CustomWebComponent.tsx
```diff
@@ -32,7 +33,8 @@ export function CustomWebComponent({
   ...passThroughPropsFromGenericComponent
 }: ICustomComponentProps) {
   const langTools = useLanguage();
-  const { language, langAsString } = langTools;
+  const langAsString = langTools.langAsString;
+  const legacyLanguage = useLegacyNestedTexts();
```
bjosttveit, line 37: Should we plan to deprecate this in the next major or keep it around?

bjosttveit (approved):
This is great, I keep forgetting to add `general.` when using language and getting frustrated 😅

src/layout/Custom/useLegacyNestedTexts.ts
olemartinorg: Woah, I didn't even think of that! Very good point, I'll do one more pass here.

src/layout/Custom/CustomWebComponent.tsx
olemartinorg: Definitely, but I think deprecation now may be a bit early. I don't remember if you were in that meeting, but a short while ago when we talked about Custom components there was a general consensus that we should do a more thorough job to define a next-level SDK/API for these custom components, and do it in a way that makes it much more useful than what we have now. If we decide to do that, I really hope that giving the custom components our internal data structures to play with is the first thing to go..

PR #3402, opened 2025-05-28
Fixes for embedded FormProvider

olemartinorg:
## Description

I recently added a few `Summary2` components to `ttd/frontend-test` which references all the previous tasks. When doing that, I noticed a few errors in DevTools that weren't there before. It turns out that two unanticipated things happened here:

1. The `Effect*` components were rendered inside the node hierarchy, which caused one of the components to try to set `preselectedOptionIndex` in the database. Since the data model here belonged to a previous task and was locked, `FormDataWrite` stopped to attempt with an error. This PR prevents these effects from running at all when in a "read only" state such as rendering for Summary2.
2. If an embedded layout-set sets up dynamics, i.e. rules that run via javascript objects, loading the javascript could end up overwriting dynamics for the parent FormProvider and ruining dynamics there. This can happen if set up in a subform layout-set, but also if rendered via Summary2 pointing to a previous task. This PR adds an error to the DevTools about this, and refuses to run rules in embedded `FormProvider`.

I still have left to check existing apps and who might be affected by this slight change. I should inform app developers that have set up dynamics when using FormProvider embedding, in addition to documenting when doing such doesn't work as one might think.
- [x] Write docs (deferred to #3421)
- [x] Test existing apps
  - [x] Inform app developers

## Related Issue(s)

- https://altinn.slack.com/archives/C065CM2V6V7/p1748602850388139

## Verification/QA

  - [X] I have tested these changes manually
- Automated tests
  - [X] No automatic tests are needed here (no functional changes/additions)
  - [X] No testing done/necessary (no DOM/visual changes)
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
  - [X] This change/feature does not require any changes to Altinn Studio
- Sprint board
  - [X] The original issue (or this PR itself) has been added to the Team Apps project and to the current sprint board
- Labels
  - [X] I have added a `kind/*` and `backport*` label to this PR for proper release notes grouping


cammiida:
Is this only relevant for the Rulehandler file that will be deprecated for v5/v9? And has anyone reported that they need this? 

If the answers are yes and no to these questions, respectively, I would suggest maybe not merging this since it introduces more complexity. That is if it is not really a current and reported user need. Furthermore, it is also implemented with the node hierarchy, which you are working on removing, so the implementation might be outdated shortly anyways?

olemartinorg:
It's not just the RuleHandler, it's also about the effects that run (such as setting `preselectedOptionIndex`).

As for RuleHandler, while we can probably deprecate it for v5/v9, we can't realistically remove it out unless we provide expression-support for everything that was possible in RuleHandler. I suspect it will take longer than we might think - remember the recent comments about removing the last remnants of support for the older PDF generator. Having RuleHandler in a form while using the `Subform` or `Summary2` components (when pointing to a previous task) will break the rules in the parent FormProvider, which this PR fixes. That affected one form I found in production, but this introduces an error message letting app developers know about the problem in case they try to add those `Subform` or `Summary2` components in other existing apps with legacy rules. 

> Furthermore, it is also implemented with the node hierarchy, which you are working on removing, so the implementation might be outdated shortly anyways?

Yes, the effects are implemented there, but I can't refactor that out from there ahead of time just for this change. Surely it might be outdated soon, but I think the functionality here should be kept regardless of what we do in the future (i.e. don't run effects in readOnly embedded forms, like Summary2 pointing to a previous task).

adamhaeger:
I see both points here. If its functionality we're gonna need for the foreseeable future I think it should at least work so maybe we need to live with this extra complexity. I do think we should create a plan to migrate apps that use this functionality somehow if we are getting rid of this support though, I dont see what that shouldnt be possible.

However, Im still not a fan of introducing ever more contexts, especially not with this `Maybe*` pattern, but I've banged that drum enough 😆

olemartinorg:
Thanks for the feedback, everyone. I do agree that it's worth fixing problems now, even if code will be changed later on.

1. There are not really any new contexts introduced here, just a property added to the existing `FormProvider` to make it possible for a `readOnly` variant.
2. The `Maybe*`-providers were a way to gather `FormProvider` into one again, not using `Outer` and `Inner` to conditionally provide `PaymentProvider` (like we do now). To make sure rule handler was not injected for embedded forms, the same thing (conditionally applying a provider) was needed for `RulesProvider`, so I made wrappers for both `PaymentProvider` and `RulesProvider`. I think it made `FormProvider` easier to read, but if there are alternative suggestions I'm all ears.

martinothamar (approved):

PR #3411, opened 2025-06-02
Add more shared tests for ["text"] expression

ivarne:
I wrote some extra tests to double check the backend implementation for ["text"] expressions, just want to sync up with frontend.


olemartinorg (approved):
Looks good! 🙌

Issue #3147, opened 2025-03-21
Refactoring route: Getting to a simpler architecture, solving performance issues


olemartinorg:
We've been working on a proof-of-concept for what the architecture can look like in app-frontend in the future. Our goal is to simplify as much as possible, making code easier to reason about. This is especially regarding the node hierarchy/layout-generator structures that have proven to be both difficult to maintain, understand, and introduce performance/memory-usage issues on their own.

Below is a list of concrete steps we can implement to get away from this situation. It is meant to be executed in this order, but may not be exhaustive. As of writing this issue, this list is meant as a suggestion for one alternative route we can take (the other being expanding on the proof-of-concept to make it feature-complete).

1. ✔️ Refactor those places using `nodeData.item` somewhat directly
    - `Form.tsx` looks up every `required` property for all components on a page to display the 'Obligatoriske felter er markert med *' bar at the top of the page. This should instead look at the external layout config, not the resolved expressions, because an expression just means _there is a component here that potentially can be required_, and if we toggle that bar on/off according to the live state of the page, we just get layout-shifts. The header should not be rendered on the top when we _know there are no required components on the page, ever_. #3447
    - `useExtendedRepeatingGroupState()`, `RepeatingGroupsFocusProvider`, `RepeatingGroupTableRow` can be refactored to not rely on rows being stored in `nodeData.item` (and then we can remove `RepeatingChildrenStorePlugin` entirely). The number of rows should be looked up in the data model instead, and we can resolve row-level expressions directly. #3448
    - #3449 
    - #3450    
    - `useNodeItem()` can temporarily be rewritten to just resolve all expressions for the node. That will de-optimize those places we select only a few values. #3451
2. ✔️ Find usages of `nodeData.layout` and refactor those
    - Some of these can use `layoutLookups` instead (if just looking up the external component definitions)
    - Others read the (indexed) node id, or indexed data model bindings. We can expand the new `DataModelLocationProvider` to also pass on mutators, and using that we can make hooks that will give you the current indexed node-id, mapping and data model bindings.
    - #3452
3. ⌛ Build a better way to evaluate single expression
    - I've been thinking about this, and I have at least one suggestion. This should focus on the best possible DX, while maintaining full typescript support for return values. Preferably something like `const something = useEvalExpression(component.someExprProperty)`.
    -  This should make our components re-render only when the expressions we actually needed gets new values (fixing the problem introduced by refactoring `useNodeItem()` in `#1`.
[...]

PR #3443, opened 2025-06-12
Evaluating expressions only as-needed

### LINKED ISSUE #3451: Node-refactor 1.5: Evaluate expressions in `useNodeItem()` directly
`useNodeItem()` can temporarily be rewritten to just resolve all expressions for the node. That will de-optimize those places we select only a few values, but we can fix that properly later.

olemartinorg:
## Description

This change rewrites the `useNodeItem()` hook and rewrites every other piece of code that relied on `nodeData.item`. Instead of evaluating expressions in the node hierarchy and storing them in `NodesContext`, we'll now evaluate them when `useNodeItem()` is called.

This should bring lower memory usage, and overall better performance, but it's still not optimal. Wherever only a part of the item is selected (or actually used), all properties inside will be evaluated anyway (according to the `evalExpressions()` function in the component `index.tsx` file). In the future I want to evaluate each property when needed.

A few things to note here:
- I removed the `AlertOnChange` plugin, as it's the only one to implement expression-evaluation in it. It's simpler to just evaluate these in each component that needs them.
- All container components that have `children` (or similar properties) will now transparently use that property for listing out children, instead of having an 'internal' type that has gone through `itemFactory()` of some plugin first. This also means that we have to filter out children that are not supposed to render there (according to their `CompCapabilities`, such as `renderInAccordion: boolean`). I made a new hook for this, called `useHasCapability()`.
- Since all `children` now have base component ids in them (not indexed ones), I made a hook called `useIndexedId()` to transform them. However, there's starting to be quite an overlap here with the (more complex) functionality around the existing `useIndexedComponentIds()` hook. However, that hook implements some useful functionality to check that a component id actually exists and is valid in that context, so removing it (or simplifying it) would break shared expression tests. I made a note about the similarity in a doc comment, but sometime in the future it would be nice to not have to deal with all this.

## Related Issue(s)

- #3147
- closes #3451

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels
  
…n nodeData.item. This breaks stuff, but I'll fix that later - it's easier to figure out what I need to fix when TypeScript tells me what it is.


src/components/label/Label.tsx
```diff
@@ -27,10 +28,23 @@ export type LabelProps = PropsWithChildren<{
...
+  return (
+    <LabelInner
+      item={_item}
+      nodeId={props.node.id}
+      {...props}
+    />
+  );
+}
+
+export function LabelInner(props: LabelInnerProps) {
```
cammiida, line 44: Er det en spesiell grunn til at du har delt opp i Label og InnerLabel her? Syns det kanskje blir litt ekstra effort å lese 😊

PR #3509, opened 2025-06-27
Refactor: No longer passing node object to `useNodeItem()`

olemartinorg:
## Description

This was simply a chore. To get to a place where we don't need the `LayoutNode` objects anymore, we have to get rid of the places that pass them around. I would assume the most-used hook using taking that object in is the `useNodeItem()` hook that returns the layout config with expressions already evaluated.

I wanted to simply pass `baseComponentId` instead, but it's not possible to replace `useNodeItem()` with just one hook. Instead we get three:
- `useItemWhenType(baseComponentId, type)` is the one you'll most likely want to use. Given a component id (without indexes) and a component type, you'll get back the layout config for that component id, with expressions resolved, but only when the type matches. If the type doesn't match, things will crash. That's a good thing - it prevents sneaky bugs where you'll get config you didn't expect for another component type.
- `useItemIfType(baseComponentId, type)` is useful in a handful of cases where the previous hook crashing isn't optimal. If the type doesn't match the type you expect, this will instead just return `undefined`. This is useful when you don't know the type beforehand, but only need to evaluate expressions if the type is the one you're looking for. Rarely used in practice.
- `useItemFor(baseComponentId)` is for everything else. You don't know the type, and you don't care which type it is. You just want the evaluated expressions, and that's it. You'll get a `CompInternal` back, but nobody knows the exact type unless you check it afterwards.

In some cases I used `useExternalItem()` or other hooks instead, because the code wasn't really using any properties with possible expressions in them, so then there's no need to evaluate them either.

I do think these hooks will only be one stop along the way to a better solution, hopefully one where each property with an expression can be resolved individually, so that we don't get data sources for things we don't need in that context - and only re-render when the data we used is the data that changes. However, to get rid of `LayoutNode`, `LayoutPage` and `LayoutPages`, this is the obvious first thing to tackle.

## Related Issue(s)

- #3147

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels

### DIFF
--- a/src/features/devtools/components/NodeInspector/NodeInspectorDataField.tsx
+++ b/src/features/devtools/components/NodeInspector/NodeInspectorDataField.tsx
@@ -141,8 +141,8 @@ function ExpandArray(props: { path: string[]; property: string; elements: unknow
 
 export function NodeInspectorDataField({ path, property, value: inputValue }: NodeInspectorDataFieldParams) {
   const { node } = useNodeInspectorContext();
-  const firstRowExpr = RepGroupHooks.useRowWithExpressions(node?.isType('RepeatingGroup') ? node : undefined, 'first');
-  const itemWithExpressions = useIntermediateItem(node?.baseId);
+  const firstRowExpr = RepGroupHooks.useRowWithExpressions(node.isType('RepeatingGroup') ? node : undefined, 'first');
+  const itemWithExpressions = useIntermediateItem(node.baseId);
 
   let value = inputValue;
   const preEvaluatedValue = dot.pick(path.join('.'), itemWithExpressions);
--- a/src/features/devtools/components/NodeInspector/NodeInspectorDataField.tsx
+++ b/src/features/devtools/components/NodeInspector/NodeInspectorDataField.tsx
@@ -11,6 +11,7 @@ import { canBeExpression } from 'src/features/expressions/validation';
 import { RepGroupHooks } from 'src/layout/RepeatingGroup/utils';
 import { useIntermediateItem } from 'src/utils/layout/hooks';
 import { LayoutNode } from 'src/utils/layout/LayoutNode';
+import type { GroupExpressions } from 'src/layout/RepeatingGroup/types';
 
 interface NodeInspectorDataFieldParams {
   path: string[];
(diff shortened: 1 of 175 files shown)


src/utils/layout/useNodeItem.ts
```diff
@@ -1,51 +1,84 @@
...
+ * @see useNodeItemIfType Use this one if you only want to evaluate expressions _if_ the type is the expected one.
  */
-export function useNodeItem<N extends LayoutNode, Out>(node: N, selector: (item: NodeItemFromNode<N>) => Out): Out;
-// eslint-disable-next-line no-redeclare
-export function useNodeItem<N extends LayoutNode>(node: N, selector?: undefined): NodeItemFromNode<N>;
-// eslint-disable-next-line no-redeclare
-export function useNodeItem(node: LayoutNode, selector: never): unknown {
-  const intermediate = useIntermediateItem(node.baseId);
-  const location = useDataModelLocationForNode(node.id);
+export function useItemWhenType<T extends CompTypes>(
```
adamhaeger, line 18: Why is it so important to know the type in all these functions? `useIntermediateItem` will return the type anyway right?

src/features/devtools/components/NodeInspector/NodeInspectorDataField.tsx
```diff
@@ -139,19 +140,59 @@ function ExpandArray(props: { path: string[]; property: string; elements: unknow
...
+  return (
+    <NodeInspectorDataFieldInner
+      node={node}
+      firstRowExpr={firstRowExpr}
+      {...rest}
+    />
+  );
+}
 
+function NodeInspectorDataFieldInner({
```
adamhaeger, line 180: Using this Inner pattern is something I think we should move away from, but fine for now in the spirit of getting things done :)

olemartinorg: Yup, it's not exactly beautiful. If you have a suggestion for how to solve this differently, please create an issue with `quality/debt`. :pray: If it's just the naming you disagree with, what would you have called it?

PR #3538, opened 2025-07-11
Refactor: Moving `hidden` state out from `NodesContext`

olemartinorg:
## Description

This removes the `hidden` state from the node hierarchy, and removes a lot of cruft that we no longer need.

Removed:
- Node readiness (this means component rendering is now fully independent from the node hierarchy, and in principle `NodesContext` should not make the user wait anything during app start-up).
- Stages (the plugins that remain and still saves to the node hierarchy now just saves their data when the node has been added)
- Commit queuing (changes will just be set into the store immediately now)
- Logging and debug output for node hierarchy
- Checking if a node is hidden before navigating to it. It shouldn't be an issue anyway, because we shouldn't render navigation buttons that makes it possible to navigate to a component that is hidden.

Changed:
- Everything needed to figure out if a component is hidden has now been moved to `src/utils/layout/hidden.ts`. Expressions for the current component and parent components will be evaluated if they exist.
- Some expression lookups needs to know if a component is hidden (such as the `component` function). That now works much the same as the `displayValue` function - we'll find all lookups inside the expression and run the `useIsHiddenMulti()` hook to figure out which ones are hidden before running the expression. This means `component` lookups can no longer dynamically switch between different component ids, but luckily no apps had implemented expressions like that.
- Sadly it was hard to change the validation system to run a hook for figuring out if a component is hidden, so instead it still opts to store the `hidden` state in `NodesContext` :face_with_peeking_eye: But, that value is only read for validations, and will be removed when validations are moved out from the node hiearchy.
- Lots of minor test failures here, but that's mostly because apps are noticeably faster now. A good example was `saving.ts` which failed because form data was able to debounce and trigger a save before cypress was able to focus to the next field and type there - previously that would delay the save until the user stopped typing in the next field. I made that more explicit by adding a feature toggle for turning off debounce-on-blur.

## Related Issue(s)

- closes #{issue number}

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels


src/layout/Input/InputComponent.tsx
```diff
@@ -138,7 +139,7 @@ export const InputVariant = ({
     readOnly,
     textonly: overrideDisplay?.rowReadOnly && readOnly,
     required,
-    onBlur: FD.useDebounceImmediately(),
+    onBlur: () => debounce(`blur`),
```
paal2707, line 142: nitpick: vanlige fnutter her kanskje? `' '`

paal2707 (approved):
Looking good

PR #3734, opened 2025-09-23
ADR for backend data initialization

adamhaeger:
## Description

The Frontend Next task force is proposing to load initial app data on the backend.

## Related Issue(s)

- closes #16309

## Summary by CodeRabbit

* **Documentation**
  * Added an Architecture Decision Record describing a move to backend-driven initial data bootstrapping, embedding immutable app and optional instance data in server-rendered payloads to reduce redundant refetches and target ≤2 requests.
  * Documents adoption of path-based routing (preserving legacy hash URLs), planned pagination upgrades, rollout and security considerations; no immediate visible user-facing changes.

### DIFF
--- a/adr/002-backend-data-initialization.md
+++ b/adr/002-backend-data-initialization.md
@@ -123,7 +123,7 @@ The frontend receives either:
 
 Flow Comparison
 
-### Before:
+### Before (simplified):
 
 ```mermaid
 flowchart TD
@@ -139,9 +139,11 @@ B7 -->|Has instances| B9[Client-side redirect<br/>to instance selection]
 B7 -->|Direct access| B10[Render app]
 end
 
+style Before fill:#ffeeee
+
 ```
 
-### After:
+### After (simplified):
 
 ```mermaid
 flowchart TD
@@ -154,7 +156,6 @@ subgraph After["After: Backend-Driven Redirects"]
     A4 -->|Direct access| A7[Render HTML with<br/>window.AltinnAppData]
 end
 
-style Before fill:#ffeeee
 style After fill:#eeffee
 ```
 


olemartinorg (approved):

adr/002-backend-data-initialization.md
```diff
@@ -0,0 +1,157 @@
...
+  </script>
+  <script>
+    window.AltinnInitialInstanceData = JSON.parse(
+      document.getElementById('__initialInstanceData')?.textContent || 'null',
+    );
+  </script>
+  ```
+
+- **Routing:**
+  - Backend maps `/` `/instances/:id/*` `/tasks/:taskId` etc. to a single `HomeController` action that embeds the correct initial payload and serves static assets.
```
martinothamar, line 104: Is it actually possible to make this work with our current API routes? E.g. we have `{org}/{app}/instances/{instanceOwnerPartyId:int}/{instanceGuid:guid}` as a route in `InstancesController`. Do we get a collision if we remove the hash? Or was the frontend route `instance` (singular)?

adr/002-backend-data-initialization.md
```diff
@@ -0,0 +1,157 @@
...
+    type="application/json"
+  >
+    { ... AppData ... }
+  </script>
+  <script
+    id="__initialInstanceData"
+    type="application/json"
+  >
+    { ... } <!-- present only when opening an instance -->
+  </script>
```
martinothamar, line 92: Is this including stuff like forms? If not, I guess we control the `Instance` model mostly, apart from dataValues and some stuff like that?

martinothamar: Asking partly due to security, partly due to size/perf considerations. Is there a size at which it makes more sense to still load these asynchronously? I.e. blocking initial render due to parsing these elements

martinothamar: Alternatives I think could be SSE or HTTP 2 server push. Unsure of browser compatibility, but would be nice if they were mentioned.

adamhaeger: Yes we have briefly discussed this, but we wanted to keep the ADR high level, I suggest we bring these considerations into the implementation phase :)

PR #3882, opened 2025-12-02
Store ajv validators in query cache

bjosttveit:
## Description

Improves performance when loading an app PDF with many subform entries. Generate ajv validator inside of the query function so it gets cached so we don't have to do the same work for each subform. Loading subform-test PDF with ~100 subforms took around 4x longer before this.

## Summary by CodeRabbit

* **Refactor**
  * Unified how data model schemas and validators are stored and accessed, simplifying schema lookup and reducing redundant validation work.
  * Components now consume a single schema+validator result, improving consistency and load gating across the app.
* **Tests**
  * Updated tests to reflect the new schema+validator result shape used by consumers.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/src/features/validation/schemaValidation/SchemaValidation.tsx
+++ b/src/features/validation/schemaValidation/SchemaValidation.tsx
@@ -19,7 +19,7 @@ export function SchemaValidation({ dataType }: { dataType: string }) {
   const updateDataModelValidations = Validation.useUpdateDataModelValidations();
 
   const formData = FD.useDebounced(dataType);
-  const { validator, rootElementPath, schema } = DataModels.useDataModelSchema(dataType);
+  const { validator, rootElementPath, schema } = DataModels.useDataModelSchema(dataType) ?? {};
   const dataElementId = DataModels.useDataElementIdForDataType(dataType) ?? dataType; // stateless does not have dataElementId
 
   /**


src/utils/layout/generator/validation/GenerationValidationContext.tsx
```diff
@@ -51,18 +51,23 @@ function FetchLayoutSchema({
...
+      }
+      return makeValidateFunc(createLayoutValidator(schema));
+    },
   });
 
   useEffect(() => {
-    if (isSuccess && layoutSchema) {
-      const ajv = createLayoutValidator(layoutSchema);
-      setSchemaValidator(() => makeValidateFunc(ajv));
+    if (isSuccess && validate) {
```
framitdavid: Jeg tror vi kan kvitte oss med hele denne `useEffect`-en? Ved å bruke onSuccess i TanStack? 😊 
```ts
const { data: validate } = useQuery({
  enabled,
  queryKey: ['fetchLayoutSchema'],
  queryFn: async () => {
    const schema = await fetchLayoutSchema();
    if (!schema) {
      return null;
    }
    return makeValidateFunc(createLayoutValidator(schema));
  },
  onSuccess: (validate) => {
    if (validate) {
      setSchemaValidator(validate);
    }
  },
});

PR #4025, opened 2026-02-27
Handle webhook pattern for payment

Magnusrm:
## Description

Remove automatic calling of process next, both in the case the payment-status is "paid" or "skipped", since this will now be handled in backend using the new webhook. 

In the case we get a mismatch between the current task state in the browser and the current task in backend, we treat the backend as the source of truth and navigate to the correct task without calling process next.

## Related Issue(s)

- closes #{issue number}

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels
  

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed an unintended automatic "skip" action that previously advanced users when payments reached certain statuses, preventing unexpected navigation.
  * Navigation now verifies backend state before advancing, preventing stale screens and incorrect task transitions.

* **New Features**
  * "Next" performs a real-time verification before proceeding and is temporarily disabled while checking to prevent duplicate actions and race conditions.


src/utils/urls/appUrlHelper.ts
```diff
@@ -30,6 +30,11 @@ export const getPaymentInformationUrl = (instanceId: string, language?: string)
   return `${origin}/${org}/${app}/instances/${instanceId}/payment${queryString}`;
 };
 
+export const getPaymentInformationForTaskUrl = (instanceId: string, language?: string, taskId?: string) => {
```
walldenfilippa, line 33: Minor nitpick 😊 Both implementations getPaymentInformationForTaskUrl  and getPaymentInformation build the same URL pattern. The only difference is the taskId in the query string. Can you add taskId? as an optional parameter to the existing function instead?

src/layout/Payment/PaymentComponent.tsx
```diff
@@ -23,12 +27,70 @@ export const PaymentComponent = ({ baseComponentId }: PropsFromGenericComponent<
...
     throw new Error('Cannot use PaymentComponent in a subform');
   }
 
-  const disabled = isAnyProcessing || isConfirming || isRejecting;
+  const [isChecking, setIsChecking] = React.useState(false);
+  const disabled = isAnyProcessing || isConfirming || isRejecting || isChecking;
+
+  const navigateBasedOnProcess = useCallback(
+    async (shouldConfirmIfNoNavigate: boolean) => {
+      if (!(paymentInfo?.status === PaymentStatus.Paid || paymentInfo?.status === PaymentStatus.Skipped)) {
```
walldenfilippa, line 44: I might be missing something, but I’m not sure how the `Skipped` case is meant to be used?`navigateBasedOnProcess` checks for both `Paid` and `Skipped`, while `useBackoff` only passes `goToCurrentTask `for `Paid`. 
Is `Skipped` a valid status in the webhook flow? If yes, it may make sense to include it in the `usebackoff` condition as well, otherwise, it may make sense to remove it from `navigateBaseOnProcess`?

walldenfilippa (commented):
Nice done! Tested it manually and it worked great 😄  
I added to minor comments.

src/layout/Payment/PaymentComponent.tsx
Magnusrm: Good catch! Skipped is a valid status for the paymentStatus. It is used f.ex. when someone filling out the form has nothing to pay for, but still has to pass through the payment step. I think also it might be used when choosing to pay via invoice but i might remember wrong.

I agree the logic should be the same in the backOff. I might even be able to simplify it by only having the logic in one of the places. 🙌

src/utils/urls/appUrlHelper.ts
Magnusrm: I did this in order to make a clear separation between the old endpoint and the new and to make sure i did not end up hitting the old endpoint. It might work to just add taskId as an optional parameter, I'll check it out!

walldenfilippa (approved):
Looks good! 😄

PR #4047, opened 2026-03-06
chore: display error message to match grid label

### LINKED ISSUE #4030: Better handle how error messages are shown for components using innerGrid
Image

Tilby en konfigurasjonen for lengden (innerGrid) til feilmelding

Alternativt alltid følge griden til label 

Alternativt så kan denne løse det:  https://github.com/Altinn/app-frontend-react/issues/4031

JamalAlabdullah:
## Description

- Refactor to match this [suggestion from Ole Martin](https://github.com/Altinn/app-frontend-react/issues/1949#issuecomment-4110024787) 
- Link to [docs PR ](https://github.com/Altinn/altinn-studio-docs/pull/2817)

This PR improves `ComponentStructureWrapper` by simplifying the inner/validation grid merging logic and making exact-width input behavior easier to read. by `mergeMaxGridSizePair` and `mergeMaxGridStyling`.
It also adds `validationGrid` to the shared generated grid type in `codegen/common.ts` ;

```
 new CG.prop(
        'validationGrid',
        CG.common('IGridStyling')
          .optional()
          .setTitle('Validation grid')
          .setDescription('Column span for validation messages; defaults to innerGrid when omitted'),
      ),
```

For now the validation messages use  grid.validationGrid and when it’s missing, it falls back to innerGrid.

## Related Issue(s)

Test in ra0479-02;

https://github.com/user-attachments/assets/466e6f5c-296f-49b2-b736-9056fe72538a

Test in forntend-test app:

https://github.com/user-attachments/assets/f022cd19-67b4-4de8-a3f1-571964702daf

- closes #4030 

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels
  

## Summary by CodeRabbit

* **New Features**
  * Improved responsive input sizing with breakpoint-aware width options and an "exact width" constraint for more consistent input layout.

* **Refactor**
  * Layout structure updated so form content and validations are arranged in dedicated blocks, improving alignment and wrapping while preserving validation visibility.

* **Tests**
  * E2E assertion made more tolerant to minor CSS flex variations to reduce flakiness.


lassopicasso (commented):
One comment about the width👍 

I am unsure, but @Magnusrm mentioned that this change can perhaps be a breaking change for existing apps' layouts. 
Maybe we should discuss this with the team on slack and see if someone has any more thoughts around this before we merget it.

lassopicasso (commented):
It doesn't seem to work when I set the config. The validation still follows the same width as the input field. Am I doing something wrong? I suspect something unexpected might be happening around `inputGridSizeAndExactWidth`.

I tested in frontend-test app in the first input field in form layout.
Config:
Image

Result:
Image

I’m also finding it a bit difficult to follow the flow here, so could you explain the logic behind the decisions in `ComponentStructureWrapper` in details, especially `inputGridSizeAndExactWidth`? I would try to avoid the number of abstraction levels. And a nitpick; from the naming, I would expect this to apply only to the input, but it also set the validation width.

I'm also curious why we need to set specific grid breakpoints and styling here, when `Flex` already handles this based on the size prop.

My initial thought is that the validation width logic could potentially be handled at a lower level when `showValidationMessages` is defined, but I may be missing some context.

What I think is happening now:

- `Flex` wrapper sets the max grid size using `contentGridSize`
- `contentGridSize` is set to the largest of innerGrid and validationGrid
- The inner Flex components then cannot exceed the outer Flex grid size

If you need a quick review before we go on vacation on friday (I'll be away for 3 weeks), it might be helpful to also check with someone in same time zone for another round. I’ll likely only have time for one more review, and because of the time difference I unfortunately won’t be able to provide quick follow-up.

src/layout/ComponentStructureWrapper.tsx
```diff
@@ -31,16 +70,35 @@ export function ComponentStructureWrapper({
...
     <Flex
       id={`form-content-${indexedId}`}
       className={className}
-      size={{ xs: 12, ...grid?.innerGrid }}
+      size={contentGridSize}
       style={style}
       item
+      container
+      direction='column'
+      flexWrap='nowrap'
```
lassopicasso: What is the purpose of changing the default behavior for `direction` and `flexWrap`?
- `flexWrap`: this is currently wrap by default in `Flex`, so changing it could potentially affect existing layouts. If this change is intentional, it would be good to test a variety of cases, since `ComponentStructureWrapperProps` is used extensively across many components.
- `direction`: setting this to "column" changes the default behavior from `row`. This might introduce breaking changes in existing layouts, although I may be mistaken.

lassopicasso (commented):
Great job! The code looks much better now and it's easier to follow the flow.
I have a few comments, but overall the changes look good 👍

I also tested it, and it seems I can now set a custom width for the validation message. However, it looks like the input and validation together only take half of the width they’re supposed to.

For example, this config should take half of the layout width, but the input and validation message only take half of that again. So it seems like the width is being applied twice somewhere. I compared it with main branch to make sure my app isn't wrong, and in main it took 50% width of the layout.

```
        "grid": {
          "innerGrid": {
            "md": 6
          },
          "validationGrid": {
            "md": 6
          }
        }

```
Image


Since I won’t have the opportunity to do another review today, and I’m leaving for holiday for 3 weeks, I’d suggest asking someone else to take a second look after updating the PR based on the feedback. 🙂

walldenfilippa (approved):
Nice work! 🥳

PR #4135, opened 2026-04-23
fix: fixed align tabel title

### LINKED ISSUE #4122: Tittel i tabell alignes til høyre
### Description of the bug

"Antall" som table header på kolonnen her (på repeating group med `editInTable: true` havnet til høyre, mens de andre er til venstre. Klarer ikke se hvorfor)
Image

### Steps To Reproduce

Aktuelt repo dersom dere vil se nærmere på koden: https://altinn.studio/repos/mat/akvakultur-driftsplan
Skjemaet ligger på https://mat.apps.tt02.altinn.no/mat/akvakultur-driftsplan#/Info
Dere kan bruke denne testbrukeren for å komme inn på skjemaet:

Org: 310043842 - FREDFULL GEOMETRISK TIGER AS
Daglig leder: 15892849339

### Additional Information

_No response_

JamalAlabdullah:
## Description

`useColumnStylesRepeatingGroups` now takes a boolean `isTitle` props, and passes it to `useTextAlignment` so header/title alignment is handled centrally in one place.
This keeps numeric right-alignment for data cells while ensuring header titles default to left unless explicitly configured via alignText.

## Related Issue(s)

- closes #4122 

Screenshot 2026-04-23 at 13 43 24

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels
  - [ ] I have added a `kind/*` and `backport*` label to this PR for proper release notes grouping
  

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed inconsistent text alignment in repeating-group table headers and titles so column headers and titles respect per-column alignment (defaults to left for title/header when appropriate).

* **Tests**
  * Added a test ensuring numeric/editable table columns render left-aligned headers to prevent regressions.

### DIFF
--- a/src/layout/RepeatingGroup/Table/RepeatingGroupTableTitle.tsx
+++ b/src/layout/RepeatingGroup/Table/RepeatingGroupTableTitle.tsx
@@ -16,11 +16,7 @@ interface IProps {
 }
 
 export const RepeatingGroupTableTitle = ({ baseComponentId, columnSettings }: IProps) => {
-  const style = useColumnStylesRepeatingGroups(baseComponentId, columnSettings);
-  const titleStyle = {
-    ...style,
-    '--cell-text-alignment': columnSettings[baseComponentId]?.alignText ?? 'left',
-  } as React.CSSProperties;
+  const style = useColumnStylesRepeatingGroups(baseComponentId, columnSettings, { isTitle: true });
   const tableTitle = useTableTitle(baseComponentId);
   const { getRequiredComponent, getOptionalComponent } = useLabel({
     baseComponentId,
@@ -30,7 +26,7 @@ export const RepeatingGroupTableTitle = ({ baseComponentId, columnSettings }: IP
   return (
     <span
       className={classes.contentFormatting}
-      style={titleStyle}
+      style={style}
     >
       <Lang id={tableTitle} />
       {editInTable && getRequiredComponent()}
--- a/src/layout/RepeatingGroup/Table/RepeatingGroupTableTitle.tsx
+++ b/src/layout/RepeatingGroup/Table/RepeatingGroupTableTitle.tsx
@@ -16,7 +16,7 @@ interface IProps {
 }
 
 export const RepeatingGroupTableTitle = ({ baseComponentId, columnSettings }: IProps) => {
-  const style = useColumnStylesRepeatingGroups(baseComponentId, columnSettings, { isTitle: true });
+  const style = useColumnStylesRepeatingGroups(baseComponentId, columnSettings, true);
   const tableTitle = useTableTitle(baseComponentId);
   const { getRequiredComponent, getOptionalComponent } = useLabel({
     baseComponentId,
--- a/src/layout/RepeatingGroup/Table/RepeatingGroupTableTitle.tsx
+++ b/src/layout/RepeatingGroup/Table/RepeatingGroupTableTitle.tsx
@@ -17,6 +17,10 @@ interface IProps {
 
 export const RepeatingGroupTableTitle = ({ baseComponentId, columnSettings }: IProps) => {
   const style = useColumnStylesRepeatingGroups(baseComponentId, columnSettings);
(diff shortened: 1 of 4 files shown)


src/utils/formComponentUtils.ts
```diff
@@ -145,8 +146,9 @@ export function useColumnStylesRepeatingGroups(
   }
 
   const columnCopy = { ...column };
-  columnCopy.alignText = columnCopy.alignText ?? textAlignment;
-
+  if (options?.deriveAlignmentText !== false) {
+    columnCopy.alignText = columnCopy.alignText ?? textAlignment;
+  }
```
lassopicasso: With the hardcoding `deriveAlignmentText: false` in `TitleCell` in file `RepeatingGroupTableTitle.tsx` it will never actually allow user to config the `alignText` as far as I understund. Because of `if (options?.deriveAlignmentText !== false) `

lassopicasso (commented):
Nice PR, but i think this solution fix only this specific problem, and not the core problem that can lead to new problems for other cases.

I added a comment about what may be the core problem here: https://github.com/Altinn/app-frontend-react/issues/4122#issuecomment-4311680680

I would suggest take a new look on the root problem, and see if the underlying problem is in the structure in above levels where `useColumnStylesRepeatingGroups` is used.


Maybe a solution could be to pass down `isTitle:boolean` prop to `useColumnStylesRepeatingGroups` -> `useTextAlignment` could solve this, and keep the styling logic in `useTextAlignment` instead of adding extra styling other places that is harder to maintain later.
```
function useTextAlignment(baseComponentId: string, isTitle: boolean): 'left' | 'center' | 'right' {
  const component = useLayoutLookups().getComponent(baseComponentId);
  const formatting = component.type === 'Input' ? component.formatting : undefined;
  if (!formatting || && isTitle) {
    return 'left';
  }
  if (formatting.align) {
    return formatting.align;
  }
  return formatting.number ? 'right' : 'left';
}

```

What do you think?

olemartinorg:
@lassopicasso That seems like a good solution to me - it looks to be a very targeted fix that won't affect anything else. :raised_hands:

src/utils/formComponentUtils.ts
JamalAlabdullah: I guess passing the props as you suggested is better 👍

lassopicasso: Yeah, I think that is cleaner and more safe.

src/layout/RepeatingGroup/Table/RepeatingGroupTableTitle.tsx
```diff
@@ -16,7 +16,7 @@ interface IProps {
 }
 
 export const RepeatingGroupTableTitle = ({ baseComponentId, columnSettings }: IProps) => {
-  const style = useColumnStylesRepeatingGroups(baseComponentId, columnSettings);
+  const style = useColumnStylesRepeatingGroups(baseComponentId, columnSettings, { isTitle: true });
```
lassopicasso: is it a reason for passing it down as a object with a prop, and not just a single prop itself?

src/utils/formComponentUtils.ts
```diff
@@ -137,16 +137,17 @@ function useTextAlignment(baseComponentId: string): 'left' | 'center' | 'right'
 export function useColumnStylesRepeatingGroups(
   baseComponentId: string,
   columnSettings: IGroupColumnFormatting | undefined,
+  options?: { isTitle?: boolean },
```
lassopicasso: Since this is the only prop and options is new param, I would keep the new param as a boolean instead of a object with a boolean, and keep the same name for consistent: `isTitle` instead of `options`

lassopicasso (commented):
Nice, a few comments to simplify the new param

lassopicasso (approved):
🥳 
@JamalAlabdullah great job, you can merge it now when tests run green

PR #4268, opened 2026-06-15
Fix focus- and SR-issues in repeating group

### LINKED ISSUE #1031: Notify screen readers that an element is deleted
### Description

When the user deletes an element in the repeating group there is no confirmation read by the screen reader. Visually you can see that element is gone, but this is not clear for screen reader users. 

Example solution: 
`<span class="visually-hidden" role="status">Rad slettet</span>`

### Additional Information

_No response_

### Tasks

_No response_

### Acceptance Criterias

_No response_

### LINKED ISSUE #1039: WCAG: Tab order is wrong in repeating group v2
### Description of the bug

This is related to issue #908 where the overall issue was fixed, but with additional scenarios the focus would stop working as intended. This is related to how the HTML is constructed for repeating groups. 

The repeating group container consists of a table of rows. Each row represents a group. New groups is added above the "add new group"-button. Hence the tab order doesn't visit the new group, but continues its path below.

So, the current workaround for #908, is to shift the focus upwards to the first focusable element in the edit container of the new group - using useRef and useEffect.

However, if the reference group is open and we then also opening a repeating group, the focus would be on reference instead of repeating. The reason for this is that both groups are rendered (reference is re-rendered when repeating is opened). And since reference is further down in the HTML DOM it will result in useRef using reference group as ref when the logic of focus is executed in useEffect. Ref. Video 1.

A similar issue occurs if we navigate back and forth in the edit container of repeating group. The focus will stop working as intended because it is no longer focused on the top element. Ref. Video 2.

We could have added more conditional statements to deal with these scenarios before executing the logic to shift focus. However, this would not be easy to work with in the long run.

A possible solution (in the short run?) is to move the buttons up in the HTML DOM so that new groups added are below them, but visually displayed as they are above. Here I also refer to the navigation buttons in the edit container of a repeating group. This would make it easier to deal with issues on the WCAG tab order.

In the longer run we may consider investing in an overall restructuring of repeating groups (and also group container?). 

**Update from UU test by SSB**
When pressing "Lagre og åpne neste" focus is moved to the previous "rediger" button. Focus should be on first field in the group. And if you press "Lagre og lukk" focus should be on "rediger" on the same row.

Comment from Lasse
Maybe some of theese issues will be fixed if we decide to go trough with this issue: https://github.com/Altinn/app-frontend-react/issues/1630
Making it a modal and locking the tab order inside the modal should be easier to control focus than opening up all of the tables.

Video 1:

https://user-images.githubusercontent.com/74791975/227233659-c17a934e-d8e0-4ee8-81fa-75544686c15d.mov

Video 2:

https://user-images.githubusercontent.com/74791975/227228927-a1f0c0ed-b44e-4e4c-aa71-b4b2702a1e44.mp4

### Steps To Reproduce

1. Go to a form with repeating groups

Issue 1:
2. Open a "reference group" then open a "repeating group"
3. Focus is on the reference group.. 

Issue 2: 
2. Open a "reference group".
3. Navigate back and forth. Focus stops working as intended.

### Additional Information

_No response_

Magnusrm:
## Description

- Screen reader not notified when a row is deleted
    - Added a visually-hidden aria-live status region driven by a deletedRowsCount counter at the single endDeletingRow choke point, covering both delete entry points.
- Focus fell to the page body after deletion (page title read, preempting the announcement)
  -  Focus now moves to the previous row (race-free), or the Add button when no rows remain; registered the table row as a focus target for read-only tables.
- multiPage navigation left focus on the Next/Back button
  - Focus now moves to the top (first field) of the edit container in both directions.
- Save & Close focus target 
  - Now returns focus to the same row's edit button instead of the next row's.
- Save & Open Next focus target
  -  Now moves focus to the first field of the next row's edit container instead of the next row's edit button.

## Related Issue(s)

- closes #{issue number}

## Verification/QA

- Automated tests
  - [ ] Has been added/updated
  
- Support in Altinn Studio
  
- Sprint board
- Labels
  

## Summary by CodeRabbit

* **New Features**
  * Added screen reader announcements for repeating-group row deletions, including updated remaining-row counts.
  * Enhanced keyboard focus management for multi-page editing and save flows, restoring focus to the appropriate edit controls and advancing focus after “save and next/close.”
  * Updated localized screen-reader text for row-deletion status in English, Bokmål, and Nynorsk.

* **Tests**
  * Added/extended accessibility tests covering live-region updates and focus movement across consecutive deletions and edit navigation scenarios.

### DIFF
--- a/src/layout/RepeatingGroup/Container/RepeatingGroupContainer.tsx
+++ b/src/layout/RepeatingGroup/Container/RepeatingGroupContainer.tsx
@@ -1,4 +1,4 @@
-import React, { forwardRef } from 'react';
+import React, { forwardRef, useEffect } from 'react';
 import type { JSX } from 'react';
 
 import { PlusIcon } from '@navikt/aksel-icons';
@@ -74,18 +74,15 @@ function RowDeletionAnnouncement() {
   const { numVisibleRows } = useRepeatingGroupRowState();
   const [message, setMessage] = React.useState('');
 
-  // Only announce when the count actually increases. Initializing the ref to the current count means a
-  // remount (e.g. the group being hidden then shown again) with a non-zero count announces nothing.
+  // Only announce when the count actually increases.
   const lastAnnouncedCount = React.useRef(deletedRowsCount);
 
-  React.useEffect(() => {
+  useEffect(() => {
     if (deletedRowsCount <= lastAnnouncedCount.current) {
       return;
     }
     lastAnnouncedCount.current = deletedRowsCount;
-    // Include the remaining row count so the message text differs between consecutive deletions.
-    // Screen readers do not reliably re-announce live-region text that is identical to the
-    // previous announcement.
+    // Include the remaining row count so the message gets re-announced on later deletions
     setMessage(langAsString('group.row_deleted_sr', [numVisibleRows]));
   }, [deletedRowsCount, numVisibleRows, langAsString]);
 
--- a/src/layout/RepeatingGroup/Container/RepeatingGroupContainer.tsx
+++ b/src/layout/RepeatingGroup/Container/RepeatingGroupContainer.tsx
@@ -23,6 +23,7 @@ import {
 import { useRepeatingGroupsFocusContext } from 'src/layout/RepeatingGroup/Providers/RepeatingGroupFocusContext';
 import { RepeatingGroupTable } from 'src/layout/RepeatingGroup/Table/RepeatingGroupTable';
 import { RepGroupHooks } from 'src/layout/RepeatingGroup/utils';
+import utilClasses from 'src/styles/utils.module.css';
 import { DataModelLocationProvider, useIndexedId } from 'src/utils/layout/DataModelLocation';
 import { useIsHidden } from 'src/utils/layout/hidden';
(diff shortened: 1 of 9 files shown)


src/layout/RepeatingGroup/Container/RepeatingGroupContainer.tsx
```diff
@@ -61,11 +62,44 @@ export const RepeatingGroupContainer = forwardRef((_, ref: React.ForwardedRef<HT
...
+  const { langAsString } = useLanguage();
+  const deletedRowsCount = useRepeatingGroupSelector((state) => state.deletedRowsCount);
+  const { numVisibleRows } = useRepeatingGroupRowState();
+  const [message, setMessage] = React.useState('');
+
+  // Only announce when the count actually increases. Initializing the ref to the current count means a
+  // remount (e.g. the group being hidden then shown again) with a non-zero count announces nothing.
+  const lastAnnouncedCount = React.useRef(deletedRowsCount);
+
+  React.useEffect(() => {
```
adamhaeger: maybe import useEffect? We will be removing the import React soon :)

adamhaeger (approved):
Tested and works well! Code also look good, but maybe consider if we need all the comments or if they can be worded a bit simpler?

